Issue #369, opened 2025-04-30
Trigger SchemaEditor from SubjectEditor


JeroenDeDauw:
AC:

* Schema edit link is present
    * Only if the user has schema edit permissions (use existing `SchemaAuthorizer`, optionally via `NeoWikiServices`) 
    * It looks good. Details: https://github.com/ProfessionalWiki/NeoWiki/pull/418#discussion_r2706035511
* Clicking edit link opens SchemaEditor dialog with the correct schema

For follow-ups:
* (optionally) the schema edit permission conditional behavior
* Dealing with closing of the SchemaEditor and updating the SubjectEditor UI accordingly

![Image](https://github.com/user-attachments/assets/fd71f149-7e7d-483e-9b96-15bde7fcda19)
                     
https://www.figma.com/design/tM9m8P0WOjKANlZ6KCBXFq/NeoWiki?node-id=352-13188&t=m0LmWDJI2BLxPE30-1

JeroenDeDauw:
I verified this is done. SubjectEditor is updated correctly on save of SchemaEditor

PR #428, opened 2026-01-20
Only show edit schema button to authorized users in SubjectEditorDialog

alistair3149:
For #369 

https://github.com/user-attachments/assets/d542403e-0dd9-4ecc-ab08-feb7a11f9be0

### DIFF
--- a/resources/ext.neowiki/src/components/SubjectEditor/SubjectEditorDialog.vue
+++ b/resources/ext.neowiki/src/components/SubjectEditor/SubjectEditorDialog.vue
@@ -22,6 +22,7 @@
 					<p class="cdx-dialog__header__subtitle">
 						{{ props.subject.getSchemaName() }}
 						<a
+							v-if="canEditSchema"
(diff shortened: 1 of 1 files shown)


resources/ext.neowiki/src/components/SubjectEditor/SubjectEditorDialog.vue
```diff
@@ -103,21 +105,36 @@ const open = ref( false );
...
+		await loadSchemaPermissions();
+		await loadSchema();
 	}
 } );
 
+async function loadSchemaPermissions(): Promise<void> {
+	try {
+		canEditSchema.value = await NeoWikiServices.getSchemaAuthorizer().canEditSchema( props.subject.getSchemaName() );
+	} catch ( error ) {
+		console.error( 'Failed to check schema permissions:', error );
```
alistair3149: For the error handling, should we aim for consistency (i.e., logging to both the console and notifications) or be situational?

In this case, I think that the schema edit permission error isn't critical for the SubjectEditor, so I did not display the error as a notification.

Issue #430, opened 2026-01-20
SubjectEditor: Improve Schema view/editing link UI


JeroenDeDauw:
We want
* Clear UI that indicates which Schema is used and does not confuse novice users and does not get in the way
* Easy schema editing for users that have schema editing permission

Nice to have: even users without the editng permission can view the schema definition

This task follows up to
* https://github.com/ProfessionalWiki/NeoWiki/issues/369

alistair3149:
For the entry point, I am leaning towards to the last option for now. The edit icon still have the confusion issue with switching the schema.

I agree that having the SchemaEditor in a read-only mode would be the superior option in the product standpoint. While there are more engineering work required within the SchemaEditor, there are a few benefits other than the UX improvements:
- Allow more granular permissions for schema editing (e.g. restrict editing certain properties or values)
- Move the schema editing permissions from SubjectEditor to the more appropriate SchemaEditor. It is cleaner semantic-wise.
- Display SchemaEditor instead of the JSON view on the schema page directly. That allows us to enhance the UX viewing the schema, and remove the friction where the user has to click "edit with form" to open the editor.

I am not 100% sure if this is a possible scenario. If an user somehow triggers the SchemaEditor without the edit schema permission, we don't have any permission handling in the SchemaEditor UI currently. Therefore, it is probably safer to handle the permission in SchemaEditor.

alistair3149:
> an icon clearly indicates intent (I want to edit/change something)
the rest of the UI (and I would include MW), either use icons or action verb links, unless it is something like "Preferences" or "Settings"

Agreed

> Would there not be similar confusion with the infobox pencil where the user might think it's for replacing or switching to a different Subject?

The edit icon in the infobox is aligned to the infobox border rather than the subject text, which helps to communicate that you're charging the infobox. Whereas for the schema, that isn't an option.

> if a user needs to look at the Schema(Editor) in order to use the SubjectEditor, then we have a serious UX issue

It is not the case, users shouldn't need access to the schema to edit the subject. 

> if a user wants to recreate the Schema on a different wiki, then getting access to the JSON makes more sense

That can be a valid use case. 

> if a user is just curious, then okay, but that does not seem like a common use case, and I'd argue the link is clutter because it is a mini call-to-action that does not serve a purpose in that workflow

On the flip side, it seems like a common use case in the wiki world. Templates such as Navbox or Infobox often contain links to the source code to facilitate editing, similar logic is also applicable for the schema.

[...]

PR #438, opened 2026-01-21
Improve schema link handling in SubjectEditor

alistair3149:
For #430 

With permission
image

Without permission
image

### DIFF
--- a/resources/ext.neowiki/tests/VueTestHelpers.ts
+++ b/resources/ext.neowiki/tests/VueTestHelpers.ts
@@ -24,7 +24,7 @@ export function createTestWrapper<TComponent extends DefineComponent<any, any, a
 }
 
 export function mockMwMessage(
-	messages: Record<string, string | ( ( ...params: string[] ) => string )>
+	messages: Record<string, string | ( ( ...params: string[] ) => string )>,
 ): void {
 	( global as any ).mw = {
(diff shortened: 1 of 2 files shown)


resources/ext.neowiki/src/components/common/I18nSlot.vue
alistair3149: This is used for cases where we want to inject the whole Vue slot into parameters in i18n messages, when `mw.message.parse()` does not cut it. In this example, we want to retain the event handling for the schema link.

malberts: Did you consider alternative approaches? I don't have anything specific in mind, but this feels like introducing complication for the sake of doing everything in a single message string.

Where else would we use this? Only if we follow this same design pattern of having a dynamic string in a message string?

JeroenDeDauw: Alistair and I briefly talked about this

Notes from me:
* Better to have done the trivial bad-i18n lego approach in this commit
* Then either create issue for discussion and priotization of fixing, or make a dedicated PR right away
* For prototype we do not need to have proper i18n everywhere. This is something we can fix later. Fixing the last few % here does not help us testing product fit and gathering feedback

PR #467, opened 2026-01-27
Disable Composer block-insecure in CI

JeroenDeDauw:
PHPUnit 9.x is blocked by security advisory PKSA-z3gr-8qht-p93v. This is not a concern in CI, so disable block-insecure globally.


JeroenDeDauw:
Looks like upstream fixed it: https://github.com/wikimedia/mediawiki/commit/314f8170d677cc588b26eeff81e4e7c8e9685669

PR #468, opened 2026-01-28
Fix: Make Schema reactive in AutomaticInfobox

JeroenDeDauw:
Schema was fetched once at mount time and never updated, so changes
made via the SchemaEditor were not reflected in the infobox until
page reload. Wrapping it in computed() makes it reactive.

Steps to reproduce:
1. Open a page with a Subject infobox (e.g. Professional Wiki)
2. Click edit on the infobox to open SubjectEditorDialog
3. Open the schema editor and change a property type (e.g. change
   Websites from URL to Text)
4. Save the schema and close the dialogs
5. Bug: the infobox still shows the old property type until reload

Manually verified, both the bug and the fix

### DIFF
--- a/resources/ext.neowiki/src/components/Views/AutomaticInfobox.vue
+++ b/resources/ext.neowiki/src/components/Views/AutomaticInfobox.vue
@@ -14,7 +14,7 @@
 					role="heading"
 					aria-level="3"
 				>
-					{{ schema.getName() }}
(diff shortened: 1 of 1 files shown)


alistair3149:
We should save in SchemaEditor directly because users might be using the SubjectEditor to get to the SchemaEditor, in which there are only changes to save for schema but not the subject.

PR #470, opened 2026-01-28
Add length constraint fields to Text property editor

JeroenDeDauw:
Add minLength and maxLength fields to TextAttributesEditor with natural language UI
("between X and Y characters") and validation:
- Positive integers >= 1 only
- minLength cannot exceed maxLength
- Invalid values are not emitted to parent

Also fixes Text.createPropertyDefinitionFromJson to include minLength/maxLength.

Manually tested in browser. All code changes thoroughly reviewed by Jeroen and Claude

Co-Authored-By: Claude Opus 4.5 <noreply@anthropic.com>


malberts:
Some additional concerns:
* (Future) Accessibility. Not sure how screen readers handle this.
* Boundary ambiguity. Might or might not be worse in other languages. "Minimum" and "Maximum" are explicit. In some contexts "between" excludes boundaries (`20<x<100`).

---

If we're going with a natural language approach, perhaps it would look better if the inputs do not look like typical input boxes stuck in between text.

image

(Just a rough GIMP edit.)

The "input" should resize according to the number, and be smaller when it's empty. 

But then this looks out of place with the rest of the forms.

I assume the natural language approach is not to try to save space.

---

Are there nice examples of UIs doing this? 

I might just be old school, but I think I prefer a form with a regular/consistent label+input structure.

JeroenDeDauw:
Observation: there might not be a single approach without a list of downsides. We also do not need to find the perfect one.

So, we identified a bunch of downsides for the approach in this PR. Are there approaches that have a more appealing set of tradeoffs?

In particular, the original approach, which we have still in use at the number type, does not have a clear spot to show the "min should be lower than max" error, nor does it visually group these fields in any way to suggest there is a relation between them. This might still be overall better than the approach in this PR. Is that what you think?

Issue #88, opened 2024-09-25
Editor: Show property descriptions


JeroenDeDauw:
Design up to implementer

From old UI:

![image](https://github.com/user-attachments/assets/a8910be7-566a-4478-aefe-2efc8b95ebb2)

The description of Property Definitions is optional

PR #500, opened 2026-02-01
Show property descriptions in SubjectEditor

JeroenDeDauw:
Fixes https://github.com/ProfessionalWiki/NeoWiki/issues/88

image

### DIFF
--- a/extension.json
+++ b/extension.json
@@ -189,7 +189,7 @@
 						"cdxIconAdd",
 						"cdxIconTrash",
 						"cdxIconArrowPrevious",
-						"cdxIconInfoFilled",
+						"cdxIconInfo",
 						"cdxIconEdit",
 						"cdxIconLink",
 						"cdxIconCheck",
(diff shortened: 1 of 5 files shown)


JeroenDeDauw:
Second commit swicthes to a vertical-space-saving approach, which seems slightly better, but I definitely want UI input on this one @alistair3149 @malberts. Also by any means tweak this further yourselves, I am done with this PR at present.

One thing to keep in mind: we likely also want descriptions to somehow be visible in the display UIs like the automatic infobox.

image

JeroenDeDauw:
@alistair3149 I would hold off on that extraction. Benefit is small and there are costs as per https://www.entropywins.wtf/blog/2017/09/06/the-fallacy-of-dry/

malberts:
> implemented as [help text](https://doc.wikimedia.org/codex/v1.14.0/components/demos/field.html#with-rich-description-and-help-text) in the Codex Field component

I think the main issue with using the semantically correct help text is it is going to cause the form to get really bloated, if the help text is actually meant to help and not just turn the label into a sentence.

At least in Wikidata this extra help text would have required you to go to the Property page (unless there's a gadget that shows it immediately?).
Some of these Wikidata descriptions take up a non-trivial amount of space:

Screenshot_20260202_225652

Issue #481, opened 2026-01-29
Make Schema name in Infobox interactive


malberts:
From https://github.com/ProfessionalWiki/NeoWiki/pull/468#issuecomment-3818830666

Idea: Change Schema name into a link and open SchemaEditor if the user has permission, otherwise link to Schema page. Or skip the latter and just copy the same Schema link logic in SubjectEditor.

PR #501, opened 2026-02-01
Make Schema name interactive in AutomaticInfobox

JeroenDeDauw:
Fixes https://github.com/ProfessionalWiki/NeoWiki/issues/481

Alternative to #490.

The Schema name in AutomaticInfobox is now a link to the `Schema:` page. This gives users access to version history,
editing, and other page actions.

## Test plan

- [x] All existing tests pass
- [x] New test for the schema link
- [x] Verified in browser


JeroenDeDauw:
TBH I am unsure now if opening the editor is the right move. Always linking to the schema page might be better. You can easily open the edit with form tab there anyway. You might want to go to the schema page to see version history or some other such action.

Issue #504, opened 2026-02-02
Replace table implementation in `SchemaDisplay` with `CdxTable`


alistair3149:
Follow-up to #502 

Use the Codex [Table](https://doc.wikimedia.org/codex/v1.14.0/components/demos/table.html) component instead of relying on `<table>` elements and custom styles.

Issue #505, opened 2026-02-02
Use `CdxInfoChip` for property type value


alistair3149:
Follow-up to #502 

Use Codex [InfoChip](https://doc.wikimedia.org/codex/v1.14.0/components/demos/info-chip.html) component to display the property type value, including the icon and the label.

PR #509, opened 2026-02-02
Improve SchemaDisplay UI

alistair3149:
Follow-up: #502 
Closes: #504, #505

- Wrap the whole UI into Codex Table
- Use Codex InfoChip for displaying property types
- Add edit button for editing schema
- Integrate schema description into the component
- Integrate empty state into the component
- Extract useSchemaPermissions
- Remove "edit with form" from MW view menu

image

### DIFF
--- a/extension.json
+++ b/extension.json
@@ -50,9 +50,7 @@
 
 		"SpecialPage_initList": "ProfessionalWiki\\NeoWiki\\EntryPoints\\NeoWikiHooks::onSpecialPageInitList",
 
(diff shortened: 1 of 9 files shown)


alistair3149:
> > Remove "edit with form" entry point (?)
> 
> If we remove that, will we still have a way of triggering the edit form immediately via a direct link?

I should have been more specific. The entry point is only referring to the "edit with form" link in the views menu of the skin. The edit button will still trigger the link for now

PR #510, opened 2026-02-03
Add European museums demo data

JeroenDeDauw:
Fixes https://github.com/ProfessionalWiki/NeoWiki/issues/506

Add a European museums dataset (museums, artists, artworks, exhibitions,
cities) that naturally showcases all currently implemented NeoWiki features
including cross-schema relations, child subjects, multiple value types,
and property constraints.

## Summary

- **5 new schemas**: Museum, Artist, Artwork, Exhibition, Attendance
- **21 new subject pages**: 4 cities, 4 museums, 5 artists, 6 artworks, 2 exhibitions
- **3 child subjects** (Attendance records on Musée d'Orsay page)
- **1 new page**: Museum Collection (cross-page infobox display)
- **Modify** City schema (add Population, Website), Main Page
- **Fix** default wikitext fallback in SubjectPageSource (remove defunct `{{#infobox:}}`)

### DIFF
--- a/DemoData/Page/Museum_Collection.wikitext
+++ b/DemoData/Page/Museum_Collection.wikitext
@@ -1,0 +1,14 @@
+This page shows infoboxes with data from various museum and artwork pages, demonstrating cross-page subject display.
+
+== Museums ==
+
+<div class="ext-neowiki-view" data-subject-id="sEpfwJLnx3Kze3t"></div>
+<div class="ext-neowiki-view" data-subject-id="sEpfwJLnxyQy6vR"></div>
+<div class="ext-neowiki-view" data-subject-id="sEpfwJLnyuR5E4i"></div>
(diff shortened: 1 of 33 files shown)


DemoData/Page/Museum_Collection.wikitext
malberts: Unrelated to this PR, but I think this page highlights a styling issue:
Image

Likely a page would have more text, but if your page is basically what this is, which is valid if the point is to just group and list subjects, then the headings seem weird. There's still going to be massive whitespace, but I expect "Selected Artworks" to be below the museums, but above the artworks. At least right now, that heading loses meaning when everything just floats to the side.

alistair3149: Since we don't have the parser functions for View currently, everything is being injected into the app container. It's a to-do when we implement Views.

PR #511, opened 2026-02-03
Hide schema selector after schema is selected in SubjectCreatorDialog

malberts:
For https://github.com/ProfessionalWiki/NeoWiki/pull/491#issuecomment-3833732765

## Summary

- Wrap the schema selection section (title, toggle buttons, and schema lookup/new-schema blocks) in `v-if="!selectedSchemaName"` so it disappears once a schema is chosen
- `resetForm()` already clears `selectedSchemaName`, so reopening the dialog restores the selector

## Test plan

- [x] Verify in the browser that schema selector disappears after picking a schema
- [x] Verify reopening the dialog shows the selector again
- [x] All existing and new TS tests pass (`make ts-test`)

### DIFF
--- a/resources/ext.neowiki/src/components/SubjectCreator/SubjectCreatorDialog.vue
+++ b/resources/ext.neowiki/src/components/SubjectCreator/SubjectCreatorDialog.vue
@@ -13,29 +13,31 @@
 			:use-close-button="true"
 			@default="open = false"
 		>
-			<p>
-				{{ $i18n( 'neowiki-subject-creator-schema-title' ).text() }}
-			</p>
+			<template v-if="!selectedSchemaName">
+				<p>
+					{{ $i18n( 'neowiki-subject-creator-schema-title' ).text() }}
+				</p>
 
-			<CdxToggleButtonGroup
-				v-model="selectedSchemaOption"
-				class="ext-neowiki-subject-creator-schema-options"
-				:buttons="toggleButtons"
-			/>
-
-			<div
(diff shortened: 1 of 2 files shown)


malberts:
This only hides the selector after a schema is selected.

Follow-up questions still related to https://github.com/ProfessionalWiki/NeoWiki/pull/491#issuecomment-3833732765:
* do we want a quick way to change the schema inside the dialog? It seems easy enough to just close, but maybe that is annoying? Main concern is switching schema should not make you lose entered data by accident.
* do we want to show the selected schema as a link, as in `SubjectEditorDialog`? Allowing user to modify the schema. Or do we intentionally not want that?

alistair3149:
> do we want a quick way to change the schema inside the dialog? It seems easy enough to just close, but maybe that is annoying? Main concern is switching schema should not make you lose entered data by accident.

In the original mockup, there is a back button to go back to the previous step.
That can be utilize to go back to the existing schema page.

> do we want to show the selected schema as a link, as in SubjectEditorDialog? Allowing user to modify the schema. Or do we intentionally not want that?

I don't have an opinion on that, but perhaps we should link it for consistency.

JeroenDeDauw:
Agree with alistair that a back button seems plausibly the way to go

As for a link to edit the schema: idk. would hold off on that till it is more clear there that would be a net-positive change. 

The dialog header is "Creating subject". What about "Create $schema", so "Create Company" or similar? Same for the create button at the bottom.

PR #515, opened 2026-02-03
Clear floats on headings following infobox views

malberts:
Fixes https://github.com/ProfessionalWiki/NeoWiki/pull/510#discussion_r2759309458

## Summary

- Headings between groups of floated infoboxes (e.g. on Museum_Collection) appeared beside the infoboxes instead of below them
- Add a CSS rule to clear floats on headings that follow `.ext-neowiki-view` elements, scoped to tablet+ to match the float rule
- Covers both legacy heading elements (`h2`-`h6`) and MW 1.42+ `.mw-heading` wrapper for forward compatibility

## Test plan

- [x] Visit Museum_Collection page
- [x] Verify "Selected Artworks" heading appears below the museum infoboxes, not beside them
- [x] Verify infobox float behavior on tablet+ is otherwise unchanged

### DIFF
--- a/resources/ext.neowiki/src/components/Views/AutomaticInfobox.vue
+++ b/resources/ext.neowiki/src/components/Views/AutomaticInfobox.vue
@@ -188,6 +188,7 @@ const propertiesToDisplay = computed( function(): Record<string, PropertyDefinit
 	}
 }
 
+// TODO: This is a temporary fix until we implement Views.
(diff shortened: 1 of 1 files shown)


malberts:
> This is not optimal since it is applying styles to elements that are not a part of NeoWiki.

True, but then we should also avoid adding styles in NeoWiki that affect how other non-NeoWiki elements get rendered (i.e. the problem here), even if we do not directly reference those elements.

Issue #427, opened 2026-01-20
Implement Subject creation flow


JeroenDeDauw:
AC:
* A simple temporary UI element that allows starting the "create subject flow" is shown on existing pages. This does not need to look nice. It opens the (to be created) "Creation Choice Dialog"
* The "Creation Choice Dialog" allows picking an existing Schema or choosing to create a new Schema. The former opens the Subject Editor, the later opens the Schema Editor, in such a way that when the new Schema is created, the Subject Editor is opened
* On exiting the flows, AutomaticInfobox should is shown by inserting the temporary HTML content thing we have into the page

Key considerations:
* Our Figma mockups are only directionally correct; they still have a bunch of issues.
* As usual, not everything needs to be implemented at once. The "create new schema" flow in particular seems like most of the work for a small part of the value, so should not block the "use existing schema" flow. Inserting the AutomaticInfobox HTML thing is also a nice dedicated task

Figma
* [Mockups](https://www.figma.com/design/tM9m8P0WOjKANlZ6KCBXFq/NeoWiki?node-id=203-28204&t=QhKMlWPceyGOHzmf-0)
* [Prototype](https://www.figma.com/proto/tM9m8P0WOjKANlZ6KCBXFq/NeoWiki?node-id=203-28720&viewport=210%2C-841%2C0.1&t=EDz171i7O3c6kKz2-1&scaling=min-zoom&content-scaling=fixed&starting-point-node-id=203%3A28204&show-proto-sidebar=1)

Related chats:
* https://chat.professional.wiki/pro-wiki/pl/uofyzgcehfgcxbxadsgkqws6ea (UI)
* https://chat.professional.wiki/pro-wiki/pl/ss438pj18pd3feoh3zauhfzc9w (related but seperate add-infobox flow, out of scope here)

Existing code:
* SubjectEditor(Dialog).vue
* SchemaEditor(Dialog).vue
* SchemaStore (for getting Schema names ([old usage](https://github.com/ProfessionalWiki/NeoWiki/pull/332/files#diff-3bde5b6c8ad32e2b9d68ae86b0ca7ca74d71b82474d44acaff332688b2fcb5b7)), new implementation might differ, i.e. maybe we just list all Schema for now rather than doing search autocomplete)
[...]

PR #516, opened 2026-02-03
Implement "Create new" schema workflow in SubjectCreatorDialog

malberts:
## Summary

For https://github.com/ProfessionalWiki/NeoWiki/issues/427

- Replace the TODO placeholder in the "Create new" toggle with a two-step inline flow
- Step 1: Schema name input + embedded SchemaEditor with validation (non-empty, unique name) and "Create schema" button
- Step 2: Subject creation (label + SubjectEditor), identical to the existing "Use existing" flow
- Dialog widens when SchemaEditor is shown to match SchemaEditorDialog dimensions

## Test plan

- [x] Check in browser: navigate to a page without a subject, click "Create subject", toggle to "Create new"
- [x] Verify schema name input and SchemaEditor appear
- [x] Try creating with empty name — error shown
- [x] Try creating with existing schema name — error shown
- [x] Create schema with valid name — transitions to subject creation step
- [x] Create subject after schema creation — completes successfully

### DIFF
--- a/resources/ext.neowiki/src/components/SubjectCreator/SubjectCreatorDialog.vue
+++ b/resources/ext.neowiki/src/components/SubjectCreator/SubjectCreatorDialog.vue
@@ -56,7 +56,7 @@
 
 					<SchemaEditor
 						ref="schemaEditorRef"
-						:initial-schema="blankSchema"
+						:initial-schema="newSchema"
 					/>
 				</div>
 			</template>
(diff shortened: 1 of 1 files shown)


alistair3149:
> Does this UI approach make sense?

The structure looks fine, having it in the same dialog make sense. We can always polish it later.

JeroenDeDauw:
> This is a general issue with all our dialogs. Clicking close, or outside, closes the dialog immediately. We should probably make all dialogs show a confirmation if any state was changed.

Sounds plausible, I did not check the current state. I do want to highlight "if any state was changed". If we implement that, let's please do a simple approach, where any change flips a boolean flag. Let's not go down the complexity-ridden path of storing the entire initial and current state and doing a comparison again, we wasted so much time on that for no reason at some point to have some conditional enabling of the save button that no one needs.

alistair3149:
> Should the Schema be saved before the Subject is saved?

Perhaps we can make that into an option in the confirmation dialog when the subject creation flow is being aborted.
If the user has created a schema there, then we can show an additional option to allow saving the schema.

JeroenDeDauw:
It's probably time to close https://github.com/ProfessionalWiki/NeoWiki/issues/427 and create new issue(s) for the identified follow up tasks

PR #548, opened 2026-02-05
Add missing aria-label for delete property button

alistair3149:


JeroenDeDauw:
Where is this supposed to be visible? Don't forget that screenshots are often nice for UI PRs.

I looked at the code and thought I would see it here on hover or something like that, but no

image

Issue #543, opened 2026-02-05
URL formatter does not fully escape HTML — potential XSS


JeroenDeDauw:
## Problem

`UrlFormatter.formatUrlAsHtml()` in `Url.ts` does not fully escape HTML when the input is not a valid URL. The `catch` branch (line 78) passes the string through `DOMPurify.sanitize()`, which strips dangerous tags like `<script>` but allows benign tags like `<strong>` through unescaped. The expected behavior is that *all* HTML should be escaped in URL display.

The test at `Url.spec.ts:56` explicitly documents this:

```ts
).toBe( 'evil <strong>bold</strong>' ); // FIXME: this is NOT what we want. EVERYTHING should be escaped, and nothing should be omitted.
```

## Origin

Originally flagged via FIXME in test added around issue #251. Verified on 2025-02-05 that the FIXME and the underlying behavior are still present on master at `resources/ext.neowiki/src/domain/propertyTypes/Url.ts:78` and `resources/ext.neowiki/tests/domain/propertyTypes/Url.spec.ts:56`.

## Expected behavior

Invalid URL strings should have all HTML fully escaped (e.g., `&lt;strong&gt;bold&lt;/strong&gt;`), not selectively sanitized.

PR #560, opened 2026-02-06
Fix URL formatter HTML escaping

malberts:
Fixes https://github.com/ProfessionalWiki/NeoWiki/issues/543

DOMPurify.sanitize() strips dangerous tags like `<script>` but lets
benign tags like `<strong>` through as rendered HTML. Replace it with
`escapeHtml()` that converts all HTML-significant characters to entities.
Also escape the displayed URL in the valid-URL path, and remove the
now-unused DOMPurify import.

## Test plan

- [x] `make ts-test filter=Url` passes (59 tests)
- [x] `make ts-lint` passes
- [x] Valid URLs with `&` in query string display correctly (not double-escaped)
- [x] Valid URLs render as clickable links as before

### DIFF
--- a/resources/ext.neowiki/src/domain/propertyTypes/Url.ts
+++ b/resources/ext.neowiki/src/domain/propertyTypes/Url.ts
@@ -54,6 +54,12 @@ export class UrlType extends BasePropertyType<UrlProperty, StringValue> {
 
 }
 
+const ALLOWED_PROTOCOLS: readonly string[] = [ 'http:', 'https:' ];
+
+function hasAllowedProtocol( url: URL ): boolean {
+	return ALLOWED_PROTOCOLS.includes( url.protocol );
+}
+
 export class UrlFormatter {
 
 	public constructor(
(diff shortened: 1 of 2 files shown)


resources/ext.neowiki/src/domain/propertyTypes/Url.ts
```diff
@@ -85,11 +84,20 @@ export class UrlFormatter {
 			const pathName = url.pathname === '/' ? '' : url.pathname;
 			return url.hostname + pathName + url.search + url.hash;
 		} catch ( _ ) {
-			return DOMPurify.sanitize( urlString );
+			return urlString;
 		}
 	}
 }
 
+function escapeHtml( text: string ): string {
```
malberts: The custom `escapeHtml` function is functionally equivalent to `mw.html.escape()` from MediaWiki core (`mediawiki.base` module, available on every page). Both escape the same 5 characters (`&`, `<`, `>`, `"`, `'`) to the same HTML entities.

Reasons to use `mw.html.escape()`:
- Standard MediaWiki utility, avoids maintaining a custom function
- Recognized by anyone familiar with the MW JS API

Reasons to keep the custom function:
- This is domain-layer code — better to avoid coupling it to MediaWiki
- Works in Vitest without needing to mock the `mw` global

resources/ext.neowiki/tests/domain/propertyTypes/Url.spec.ts
```diff
@@ -78,6 +78,12 @@ describe( 'UrlFormatter', () => {
 			).toBe( '<a href="https://professional.wiki/en/mediawiki-development#anchor">professional.wiki/en/mediawiki-development#anchor</a>' );
 		} );
 
+		it( 'escapes ampersands in displayed URL', () => {
+			expect(
+				( new UrlFormatter( newUrlProperty() ) ).formatUrlAsHtml( 'https://example.com/?a=1&b=2' ),
+			).toBe( '<a href="https://example.com/?a=1&b=2">example.com?a=1&amp;b=2</a>' );
```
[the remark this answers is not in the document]
alistair3149: Same issue. refer to https://github.com/ProfessionalWiki/NeoWiki/pull/560#discussion_r2775438086

resources/ext.neowiki/src/domain/propertyTypes/Url.ts
```diff
@@ -85,14 +95,28 @@ export class UrlFormatter {
...
+
+export function isValidUrl( urlString: string ): boolean {
+	const protocolMatch = urlString.match( /^([a-z][a-z\d+.-]*):\/\//i );
+	if ( protocolMatch && !ALLOWED_PROTOCOLS.includes( protocolMatch[ 1 ].toLowerCase() + ':' ) ) {
+		return false;
+	}
+
 	const pattern = new RegExp(
-		'^(https?://)?' +
+		'^([a-z][a-z\\d+.-]*://)?' +
```
alistair3149: More flexible protocol matching

Issue #371, opened 2025-04-30
Add Subject selector to Relation Value editing


JeroenDeDauw:
AC:
* User can find and select the target Subject
* Subject IDs are not shown. Both during the selection process, and after selection, the user only sees Subject labels 

Follow-up to
* https://github.com/ProfessionalWiki/NeoWiki/pull/361
* https://github.com/ProfessionalWiki/NeoWiki/issues/50

PR #566, opened 2026-02-07
Add SubjectLookup component for Relation value editing

alistair3149:
For https://github.com/ProfessionalWiki/NeoWiki/issues/371

Tested all behaviors manually but haven't done an in-depth review of the code quality yet.

https://github.com/user-attachments/assets/a42cbfd9-28fa-4199-818c-c0edaf02f650

Follow-ups:
- Show the error `The value must match a subject` or something similar, when `SubjectLookup` has a value but there are no matching subject (i.e. no subject is selected in the lookup).

## Summary

- Adds **SubjectLookup** component: CdxLookup wrapper that searches stub subjects filtered by `targetSchema` and resolves labels for pre-selected subjects
- Adds **NeoMultiLookupInput** component: generic multi-input wrapper with scoped slot, handling normalization, stable keys, focus-aware slot preservation, touched tracking, and validation message display
- Refactors **RelationInput** to use SubjectLookup directly for single-value properties and NeoMultiLookupInput + SubjectLookup for multiple-value properties

## Test plan

- [x] `make ts-test`, `make ts-lint`, `make ts-build` all pass
- [x] Verify single relation input renders a single SubjectLookup
- [x] Verify multiple relation input renders multi-input with trailing empty slot
- [x] Verify searching filters by targetSchema and query text
- [x] Verify removing a middle value preserves remaining values
- [x] Verify editing a filled input preserves the slot while focused

### DIFF
--- a/Docker/SettingsTemplate.php
+++ b/Docker/SettingsTemplate.php
@@ -22,7 +22,7 @@ $wgDebugComments = false;
 ## Uncomment this to disable output compression
 # $wgDisableOutputCompression = true;
 
-$wgSitename = "SiteName";
+$wgSitename = "NeoWiki Demo";
 
 ## The URL base path to the directory containing the wiki;
 ## defaults for all runtime URL paths are based off of this.
@@ -149,9 +149,14 @@ $wgRightsIcon = "";
 # Path to the GNU diff3 utility. Used for conflict resolution.
 $wgDiff3 = "/usr/bin/diff3";
 
+## Namespace settings
+
+# Set meta namespace
+$wgMetaNamespace = 'Wiki';
+
 ## Default skin: you can change the default skin. Use the internal symbolic
 ## names, ie 'vector', 'monobook':
-$wgDefaultSkin = "vector";
+$wgDefaultSkin = "vector-2022";
 
 # Enabled skins.
 # The following skins were automatically enabled:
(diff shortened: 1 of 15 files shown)


resources/ext.neowiki/src/components/common/SubjectLookup.vue
alistair3149: Should we put `SubjectLookup` inside `common/` or `Value/`? It is technically not a `Value`.
We should consider relocating `SchemaLookup` since it functions similarly to `SubjectLookup`.

alistair3149 (commented):
I have addressed the reviews other than 1 and 3 in the minor points:

_Written by Claude Code, Opus 4.6, on behalf of @alistair3149._

> `getSubjectLabels` on the `SubjectLookup` interface mixes get-by-ID and search concerns — a separate interface would be cleaner

> `SubjectStore.getSubjectLabels` is a pure pass-through with no caching or state management — unclear if the store is the right layer for transient search results

Acknowledged. These point at the same underlying issue: the search functionality is threaded through SubjectStore and the `SubjectLookup` domain interface without adding value at either layer. Opened https://github.com/ProfessionalWiki/NeoWiki/issues/583 to discuss the right approach, as this is an architectural decision that would benefit from input from the team.

PR #587, opened 2026-02-13
Show validation error for unmatched text in SubjectLookup

JeroenDeDauw:
## Summary

- When a user types text in a SubjectLookup but does not select a subject from the dropdown, an inline error message is shown after blur
- SubjectLookup owns this validation internally and reports it to the parent via the blur event
- RelationInput suppresses its own required-field error when SubjectLookup reports unmatched text, avoiding duplicate error messages

## Test plan

- [x] Edit a subject with a relation property, type text without selecting from dropdown, blur the field — error message appears
- [x] Select a subject from the dropdown — error clears
- [x] After error shows, type new text — error clears
- [x] Clear the input entirely — no error shown
- [x] Required relation field with unmatched text shows only the SubjectLookup error, not both errors

image

### DIFF
--- a/extension.json
+++ b/extension.json
@@ -289,6 +289,7 @@
 				"neowiki-subject-editor-error",
 				"neowiki-subject-lookup-placeholder",
 				"neowiki-subject-lookup-no-results",
+				"neowiki-subject-lookup-no-match",
 				"neowiki-schemas-column-name",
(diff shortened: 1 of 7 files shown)


alistair3149:
Manually tested and LGTM.
There are some minor points that Claude pointed out but I will address them in a followup.

PR #599, opened 2026-02-17
Add missing test and fix stale state in SubjectLookup validation

alistair3149:
Follows-up to https://github.com/ProfessionalWiki/NeoWiki/pull/587

## Summary

- Add SubjectLookup test for the "clear input entirely" scenario that was in #587's test plan but had no corresponding test
- Reset `singleHasUnmatchedText` in RelationInput when a selection is made, so the flag doesn't stay stale after the unmatched text state no longer applies

## Test plan

- [x] SubjectLookup tests pass (18/18)
- [x] RelationInput tests pass (22/22)

### DIFF
--- a/resources/ext.neowiki/src/components/Value/RelationInput.vue
+++ b/resources/ext.neowiki/src/components/Value/RelationInput.vue
@@ -141,6 +141,8 @@ function validateAndUpdateMessages(): void {
 }
 
 function onSingleSelectionChanged( id: string | null ): void {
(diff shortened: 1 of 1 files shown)


alistair3149:
Yes, I tested the SubjectLookup in SubjectEditor, following the manual tests in #587 to ensure that nothing breaks.
Since this PR contains no user-facing changes that is observable from the browser, I did not include it in the PR comment.

Issue #523, opened 2026-02-04
Implement Back button in SubjectCreator


malberts:
Part of https://github.com/ProfessionalWiki/NeoWiki/issues/427

Follow-up to https://github.com/ProfessionalWiki/NeoWiki/pull/511

AC:
* After selecting "Use existing" or "Create new", show a back button to allow selecting/creating a different schema

Undecided: How to prevent accidentally losing the current state?
* Show a confirmation dialog if anything was entered? (related: https://github.com/ProfessionalWiki/NeoWiki/issues/520)
* Hold on to state, so if the user selects the same existing schema, the previously entered values are still present? Or if user selects new again, show the previously added properties?

PR #633, opened 2026-02-24
Add back button to SubjectCreator dialog

alistair3149:
Went through test plan manually.

Fixes https://github.com/ProfessionalWiki/NeoWiki/issues/523

Allows returning to schema selection after choosing "Use existing" or
"Create new". Preserves the selected schema option (toggle state) when
going back.

## Test plan
- [x] Open SubjectCreator dialog
- [x] Select an existing schema → verify back button appears
- [x] Click back → verify return to schema selection with "Use existing" still selected
- [x] Switch to "Create new", create a schema → verify back button appears
- [x] Click back → verify return to schema selection with "Create new" still selected

https://github.com/user-attachments/assets/b911cf3a-e5f5-4ca2-b522-814a9b949349

### DIFF
--- a/extension.json
+++ b/extension.json
@@ -276,6 +276,7 @@
 				"neowiki-subject-creator-schema-created",
 				"neowiki-subject-creator-save-with-schema",
 				"neowiki-subject-creator-creating-schema",
+				"neowiki-subject-creator-back",
 				"neowiki-edit-schema",
 				"neowiki-schema-editor-summary-default",
 				"neowiki-schema-editor-success",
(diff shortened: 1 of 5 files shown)


alistair3149:
> If I click on the wrong thing, hitting esc to go back seems natural.

I get the logic. Visually it feels like navigating a mini-app, but browsers and assistive tech only see a single `<dialog>` container, meaning the standard rules still apply:

1. It creates an accessibility trap: If a keyboard user on Step 3 decides to quit the process entirely, forcing them to press `ESC` three times just to close the window violates [WCAG 2.1.2](https://www.w3.org/WAI/WCAG21/Understanding/no-keyboard-trap.html). `ESC` must be an immediate exit hatch.

2. "Undo" vs. "Abort": Fixing a localized mistake is an "Undo" action meant for explicit "Back" UI buttons. `ESC` is universally hardwired as "Abort/Exit the entire task." Overriding this breaks established user muscle memory.

Avoiding stacked dialogs was absolutely the right call, but we still need to respect the strict browser rules for the single dialog we are using. Let's keep `ESC` tied to the close action.

Issue #539, opened 2026-02-05
Show correct Subject data when viewing old page revisions


JeroenDeDauw:
The Automatic Infobox always shows the latest version of the Subject data, even when viewing an old page revision. It should show the Subject data as it was at that revision.

Originally noted in https://github.com/ProfessionalWiki/NeoWiki/pull/237#issuecomment-2422671898

PR #638, opened 2026-02-25
Add revisionId parameter to GetSubject API

alistair3149:
For https://github.com/ProfessionalWiki/NeoWiki/issues/539

Adds backend support for fetching historical Subject data via the REST API. When `?revisionId=N` is
provided, the API returns Subject data as it existed at that revision instead of the latest version.

This is the first part of the revision-aware infobox feature. Follow-up PRs will wire this into the
hooks (HTML injection) and frontend (TypeScript/Vue).

**Changes:**

- `PointInTimeSubjectLookup`: new `SubjectLookup` implementation that reads Subject data from a specific
  revision. For subjects on the primary page, it reads directly from that revision's slot. For subjects
  on other pages (referenced via relations), it finds the latest revision of that page at or before the
  primary revision's timestamp.
- `GetSubjectApi`: accepts optional `revisionId` query parameter. When provided, creates a
  `PointInTimeSubjectLookup` instead of using the default repository.
- `NeoWikiExtension`: adds `newGetSubjectQueryForRevision` factory method.
- Returns 404 for non-existent revision IDs.

**Manually verified:**

```
# Latest data
curl '.../neowiki/v0/subject/sEpfwJLo1kFh84j'
→ "label": "Claude Monet (updated)"

# Old revision
curl '.../neowiki/v0/subject/sEpfwJLo1kFh84j?revisionId=35'
→ "label": "Claude Monet"

# Invalid revision
curl '.../neowiki/v0/subject/sEpfwJLo1kFh84j?revisionId=999999'
→ 404 "Revision not found: 999999"
```

> Implemented by @alistair3149 with Claude Code.
> Context: the NeoWiki codebase and implementation plan at `docs/plans/2026-02-25-revision-aware-infobox.md`.

### DIFF
--- a/src/EntryPoints/REST/GetSubjectApi.php
+++ b/src/EntryPoints/REST/GetSubjectApi.php
@@ -4,8 +4,11 @@ declare( strict_types = 1 );
 
 namespace ProfessionalWiki\NeoWiki\EntryPoints\REST;
 
+use MediaWiki\MediaWikiServices;
 use MediaWiki\Rest\Response;
 use MediaWiki\Rest\SimpleHandler;
+use MediaWiki\Revision\RevisionRecord;
+use ProfessionalWiki\NeoWiki\Application\Queries\GetSubject\GetSubjectQuery;
 use ProfessionalWiki\NeoWiki\NeoWikiExtension;
 use ProfessionalWiki\NeoWiki\Presentation\RestGetSubjectPresenter;
 use Wikimedia\ParamValidator\ParamValidator;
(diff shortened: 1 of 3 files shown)


src/EntryPoints/REST/GetSubjectApi.php
```diff
@@ -17,7 +18,20 @@ class GetSubjectApi extends SimpleHandler {
 
 	public function run( string $subjectId ): Response {
 		$presenter = new RestGetSubjectPresenter();
-		$query = NeoWikiExtension::getInstance()->newGetSubjectQuery( $presenter );
+
+		$revisionId = $this->getRequest()->getQueryParams()['revisionId'] ?? null;
+
+		try {
+			if ( $revisionId !== null ) {
```
alistair3149: Should we use a ternary operator here?

JeroenDeDauw (commented):
Note: I am quite tired ATM and have not fully validated the below solution, but at first glance it looks reasonable.

_Review by Claude Code, Opus 4.6, prompted by @JeroenDeDauw. Context: the NeoWiki codebase and PR diff._

The handler catches `Exception` around the factory call to handle non-existent revision IDs. This uses exceptions for expected input validation, and the broad catch would swallow unexpected errors (e.g. database failures) as 404s.

Prefer validating the revision in the handler and passing a `RevisionRecord` to the factory:

```php
$revisionId = $this->getValidatedParams()['revisionId'] ?? null;

if ( $revisionId !== null ) {
    $revision = MediaWikiServices::getInstance()->getRevisionLookup()->getRevisionById( $revisionId );
    if ( $revision === null ) {
        return $this->getResponseFactory()->createHttpError( 404, [
            'message' => 'Revision not found: ' . $revisionId,
        ] );
    }
}
```

Then `newGetSubjectQueryForRevision` takes a `RevisionRecord` instead of `int` and doesn't need to throw. No catch needed at all.

Issue #646, opened 2026-03-03
SubjectCreator: Create Schema Back button should return to Schema Editor


malberts:
From https://github.com/ProfessionalWiki/NeoWiki/pull/633#issuecomment-3990808922

Related issue about Schema abandonment: https://github.com/ProfessionalWiki/NeoWiki/issues/537

When creating a new Subject with a new Schema: after creating the Schema, and landing on the Subject Editor, if I go back I should return to the Schema Editor.

Current behavior takes you back to the Schema selector:

https://github.com/user-attachments/assets/d1d2df69-cdca-4795-b42c-c2a9ea5560e2

This is due to the Schema still being saved in the Schema Editor. In #537 we will change how state is tracked, but that issue is specifically about abandonment, whereas this issue is about continuing the flow, but going back to adjust the new Schema.

Issue #537, opened 2026-02-05
Handle Schema and Subject abandonment in SubjectCreator


malberts:
From https://github.com/ProfessionalWiki/NeoWiki/pull/516#issuecomment-3854510364

When creating a new subject, the user can create a new schema. This new schema can be abandoned before saving, or the subject can be abandoned after creating the schema.

In the current implementation the schema is saved once the SchemaEditor form is submitted, even if the subject is not saved afterwards.

Pros:
* User does not lose a newly built Schema if the Subject is abandoned.

Cons:
* User can create junk schemas while just exploring with no intention of creating subjects.

It is unclear whether this is the best approach. If we do not save the schema immediately, and instead wait until subject save, then there is a risk that the user can lose the schema. Possible solution inspired by https://github.com/ProfessionalWiki/NeoWiki/pull/516#issuecomment-3854629089):
* Do not save schema immediately when submitting the SchemaEditor form
* If the dialog is abandoned, show a confirmation dialog to confirm whether schema should be abandoned
* If the SchemaEditor form is submitted, but the SubjectEditor form is abandoned, show a confirmation dialog with options:
  * Abandon everything
  * Abandon subject, but save schema
  * Return to form

Related issue about confirmation dialogs for other dialog abandonment: https://github.com/ProfessionalWiki/NeoWiki/issues/520

PR #651, opened 2026-03-04
Defer schema save until subject is saved in SubjectCreator

alistair3149:
Tested with both manual and automated browser testing.

Fixes https://github.com/ProfessionalWiki/NeoWiki/issues/537
Fixes https://github.com/ProfessionalWiki/NeoWiki/issues/646

## Summary
- When creating a subject with a new schema, the schema is held as a draft and saved atomically with the subject on final save (instead of being saved immediately on "Continue")
- Back button returns to the schema editor with the draft preserved, rather than resetting to the schema selector
- Adds a three-option abandonment dialog (Discard all / Keep editing / Save schema only) when closing with an unsaved draft schema

## Test plan
- [x] Create subject with new schema: verify schema is not saved until subject is saved
- [x] Click back button after Continue: verify schema editor shows with draft preserved
- [x] Close dialog on subject step with draft: verify three-option abandonment dialog
- [x] Close dialog on schema editor step with draft: verify three-option abandonment dialog
- [x] Close dialog without draft: verify standard two-option close confirmation
- [x] "Save schema only" saves schema and closes dialog
- [x] "Discard all" closes without saving
- [x] Continue button is full-width with arrow icon
- [x] Save button shows "Create subject" (not schema name)

https://github.com/user-attachments/assets/1bbfc647-d77f-42fe-93b9-90034951c2ee

### DIFF
--- a/extension.json
+++ b/extension.json
@@ -195,6 +195,7 @@
 						"cdxIconAdd",
 						"cdxIconTrash",
 						"cdxIconArrowPrevious",
+						"cdxIconArrowNext",
 						"cdxIconInfo",
 						"cdxIconEdit",
 						"cdxIconLink",
@@ -273,8 +274,8 @@
 				"neowiki-subject-creator-success",
 				"neowiki-subject-creator-error",
 				"neowiki-subject-creator-create-schema",
+				"neowiki-subject-creator-continue",
 				"neowiki-subject-creator-schema-created",
-				"neowiki-subject-creator-save-with-schema",
 				"neowiki-subject-creator-creating-schema",
 				"neowiki-subject-creator-back",
 				"neowiki-edit-schema",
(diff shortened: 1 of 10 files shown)


JeroenDeDauw:
A glance at this PR made me wonder "does this really need 400 lines. sure, many are tests, but still, we're adding a bunch of complexity"

CC:

> The useCloseConfirmation removal is a step backward. The PR inlines 4 functions (requestClose, confirmClose, cancelClose, plus the ref) that were previously handled by useCloseConfirmation, then adds 4 more (abandonAll, saveSchemaAndClose, cancelSchemaAbandonment, plus the ref). That's 8 functions and 2 refs for close handling alone. The close logic could stay in a composable that's extended to support the draft-schema case, keeping ubjectCreatorDialog leaner.

My product thinking: we seem to be spending time on, and adding code+ui complexity for, a rare scenario. Unclear if that rare scenario justifies these costs. Any notable downside to just saving Schema? If the user abandons creation of the Subject, that leaves a potentially junk schema, but this is better than data loss, can easily be cleaned up, and as it's a presumably rare scenario, seems fine.

PR #722, opened 2026-04-14
Add min/max cross-validation to NumberAttributesEditor

### LINKED ISSUE #476: Polish min max UI in `TextAttributesEditor` and `NumberAttributesEditor`
To follow up to #470

Image

JeroenDeDauw:
> Prompted by @JeroenDeDauw, implemented autonomously by Claude Code.
> Context: the NeoWiki codebase, issue #476 and comment, and TextAttributesEditor as reference.

Fixes https://github.com/ProfessionalWiki/NeoWiki/issues/476

NumberAttributesEditor was missing the validation refinements that
TextAttributesEditor received in PR #551. This adds:

- Error state when minimum exceeds maximum (and vice versa)
- Local input refs with watchers for all fields (minimum, maximum,
  and precision) so in-progress input is preserved even when invalid
- New i18n message `neowiki-property-editor-min-exceeds-max`
- Extracted shared `minExceedsMax` helper used by both Text and Number
  editors
- Moved `FieldProps` test interface to shared `VueTestHelpers`
- CSS classes on min/max fields for robust test selectors

### DIFF
--- a/extension.json
+++ b/extension.json
@@ -284,6 +284,7 @@
 				"neowiki-property-editor-maximum",
 				"neowiki-property-editor-precision",
 				"neowiki-property-editor-precision-non-negative",
+				"neowiki-property-editor-min-exceeds-max",
 				"neowiki-property-editor-character-length",
 				"neowiki-property-editor-length-whole-number",
 				"neowiki-property-editor-length-min-exceeds-max",
--- a/i18n/en.json
+++ b/i18n/en.json
@@ -31,6 +31,7 @@
 	"neowiki-property-editor-maximum": "Maximum",
 	"neowiki-property-editor-precision": "Precision",
 	"neowiki-property-editor-precision-non-negative": "Precision cannot be negative.",
+	"neowiki-property-editor-min-exceeds-max": "Minimum cannot exceed maximum.",
 	"neowiki-property-editor-character-length": "Character length",
 	"neowiki-property-editor-length-whole-number": "Must be a whole number of at least 1.",
 	"neowiki-property-editor-length-min-exceeds-max": "Minimum cannot exceed maximum.",
--- a/i18n/qqq.json
+++ b/i18n/qqq.json
@@ -131,6 +131,7 @@
 	"neowiki-layout-delete-success": "Success notification shown after deleting a layout. $1 is the layout name.",
 	"neowiki-layout-delete-error": "Error notification title shown when deleting a layout fails. $1 is the layout name.",
 	"neowiki-property-type-select": "Label for the select/dropdown property type shown in the property type selector.",
+	"neowiki-property-editor-min-exceeds-max": "Error shown in the schema editor when the minimum value exceeds the maximum value for a number property.",
 	"neowiki-property-editor-options": "Label for the options field in the select property type editor, where schema creators define the allowed values.",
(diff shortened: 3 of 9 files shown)


alistair3149:
> Code review requested by @alistair3149.
> Context: the NeoWiki codebase, PR #722 diff, TextAttributesEditor reference from PR #551, and AGENTS.md conventions.

## Issues

### Important

**1. Precision field still has the exact bug this PR fixes for min/max.**

`resources/ext.neowiki/src/components/SchemaEditor/Property/NumberAttributesEditor.vue:42-47, 78-88`

`updatePrecision` reads `property.precision` directly and never mirrors into a local ref. When invalid input is typed (e.g. `-5`), `precisionError` is set but nothing is emitted — so the parent's `property.precision` never updates, the reactivity of `precisionError` triggers a re-render, and `:model-value` re-evaluates to the old prop value, snapping the display back and losing the user's `-5`.

Suggested fix (~5 lines, matches the min/max pattern just established):

```ts
const precisionInput = ref( props.property.precision?.toString() ?? '' );

watch( () => props.property.precision, ( newVal ) => {
    precisionInput.value = newVal?.toString() ?? '';
} );

const updatePrecision = ( value: string ): void => {
    precisionInput.value = value;
    const numValue = parseNumber( value );
    // ...rest unchanged
};
```

And bind `:model-value="precisionInput"` in the template.

The existing test `'shows error and does not emit when precision is negative'` (`NumberAttributesEditor.spec.ts:151`) only asserts error status and no emit — it starts from `precision: undefined`, so it can't observe the snap-back. A regression test for this specific behavior should be added (and will fail against the current code, confirming the bug):

```ts
it( 'preserves invalid input in the precision field', async () => {
    const wrapper = newWrapper( {
        property: newNumberProperty( { precision: 2 } ),
    } );
    const inputs = wrapper.findAllComponents( CdxTextInput );

    await inputs[ 2 ].vm.$emit( 'update:modelValue', '-5' );

    expect( inputs[ 2 ].props( 'modelValue' ) ).toBe( '-5' );
} );
```

The same pattern is worth adding for min/max — neither the Text nor Number spec currently exercises the watcher that re-syncs the local input when the parent updates the prop externally, and that's the subtlest part of the in-progress-input preservation.

**2. `minExceedsMax` is duplicated byte-for-byte across two components.**

- `resources/ext.neowiki/src/components/SchemaEditor/Property/NumberAttributesEditor.vue:79-83`
- `resources/ext.neowiki/src/components/SchemaEditor/Property/TextAttributesEditor.vue:107-111`

AGENTS.md explicitly says not to duplicate in production code. Now is the right moment to extract to a shared helper — the second copy just arrived.

### Minor

**3. Fragile positional indexing in test helpers.**

`resources/ext.neowiki/tests/components/SchemaEditor/Property/NumberAttributesEditor.spec.ts:32-40`

`fields[length - 3]` / `length - 2` breaks opaquely if another `CdxField` is ever added. Prefer class-based lookup (e.g. `number-attributes__minimum` / `__maximum`) matching the style the precision field already uses.

**4. `FieldProps` interface duplicated between Text and Number specs.** Small test-helper duplication — worth extracting to `VueTestHelpers.ts` for the next editor that needs the pattern.

JeroenDeDauw:
Should be fixed now

PR #727, opened 2026-04-16
Add component tests for DateTime property type

JeroenDeDauw:
Covers DateTimeInput, DateTimeDisplay and DateTimeAttributesEditor,
mirroring the spec files that already exist for the Number, Url, Text
and Relation property types.

Tests that pin the current Z-suffix / timezone-naive behavior carry
`FIXME(#728)` markers so the timezone fix revisits them instead of
preserving the wrong expectations mechanically.

`DateTimeDisplay.spec` asserts against substring invariants (year, time,
date) rather than calling `toLocaleString` in the test itself, so the
expectation is not a copy of the production formatter.

> Result of code review feedback on #685 from @alistair3149 together with @JeroenDeDauw, plus a round of critical review by an Opus subagent that flagged: a misleading "TZ-pinned round-trip" test that was mathematically TZ-independent (removed with a TODO(#728) note), and DateTimeDisplay assertions that duplicated the production formatter (rewritten as substring invariants).
> Context: the NeoWiki codebase and existing component specs for the Number, Url, Text and Relation property types.

### DIFF
--- a/resources/ext.neowiki/tests/components/SchemaEditor/Property/DateTimeAttributesEditor.spec.ts
+++ b/resources/ext.neowiki/tests/components/SchemaEditor/Property/DateTimeAttributesEditor.spec.ts
@@ -1,0 +1,87 @@
+import { DOMWrapper, VueWrapper } from '@vue/test-utils';
+import { describe, expect, it } from 'vitest';
+import DateTimeAttributesEditor from '@/components/SchemaEditor/Property/DateTimeAttributesEditor.vue';
+import { newDateTimeProperty, DateTimeProperty } from '@/domain/propertyTypes/DateTime';
+import { AttributesEditorProps } from '@/components/SchemaEditor/Property/AttributesEditorContract.ts';
+import { createTestWrapper } from '../../../VueTestHelpers.ts';
+
+describe( 'DateTimeAttributesEditor', () => {
+	function newWrapper( props: Partial<AttributesEditorProps<DateTimeProperty>> = {} ): VueWrapper {
+		return createTestWrapper( DateTimeAttributesEditor, {
+			property: newDateTimeProperty( {} ),
+			...props,
+		} );
+	}
+
+	function getInputs( wrapper: VueWrapper ): DOMWrapper<HTMLInputElement>[] {
+		return wrapper.findAll<HTMLInputElement>( 'input[type="datetime-local"]' );
+	}
+
+	describe( 'displaying existing values', () => {
+		// FIXME(#728): pins the current Z-stripping behavior; revisit when the timezone story is decided.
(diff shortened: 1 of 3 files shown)


alistair3149:
> Review requested by @alistair3149. Context: the NeoWiki codebase, the PR diff, the production components under review (`DateTimeDisplay.vue`, `DateTimeInput.vue`, `DateTimeAttributesEditor.vue`, `DateTimeType`), and the sibling Number/Url/Text/Relation specs for convention comparison.

### Important

**Locale brittleness at `DateTimeDisplay.spec.ts:28`** — `toContain('12:00:00')` relies on jsdom's ICU locale for `toLocaleString(undefined, { timeZone: 'UTC' })`. Safe in common locales (`en-US` → `6/15/2025, 12:00:00 PM`, `de-DE` → `15.6.2025, 12:00:00`), but `vitest.config.ts` doesn't pin `LANG`/`LC_ALL`, so a CI host in a locale that formats noon as `12:00:00 AM` would break it. Either pin the locale in vitest config or soften the assertion to `':00:00'` + `'2025'`.

### Minor

- `DateTimeAttributesEditor.spec.ts` skips the `mw` stub that sibling specs set up. Works today because the component uses the `$i18n` mixin via `createTestWrapper`, but worth adding the stub defensively so the spec doesn't silently break if the component starts calling `mw.message` directly.
- `"renders correctly"` at `DateTimeInput.spec.ts:49` bundles five assertions; matches sibling `NumberInput.spec` convention but contradicts the "test one thing per test function" rule. Consider splitting here and in the sibling spec as a follow-up.
- `DateTimeInput`'s `watch(() => props.property, ...)` (re-validation on property change) is untested.
- `DateTimeDisplay` only reads `parts[0]`; no test pins that additional parts are ignored.

malberts:
> Follow-up to @alistair3149's locale-brittleness point, refined in back-and-forth with @malberts.
> Context: the NeoWiki codebase, the PR diff, and analysis of jsdom/Node ICU behavior across locales.

**Locale brittleness at `DateTimeDisplay.spec.ts:28`** — `toContain( '12:00:00' )` depends on jsdom's ICU locale formatting noon with `:` separators and Latin digits. CI risk is effectively zero (GitHub Actions runs `en_US.UTF-8`, and most European locales — en-GB, de-DE, fr-FR — also render `12:00:00`). Real but narrow exposure for contributors on Finnish/Swedish locales (`.` separator → `12.00.00`) or Arabic/Persian locales with Eastern Arabic-Indic digits. Pinning \`LANG\`/\`LC_ALL\` in \`vitest.config.ts\`, or softening to \`':00:00'\` + \`'2025'\`, is a cheap one-line fix — worth doing as defensive hygiene, but not a blocker.

Issue #728, opened 2026-04-16
DateTime: correct timezone handling for Input and AttributesEditor


JeroenDeDauw:
Follows-up to #685 ([review comment](https://github.com/ProfessionalWiki/NeoWiki/pull/685#issuecomment-4156019)).

> Result of extensive back-and-forth with @JeroenDeDauw on splitting the PR #685 review follow-ups into issues.
> Context: the NeoWiki codebase and the review of #685.
> <sub>Written by Claude Code, \`Opus 4.7\`</sub>

\`DateTimeAttributesEditor.vue\` and \`DateTimeInput.vue\` treat the \`datetime-local\` input as UTC:

\`\`\`ts
function toLocalInputValue( isoString ) { return isoString.replace( /Z\$/, '' ).slice( 0, 16 ); }
function fromLocalInputValue( localValue ) { return localValue ? localValue + ':00Z' : undefined; }
\`\`\`

A user in Berlin entering \"2025-06-15 14:00\" has it stored as \`2025-06-15T14:00:00Z\`, which is actually 16:00 Berlin time — data comes back two hours off. ISO strings with explicit offsets (\`...+02:00\`) fail the \`Z\$\` regex, get sliced, and silently re-encoded as UTC. The \`:min\`/\`:max\` bounds on the input use the same broken conversion.

## Requirements

- **Pick one story — UTC-throughout (with a visible \"UTC\" label) or user-local-with-UTC-storage — and apply it consistently across \`DateTimeInput\`, \`DateTimeDisplay\` and \`DateTimeAttributesEditor\`.**
- Round-trip an ISO string with an explicit offset (e.g. \`...+05:00\`) without silent mangling.
- Add a component test that runs under a non-UTC \`TZ\` to pin the behaviour.

(The existing Input/Display/AttributesEditor component tests added in #727 pin *current* behaviour and should be revised alongside the fix.)

malberts:
> Pick one story — UTC-throughout (with a visible "UTC" label) or user-local-with-UTC-storage

I am not sure if these are the only 2 options we have, or if it should be possible to make a datetime input capture and display in a specific timezone. See https://github.com/ProfessionalWiki/NeoWiki/pull/753#issuecomment-4306129041.

PR #734, opened 2026-04-16
DateTime: stricter ISO 8601 value validation

### LINKED ISSUE #729: DateTime: stricter ISO 8601 value validation
Follows-up to #685 ([review comment](https://github.com/ProfessionalWiki/NeoWiki/pull/685#issuecomment-4156019)).

> Result of extensive back-and-forth with @JeroenDeDauw on splitting the PR #685 review follow-ups into issues.
> Context: the NeoWiki codebase and the review of #685.

`DateTime.ts:45` uses `Date.parse` for validation. `Date.parse( '2025' )`, `Date.parse( '2025-06' )`, `Date.parse( '2025-02-30' )` all return valid timestamps in various engines. Timezone-less strings are interpreted as local time per spec, so the min/max comparison mixes local and UTC epochs depending on input. For ECHOLOT's xsd:dateTime round-tripping this matters.

`DateTimeProperty` (PHP, `tests/RedHerb/src/DateTimeProperty.php`) also accepts any `?string` for `default` / `minimum` / `maximum` with no shape check.

## Requirements

- Gate with a regex (or stricter parser) before `Date.parse`. Require stored values to carry an explicit offset or `Z`.
- Add the same shape validation on the PHP side — stored bounds that aren't well-formed xsd:dateTime should be rejected at load/save time (or at least on construction).
- Add an explicit empty-value test to pin that `newStringValue` stripping empties keeps DateTime validation clean (currently implicit): `newStringValue('')` yields `{ parts: [] }` because `Value.ts` filters empties, so `value.parts.length > 0` at `DateTime.ts:43` is false and the empty string never reaches `Date.parse`. DateTime relies on this upstream filter rather than its own check — worth pinning with a test.
- Tighten the bounds tests (`DateTime.spec.ts:68-75` / `:101-108` would pass a trivial `return []` implementation): add "1 ms above min" and "1 ms below max" cases.
- Document min/max inclusivity in qqq or a docblock (tests assert inclusive bounds but the docs are silent; for partners needing xsd:dateTime semantics this matters).

JeroenDeDauw:
Fixes https://github.com/ProfessionalWiki/NeoWiki/issues/729

`DateTimeType.validate` and `DateTimeProperty` (PHP) now gate values
through a strict ISO 8601 regex before `Date.parse` / `checkdate`,
rejecting year-only (`2025`), year-month (`2025-06`), date-only
(`2025-06-15`), timezone-less (`...T12:00:00`) and calendar-overflow
(`2025-02-30T00:00:00Z`) inputs. An explicit offset or `Z` is required
so that min/max comparisons cannot silently mix local and UTC epochs.

The calendar-overflow check reconstructs the declared local date from
the UTC timestamp plus the declared offset and compares it against the
regex-extracted Y/M/D using UTC getters — no host-timezone influence.

PHP rejects malformed \`minimum\` / \`maximum\` / string \`default\` at
construction (including via \`fromPartialJson\`) with \`InvalidArgumentException\`.
TypeScript is fail-open on malformed bounds during validation — a
conscious choice so the UI does not hard-fail on bad server-side data —
documented in the \`DateTimeType\` docblock and pinned with tests.

Also pins year/time-invariants the PR deliberately does not accept
(leap seconds, \`24:00\`, colonless offsets, lowercase \`t\`, expanded
years) in the docblock, and uses a named \`offset\` capture group to
avoid a magic \`match[8]\` index.

As a side effect this also closes the local/UTC epoch-mixing subset of #728
(the one where timezone-less inputs were compared against \`Z\` bounds).
The remaining #728 scope — the \`datetime-local\` input silently treating
user entries as UTC via \`Z$\` stripping — is untouched here.

> Result of extensive back-and-forth with @JeroenDeDauw and a critical review pass by an Opus subagent that flagged a fragile \`match[8]\` magic index (now a named group), missing tests for negative offsets, nanosecond precision, and fail-open-on-malformed-bound behavior, an \`as string\` cast that hid a TypeScript narrowing gap, and an unacknowledged side-effect fix for part of #728 (now documented).
> Context: the NeoWiki codebase, the review of #685, and the strict ISO 8601 / xsd:dateTime shape ECHOLOT partners need for round-tripping.

### DIFF
--- a/i18n/qqq.json
+++ b/i18n/qqq.json
@@ -146,7 +146,7 @@
 	"neowiki-field-single-value-only": "Validation error shown when multiple values are selected for a single-select property.",
 	"neowiki-select-placeholder": "Placeholder text shown in the select dropdown before any option is chosen.",
 	"neowiki-select-no-results": "Message shown in the multi-select dropdown when the typed text does not match any available option.",
-	"neowiki-field-invalid-datetime": "Validation error shown when the entered date and time value cannot be parsed.",
+	"neowiki-field-invalid-datetime": "Validation error shown when the entered date and time value cannot be parsed. DateTime values must be strict ISO 8601 / xsd:dateTime strings with an explicit timezone offset or 'Z' (for example, '2025-06-15T12:00:00Z' or '2025-06-15T12:00:00+02:00'). Partial values such as year-only or date-only are rejected, and min/max bounds are inclusive.",
 	"neowiki-property-type-datetime": "Label for the date and time property type in the schema editor",
 	"neowiki-select-unknown-option": "Placeholder shown in a select-property display when a stored option ID cannot be resolved to a label (e.g. the option was removed from the Schema)."
 }
--- a/resources/ext.neowiki/src/domain/propertyTypes/DateTime.ts
+++ b/resources/ext.neowiki/src/domain/propertyTypes/DateTime.ts
@@ -5,11 +5,85 @@ import { BasePropertyType, ValueValidationError } from '@/domain/PropertyType';
 
 export interface DateTimeProperty extends PropertyDefinition {
 
+	/**
+	 * Inclusive lower bound. Must be a strict ISO 8601 / xsd:dateTime string
+	 * with an explicit timezone offset (e.g. `2025-06-15T12:00:00Z` or
+	 * `2025-06-15T12:00:00+02:00`).
+	 */
 	readonly minimum?: string;
+
+	/**
+	 * Inclusive upper bound. Same shape rules as the minimum.
+	 */
 	readonly maximum?: string;
 
 }
 
+/**
+ * Matches xsd:dateTime-like strings with an explicit timezone offset or `Z`.
+ * A subsequent calendar-overflow check is used to reject inputs like
+ * `2025-02-30T00:00:00Z` that the regex alone cannot detect.
+ */
+// The only quantifier `\d{1,9}` is bounded and followed by a distinct
+// character class, so this is not subject to catastrophic backtracking.
+// eslint-disable-next-line security/detect-unsafe-regex
+const ISO_DATE_TIME_REGEX = /^(-?\d{4})-(0[1-9]|1[0-2])-(0[1-9]|[12]\d|3[01])T([01]\d|2[0-3]):([0-5]\d):([0-5]\d)(?:\.\d{1,9})?(?<offset>Z|[+-](?:[01]\d|2[0-3]):[0-5]\d)$/;
+
+function parseStrictDateTime( value: string ): number | null {
+	const match = ISO_DATE_TIME_REGEX.exec( value );
+	if ( match === null || match.groups === undefined ) {
+		return null;
+	}
+
+	const timestamp = Date.parse( value );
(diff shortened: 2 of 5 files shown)


alistair3149 (commented):
> Dispatched by @alistair3149 as an independent review pass.
> Context: the NeoWiki codebase (TS `DateTimeType`, the RedHerb test fixture, `i18n/qqq.json`), the PR diff, and the PR description/commit message. Verified parity claims by manually running the regexes against the deliberately-rejected forms and by checking `src/Domain/Schema/Property/` for a production PHP DateTime class (none exists — TS `DateTimeType` is the only production path for this type).

## Important

1. **Year-0000 divergence between TS and PHP.** `0000-01-01T00:00:00Z` passes the TS regex and `Date.parse`, and `getUTCFullYear()` returns `0` which matches `Number("0000")`, so TS **accepts** it. PHP `checkdate(1, 1, 0)` returns `false`, so PHP **rejects** it. The PR description and commit message claim the two sides mirror each other; this is one narrow place they don't. Options: tighten the TS regex to require year ≥ 1 (e.g. `([1-9]\d{3}|0[1-9]\d{2}|00[1-9]\d|000[1-9])`), or special-case year 0 as valid in PHP. Cheap either way. Refs: `DateTime.ts:28`, `DateTimeProperty.php:16`.

2. **Docblock claims are untested.** The five "deliberately not accepted" forms in `DateTime.ts:68-73` (lowercase `t`, colonless `+0200`, leap seconds `23:59:60`, `24:00`, expanded `+YYYYYY`) have no negative tests on either side. The regex rejects all five today, but that's only protected by manual reading. Add one-liner cases to `DateTime.spec.ts` (same shape as the existing year-only/date-only tests) and to `DateTimePropertyTest.php::malformedDateTimeProvider`.

## Minor

3. **Mixed named and positional capture access.** `DateTime.ts:45-56` uses `match.groups.offset` but still reaches into `match[1]`, `match[2]`, `match[3]` for Y/M/D. The commit message explicitly calls out naming the offset group to avoid a magic `match[8]`; consistency argues for naming all four (`year`, `month`, `day`, `offset`) or none.

4. **Add a `console.warn` when TS silently ignores a malformed bound.** PHP now rejects malformed bounds at construction, so the only realistic path for the frontend to see one is stored-schema migration from before this PR. Current fail-open silently accepts any value, giving the user no feedback. A one-shot `console.warn` when a bound fails to parse would make drift visible during development without changing the documented fail-open behaviour.

5. **Trim `i18n/qqq.json` doc for `neowiki-field-invalid-datetime`.** The new hint is ~60 words of spec ("partial values such as year-only or date-only are rejected, and min/max bounds are inclusive"). Translators translate the error string, not the ISO 8601 spec. Trim to one sentence about what the user sees; keep the spec in the `DateTimeType` docblock (where it already lives).

## Nit

6. `as DateTimeProperty` cast at `DateTime.ts:107` — pre-existing, not introduced here, but if the "named capture over magic index" discipline is the bar, this is the next thing to pay down.

7. Add a TS round-trip test through `createPropertyDefinitionFromJson` into `validate`. PHP has `testBuildPropertyDefinitionFromJsonViaType`; TS does not.

Issue #726, opened 2026-04-16
Multilingual select options


alistair3149:
Follows-up to https://github.com/ProfessionalWiki/NeoWiki/issues/719

> Written following a design discussion with @alistair3149 about #719.
> Context: NeoWiki codebase, select property type from #711, Schema JSON format.
> <sub>Written by Claude Code, `Opus 4.7`</sub>

## Summary

Select option labels are plain strings in the Schema JSON. With stable IDs landing in
#719, the label becomes presentation data decoupled from stored statement values —
which makes per-language labels feasible without breaking existing data. Worth
designing deliberately rather than retrofitting later.

## Open questions

* **Scope boundary**: just option labels, or the broader problem (Schema names,
  Property Definition labels, descriptions)? Starting narrow is safer but risks
  painting ourselves into a corner.
* **Storage shape**: inline language map on the option (e.g. `label: {en, de, ...}`),
  MediaWiki message keys, Translate extension integration, or something else.
* **Editor UX**: how a Schema editor enters and switches between languages.
* **Fallback rules**: behavior when a label isn't translated for the user's language.

## Related

* #719 (stable IDs — prerequisite for this being clean)
* #711 (select property type)
* https://github.com/ProfessionalWiki/NeoWiki/issues/710

JeroenDeDauw:
Design exploration in #747 proposes bundling this with #719's editor-side follow-up and #718, since all three want the same new option-list UI component. #747 also captures the HW 2026-04-16 preference for convention-over-configuration via MediaWiki system messages as the default direction for this issue.

JeroenDeDauw:
Part of an ECHOLOT requirement: [FR2.10 — support multilingual data and schemas/vocabularies](https://github.com/ECHOLOT-ECCCH/echolot-WP2/issues/27).

Issue #718, opened 2026-04-07
Add color support for select options


JeroenDeDauw:
Follows-up to https://github.com/ProfessionalWiki/NeoWiki/pull/711

> Created by @JeroenDeDauw with Claude Code.
> Context: NeoWiki select property type implementation, competitive analysis of Notion/Airtable/Coda.
> <sub>Written by Claude Code, `Opus 4.6`</sub>

## Summary

The select property type currently stores options as plain strings (e.g. `["Draft", "Review", "Approved"]`). All major competitors (Notion, Airtable, Baserow, NocoDB) support per-option colors, rendering selected values as colored tags/chips.

The current data model is forward-compatible — options could later become objects with `label` + `color` (e.g. `[{"label": "Draft", "color": "blue"}, ...]`) without changing the property type name or storage value type.

## What to consider

- Named colors from a fixed palette (like Notion: `"blue"`, `"green"`, `"red"`) vs hex codes (like NocoDB)
- Auto-assignment of colors when options are created
- Colored rendering in infobox display, subject editor chips, and schema display
- Migration path from plain string options to object options

Related: #719

JeroenDeDauw:
Design exploration in #747 proposes bundling this with #719's editor-side follow-up and #726, since all three want the same new option-list UI component.

krabina:
Probably this should also support bootstrap colors, since most third party wiki use Chameleon skin?
https://getbootstrap.com/docs/5.3/customize/color/
So you can set "primary", "secondary", "warning", "danger", "success" etc. and get the colors that are defined in your bootstrap theme.

PR #740, opened 2026-04-17
Add stable IDs for select options

### LINKED ISSUE #719: Add stable IDs for select options
Follows-up to https://github.com/ProfessionalWiki/NeoWiki/pull/711

> Created by @JeroenDeDauw with Claude Code.
> Context: NeoWiki select property type implementation, competitive analysis of Notion/Airtable/Coda.

## Summary

Select options are currently identified by their display string. This means renaming an option (e.g. "Draft" to "First draft") breaks the link to all existing subjects that use that value.

Notion, Airtable, and Baserow solve this by giving each option a stable ID separate from the display name. Statements then reference the ID, and renaming only changes the display label.

## What to consider

- Options would become objects: `{"id": "opt_abc123", "label": "Draft"}` instead of plain `"Draft"`
- Statements would store the option ID as value instead of the label string
- Display layer resolves IDs to labels at render time
- Migration path from string-based options
- Whether this should be tackled together with or separately from color support (#718)

Related: #718

alistair3149:
Fixes https://github.com/ProfessionalWiki/NeoWiki/issues/719

> Result of extensive back-and-forth with @alistair3149 — design discussion, implementation strategy,
> and a mid-PR scope reduction (tombstone feature deferred as YAGNI).
> Context: the NeoWiki codebase, issue #719, competitive comparison of Notion/Airtable/Baserow,
> browser smoke testing against the local dev wiki.

Tested modifying the existing select options in the editor.
There are no visible regression.

## Summary

Migrates select-property options from `string[]` to `{id, label}[]`. Stored statement values reference the
option `id`, so renaming a label no longer breaks existing statements. The write API accepts either an `id`
or a `label` and resolves to the canonical id server-side.

- `SelectOption` value objects in both PHP and TypeScript domain layers
- Statement values store option IDs (not labels)
- `SelectDisplay` resolves IDs to labels; unknown IDs show `(unknown option)`
- Write path accepts `id`, `label`, or `{id, label}` object; mismatched `{id, label}` rejected; object with
  `id` only requires a matching option (no label-fallback)
- No migration from the legacy string-option format: loading a legacy schema throws `InvalidArgumentException`

## Known limitation — schema editor UI

`SelectAttributesEditor.vue` is unchanged in this PR. It still rebuilds the entire options array from chip
labels on save, using the label as the id (`{id: label, label}`). That means saving a schema through the UI
editor after it was edited via the ID-aware API (or by hand through source) will rewrite every option's id
to its current label, orphaning any existing statements.

This is left intentionally out of scope because it requires some UI design work on the options input in the SchemaEditor.
When we work on anything that edits "properties" of an option (e.g. #718, #726), then we will need the UI.

## Design notes

- `StatementListPatcher` stays pure. Resolution lives in two new pure readonly services — `SelectValueResolver`
  (single value) and `SelectPatchResolver` (patch walker) — invoked from the Action layer where
  `SchemaLookup` is already in hand.
- When `SchemaLookup` returns `null` the patch passes through unresolved (lenient), matching the pre-PR
  behaviour. Follow-up-worthy for observability but not urgent.
- Label uniqueness is case-insensitive and whitespace-trimmed, global across all options. Tombstoning was
  scoped out; there is no live-only distinction.
- Translatable option labels are tracked separately in #726.

## Test plan

- [x] `make phpunit filter='Domain.*Schema'` — 49/49
- [x] `make phpunit filter='Application'` — 79/79 (1 pre-existing skip)
- [x] `make phpunit filter='SchemaContent'` — 9/9
- [x] `make tsci` — lint + build + Vitest green
- [x] Browser: subject display resolves IDs to labels; `(unknown option)` fallback renders when a stored
      value doesn't match any current option
- [x] Browser: subject edit picker shows labels, writes IDs (verified the stored JSON carries IDs after save)
- [x] REST: `PATCH /rest.php/neowiki/v0/subject/{id}` with `value: "Active"` (a label) resolves to the
      option ID before storing
- [x] Full `make phpunit` on CI (Neo4j-backed integration suite doesn't reliably run in the local env)

### DIFF
--- a/DemoData/Schema/Company.json
+++ b/DemoData/Schema/Company.json
@@ -20,7 +20,12 @@
 		},
 		"Status": {
 			"type": "select",
-			"options": ["Active", "Inactive", "Acquired", "Dissolved"],
+			"options": [
+				{ "id": "o1demo1aaaaaaa1", "label": "Active" },
+				{ "id": "o1demo1aaaaaaa2", "label": "Inactive" },
+				{ "id": "o1demo1aaaaaaa3", "label": "Acquired" },
+				{ "id": "o1demo1aaaaaaa4", "label": "Dissolved" }
+			],
 			"required": true
 		},
 		"World domination progress": {
--- a/DemoData/Schema/Everything.json
+++ b/DemoData/Schema/Everything.json
@@ -42,16 +42,31 @@
 		},
 		"select/select": {
 			"type": "select",
-			"options": ["Draft", "Review", "Approved", "Archived"]
+			"options": [
+				{ "id": "o1demo1aaaaaab1", "label": "Draft" },
+				{ "id": "o1demo1aaaaaab2", "label": "Review" },
+				{ "id": "o1demo1aaaaaab3", "label": "Approved" },
+				{ "id": "o1demo1aaaaaab4", "label": "Archived" }
+			]
 		},
 		"select/select (required)": {
 			"type": "select",
-			"options": ["Low", "Medium", "High", "Critical"],
+			"options": [
+				{ "id": "o1demo1aaaaaac1", "label": "Low" },
+				{ "id": "o1demo1aaaaaac2", "label": "Medium" },
+				{ "id": "o1demo1aaaaaac3", "label": "High" },
+				{ "id": "o1demo1aaaaaac4", "label": "Critical" }
+			],
 			"required": true
 		},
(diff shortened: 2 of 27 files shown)


JeroenDeDauw:
Follow-up design issue opened for the schema editor work (stable IDs, colors, translations): #747

Issue #647, opened 2026-03-03
Create rebuild script


JeroenDeDauw:
MediaWiki maintenance script that rebuilds the secondary persistence (graph DBs) by reading the primary persistence (JSON in MediaWiki revisions).

AC
* Maintenance script exists and saves all Subjects to the graph db
* Script provides reasonable output, details TBD by implementer

Ideally, we can resume the script. Perhaps with `--resume-from` parameter, which could take pageid if we go though the pages sprted by id. Ideally the user can also abort the script (i.e. ctrl+c) and then get resume instructions.

To consider: what if there is outdated info in the graph db, like Subjects that no longer exist? Should the script have a flag to first delete everything? Or better to keep this responsibility out of this script? The admin could first clear the graph db, or we could have another script that looks for junk entries in the graph db and clears them.

## Relevant code/docs

* https://github.com/ProfessionalWiki/NeoWiki/blob/master/docs/adr/019_Graph_Database_Architecture.md
* https://github.com/ProfessionalWiki/NeoWiki/pull/590

## Implementation thoughts

Since we store Subjects in a dedicated MW slot, there likley is an efficient way to find all pages with our slot present for the last revision.

Iterators are nice, perhaps a SubjectIterator is called for.

Decoupling from the presentation mechanism (CLI) could be nice, so creating a UseCase.

We have examples of decoupling from CLI and using such Iterators in our existing code, though idk which projects off the top of my head.

Supporting ctrl+c might be tricky with MediaWiki maintenance scripts. Could consider building with Symfony Console or similar like I did in https://github.com/JeroenDeDauw/Replicator? ctrl+c handling there needed https://github.com/JeroenDeDauw/Replicator/blob/master/composer.json#L17. This app uses https://github.com/JeroenDeDauw/JsonDumpReader which has some iterator stuff in there, but that's really old, still the use case is similar. Using Symfony Console is only possible if we manage to still interface with MediaWiki reliably.

PR #744, opened 2026-04-18
Reindex subject pages in Neo4j after demo data import

JeroenDeDauw:
> Result of investigation by @JeroenDeDauw with Claude Code, after observing that
> related-subject infobox links rendered as raw IDs after running `make import-demo-data`.
> Context: NeoWiki codebase, Neo4j cypher inspection of the local dev wiki.

## Summary

Re-fires the revision-created handler for every imported subject page so the Neo4j
index matches the imported state.

## Why

The standard import path only triggers indexing when a new revision is actually
created (the `RevisionFromEditComplete` hook). When `make import-demo-data` is
re-run after Neo4j has been wiped — common on dev environments after `docker
compose down --volumes` — pages whose content didn't change report
`done (no changes)` and Neo4j is never updated for them.

Worse: when an unrelated changed page (e.g. `Professional_Wiki`) has a relation
to one of those unchanged pages, `SubjectRelationUpdater` runs
`MERGE (target {id: \$targetId})` which creates an **unlabeled stub node** for the
target. The subject is now half-present in Neo4j: the node exists with `id` set,
but has no `:Subject` label and no `:HasSubject` edge from any Page. The
`Neo4jPageIdentifiersLookup` query (`MATCH (page:Page)-[:HasSubject]->(subject {id: \$id})`)
returns no match, so `GET /rest.php/neowiki/v0/subject/{id}` returns
\`{\"subject\":null}\` and the Vue `RelationDisplay` renders the raw ID.

Reproduced locally: `s1demo4sssssss1` (NeoWiki) and `s1demo6sssssss1` (ProWiki)
were stub nodes referenced from `Professional_Wiki`'s Products relation. After
the fix, both resolve correctly.

## Test plan

- [x] Manually deleted the stub nodes from Neo4j, ran `make import-demo-data`,
      confirmed `GET /rest.php/neowiki/v0/subject/s1demo4sssssss1` and
      `s1demo6sssssss1` now return the subjects, and the `Professional_Wiki`
      page renders relation links correctly.
- [ ] CI green

### DIFF
--- a/maintenance/ImportDemoData.php
+++ b/maintenance/ImportDemoData.php
@@ -7,6 +7,7 @@ namespace ProfessionalWiki\NeoWiki\Maintenance;
 use FileFetcher\SimpleFileFetcher;
 use Maintenance;
 use MediaWiki\MediaWikiServices;
+use MediaWiki\Title\Title;
 use ProfessionalWiki\NeoWiki\Application\Actions\ImportPages\ImportPagesAction;
 use ProfessionalWiki\NeoWiki\Application\Actions\ImportPages\ImportPresenter;
 use ProfessionalWiki\NeoWiki\Application\Actions\ImportPages\PageContentSource;
@@ -32,6 +33,7 @@ class ImportDemoData extends Maintenance {
 
 	public function execute(): void {
 		$this->newImportAction()->import();
+		$this->reindexSubjectPages();
 	}
 
 	private function newImportAction(): ImportPagesAction {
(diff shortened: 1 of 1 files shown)


JeroenDeDauw:
Unsure if this is the best approach, but the complexity is added to the demo data import script, so :man_shrugging: , at worst some prudent tech debt.

JeroenDeDauw:
Let's create a neo4j rebuild script instead

JeroenDeDauw:
Superseded by #746 (and ProfessionalWiki/NeoWiki-Docker#111), which moves the fix into a dedicated RebuildGraphDatabases maintenance script — the right level of abstraction per ADR 019 and the existing #647 issue.

PR #746, opened 2026-04-18
Add RebuildGraphDatabases maintenance script (MVP)

JeroenDeDauw:
My idea, as an alternative fix to 
* https://github.com/ProfessionalWiki/NeoWiki/pull/744

First part of
* https://github.com/ProfessionalWiki/NeoWiki/issues/647

Manually tested

> Result of debugging session with @JeroenDeDauw after stub Neo4j nodes broke
> relation rendering on the local dev wiki. Context: NeoWiki codebase, ADR 019
> (Graph Database Architecture), Neo4j cypher inspection.

## Summary

Adds a `RebuildGraphDatabases` maintenance script that re-saves every Subject
from the latest revision of its page through the `GraphDatabasePlugin` chain.
Useful after a graph database has been wiped or otherwise drifted from the
MediaWiki source of truth.

## Scope

This is an **MVP for #647**. It satisfies the first AC (script exists and saves
all Subjects to the graph DB) and provides per-page progress output. It deliberately
does **not** implement:

- `--resume-from` parameter
- Ctrl+C handling with resume instructions
- Cleanup of stale entries / `--delete-first` flag
- `SubjectIterator` abstraction
- Use case + presenter decoupling

Those are deferred. Issue #647 stays open for them.

## Why now

`make import-demo-data` re-runs after a Neo4j wipe leave subjects on unchanged
pages unindexed (the `RevisionFromEditComplete` hook only fires when a new
revision is actually created). Inbound relations from re-saved pages then
create unlabeled stub nodes via `MERGE (target {id: \$targetId})` in
`SubjectRelationUpdater`, which the API can't resolve. The companion
NeoWiki-Docker PR chains this script into `make import-demo-data` so the dev
flow self-heals.

## How it works

1. Queries `slots` for every page whose latest revision carries the NeoWiki
   subject slot (`slot_role_id = id of "neo"`), ordered by `page_id`.
2. For each, loads the current revision and fires
   `OnRevisionCreatedHandler::onRevisionCreated()` — the same path the hook
   takes on a real edit.
3. Outputs `Rebuilt <Title>` per page, plus a totals line at the end.

The handler is idempotent: pages already in sync get the same write, no harm.

## Test plan

- [x] Manually deleted `s1demo4sssssss1` and `s1demo6sssssss1` stub nodes from
      Neo4j, ran the script directly, confirmed both subjects resolve via the
      REST API and the `Professional_Wiki` infobox shows linked product names.
- [x] Re-ran with state already in sync; script reports `Rebuilt 26 of 26`
      with no errors.
- [x] Verified via the NeoWiki-Docker companion PR that
      `make import-demo-data` now chains the rebuild and self-heals.
- [x] CI green

### DIFF
--- a/maintenance/RebuildGraphDatabases.php
+++ b/maintenance/RebuildGraphDatabases.php
@@ -1,0 +1,104 @@
+<?php
+
+declare( strict_types = 1 );
+
+namespace ProfessionalWiki\NeoWiki\Maintenance;
+
+use Maintenance;
+use MediaWiki\MediaWikiServices;
+use MediaWiki\Page\WikiPageFactory;
+use MediaWiki\Title\Title;
+use MediaWiki\User\UserIdentity;
+use ProfessionalWiki\NeoWiki\EntryPoints\OnRevisionCreatedHandler;
+use ProfessionalWiki\NeoWiki\NeoWikiExtension;
+use ProfessionalWiki\NeoWiki\Persistence\MediaWiki\Subject\MediaWikiSubjectRepository;
+use User;
+
+$basePath = getenv( 'MW_INSTALL_PATH' ) !== false ? getenv( 'MW_INSTALL_PATH' ) : __DIR__ . '/../../..';
+
+require_once $basePath . '/maintenance/Maintenance.php';
+
+class RebuildGraphDatabases extends Maintenance {
+
+	public function __construct() {
+		parent::__construct();
+
+		$this->requireExtension( 'NeoWiki' );
+		$this->addDescription(
+			'Rebuilds the graph databases by re-saving every Subject from the latest revision of its page. ' .
+			'Useful after a graph database has been wiped or has otherwise drifted from the MediaWiki source of truth.'
+		);
+	}
+
+	public function execute(): void {
+		$pageIds = $this->getSubjectPageIds();
+
+		$this->outputChanneled( 'Rebuilding graph databases for ' . count( $pageIds ) . ' subject pages...' );
+
+		$handler = NeoWikiExtension::getInstance()->getStoreContentUC();
+		$wikiPageFactory = MediaWikiServices::getInstance()->getWikiPageFactory();
+		$user = $this->getMaintenanceUser();
+
+		$rebuilt = 0;
+
+		foreach ( $pageIds as $pageId ) {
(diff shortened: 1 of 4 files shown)


alistair3149:
> Reviewed via back-and-forth with @alistair3149 against the MVP scope stated in the PR description.
> Context: NeoWiki codebase, AGENTS.md coding standards, MediaWiki core `Maintenance`/`NameTableStore`.

### Blocker

**`lastEditor` is overwritten with the `NeoWiki` system user on every rebuilt page.** `RebuildGraphDatabases.php:71` passes `$user` into `OnRevisionCreatedHandler::onRevisionCreated()`, which threads it to `PagePropertiesBuilder::getPagePropertiesFor()` and writes it to the graph DB as `lastEditor`. So running the rebuild silently corrupts persistent state — not a "rerun it" failure; the wrong value sticks until a real edit comes in on that page. Fix: pass `$revision->getUser()` so the rebuild preserves each revision's actual author.

### Follow-up candidates (not blocking this MVP)

- **Fresh-wiki crash.** `SlotRoleStore::getId( 'neo' )` at line 83 throws `NameTableAccessException` if no subject has ever been saved. Irrelevant for the recover-from-drift use case; catch and treat as 0 pages if we ever want to run this on empty wikis.
- **Abort on first failure.** Any exception during a per-page rebuild kills the loop with no indication of which page failed. Wrap in `try { } catch ( \Exception $e )`, log the title, continue, report `Rebuilt X of Y (Z failed)`. Pairs naturally with the already-deferred `--resume-from` work.
- **`--quiet` doesn't work.** `outputChanneled()` ignores `$mQuiet`; switch the five call sites to `$this->output( "...\n" )`.
- **Non-namespaced imports.** `use Maintenance;` / `use User;` are class_alias shims; AGENTS.md requires `MediaWiki\Maintenance\Maintenance` / `MediaWiki\User\User`. (Same pre-existing debt in `ImportDemoData.php`.)
- **Testability seam.** `execute()` fetches the handler and `WikiPageFactory` straight off the service locator. Adding a protected `newHandler()` / `newWikiPageFactory()` override point would make the outer loop testable without a wider refactor.

JeroenDeDauw:
@alistair3149 solved

Issue #750, opened 2026-04-20
Create subject managment UI


JeroenDeDauw:
We have a Subject creation flow ([merged PR](https://github.com/ProfessionalWiki/NeoWiki/issues/427)), however, we do not have a way to start this flow beyond the temporary "main subject creation button" hack. Users also lack the ability to delete subjects, to change the main subject, or to add non-main subjects.

We previously explored page schemas, blocks, and similar (I cannot find these conversations; please reference them here if you find them). There is a lot of product and implementation work left. Luckily, I don't think we need to do any of that for the MVP. We don't need to solve display concerns, as we already have the `#view` parser function, which is good enough for MVP. We just need to solve the use cases listed above. 

## Capabilities

(This is a draft, might be incomplete)

Ordered by importance

* Add new subject (opens the creation flow)
* View subjects
* Delete subject
* Change main subject
* Choose if the subject is the main subject while adding it

alistair3149:
There are some relevant info about consolidating Vue entrypoints into the same special page as a SPA: https://github.com/ProfessionalWiki/NeoWiki/issues/525#issuecomment-3849486479

alistair3149:
> From an ongoing design discussion with @alistair3149 on the subject management UI.
> Context: the NeoWiki codebase, issue #750, the tools-menu pattern introduced in PR #745, the Glossary, and `PageSubjects` domain behavior.
> <sub>Written by Claude Code, `Opus 4.7 (1M context)`</sub>

Two semantic questions surfaced while designing the MVP. Both likely resolve together.

## Open question 1: should promoting a child to main rewrite the subject's label?

The [Glossary](https://github.com/ProfessionalWiki/NeoWiki/blob/master/docs/Glossary.md#subject) marks as TODO that a main subject's label must match the page title. The rule isn't enforced in code today. Before "Change main subject" can ship, the promote flow needs a defined behavior:

1. **Don't enforce.** Flip the main flag; keep the promoted subject's existing label, even if it differs from the page title.
2. **Rewrite on promote.** Overwrite the promoted subject's label with the page title. The original label is lost; warn before applying.
3. **Block with guidance.** Refuse to promote until labels match. User renames the subject (or the page) first.

Also: is this rule an invariant enforced everywhere, or a convention applied only at promote time (and not during, say, imports)?

Proceeding with MVP for View, Add, and Delete. Pausing "Change main subject" until this is decided.

## Open question 2: can a page have child subjects without a main subject?

[...]

PR #751, opened 2026-04-20
Add page subjects manifest dialog

alistair3149:
For https://github.com/ProfessionalWiki/NeoWiki/issues/750

https://github.com/user-attachments/assets/0e7f307b-cee2-400e-96ed-0dbc9d601493

Adds a read-only "Manage subjects" entry to the NeoWiki tools-menu section
that opens a modal listing every subject stored on the current wiki page --
including child subjects that have no `#view` invocation and are therefore
invisible on the page itself.

## Changes

- New public REST endpoint `GET /neowiki/v0/page/{pageId}/subjects`
  returning a paginated summary (`id`, `label`, `schema`, `isMain`) with
  main-first then children by label ascending (case-insensitive). Matches
  the param bounds of `GetSchemaSummariesApi` (limit 1-50 default 10,
  offset >= 0 default 0).
- New `PageSubjectsDialog` Vue component rendering the manifest in a
  `CdxDialog` + `CdxTable`. Inline "Main" chip on the main row with an
  aria-label "(Main subject)" for screen-reader clarity.
- New "Manage subjects" tools-menu entry, shown only when the page has at
  least one subject. Opens the dialog via a shared Pinia flag and a
  document-level click listener (same pattern as the existing "Create
  subject" entry from PR #745).

## Why a modal, not a dedicated page

Considered `?action=neowiki-subjects` on the page (mirroring `?action=history`)
and a `Special:NeoWiki/Subjects/{PageName}` subpage (mirroring the existing
`Special:NeoJson`). Picked an in-place modal because:

- Subject management is an inspection task. A full page load disrupts the
  user's flow on the page they are already looking at, and the MediaWiki
  skin chrome would wrap a short summary table in a lot of unrelated UI.
- A shareable URL for the manifest itself has no strong use case. The page
  URL already identifies the data, and cross-cutting subject lookup is
  explicitly intended to go through Cypher queries and `#view`.

## Manual Browser Check

With any content page that has a main subject (e.g. `ACME_Inc`):

1. Open the NeoWiki section of the tools menu -- "Manage subjects" is listed.
2. Click it -- modal opens titled "Subjects on ACME Inc." listing the main
   subject with an inline "(Main subject)" chip next to its label.
3. Columns render: Subject, ID (monospace), Schema (linked to `Schema:<name>`).
4. Close via X / backdrop / Esc -- returns to the page.
5. On a fresh page with no subjects, "Manage subjects" is absent; the
   "Create subject" entry still shows.


src/EntryPoints/REST/GetPageSubjectsApi.php
```diff
@@ -0,0 +1,145 @@
...
+ *   - limit:  integer, default 10, min 1, max 50
+ *   - offset: integer, default 0, min 0
+ *
+ * Sort order (public API contract):
+ *   1. The page's main subject (if present).
+ *   2. Child subjects by label ascending, case-insensitive.
+ */
+class GetPageSubjectsApi extends SimpleHandler {
+
+	public function run( int $pageId ): Response {
```
alistair3149, line 35: I use Page ID because we use it for `POST /neowiki/v0/page/{pageId}/mainSubject` and `POST /neowiki/v0/page/{pageId}/childSubjects`. That deviates from the [MW REST API](https://www.mediawiki.org/wiki/API:REST_API/Reference#Get_page) where page title is used instead. I think this might be worth noting.

tests/phpunit/EntryPoints/REST/GetPageSubjectsApiTest.php
```diff
@@ -0,0 +1,168 @@
...
+		$this->assertSame( 2, $data['totalRows'] );
+		$this->assertCount( 2, $data['subjects'] );
+		$this->assertTrue( $data['subjects'][0]['isMain'] );
+		$this->assertSame( 's1zz1111111aaa1', $data['subjects'][0]['id'] );
+		$this->assertSame( 'ACME Inc.', $data['subjects'][0]['label'] );
+		$this->assertSame( 'Company', $data['subjects'][0]['schema'] );
+		$this->assertFalse( $data['subjects'][1]['isMain'] );
+		$this->assertSame( 's1zz1111111aaa2', $data['subjects'][1]['id'] );
+		$this->assertSame( 'Widget X', $data['subjects'][1]['label'] );
+		$this->assertSame( 'Product', $data['subjects'][1]['schema'] );
```
JeroenDeDauw: Testing the full JSON with a multiline string block might be better

JeroenDeDauw: Certainly more readable, and provides an example of the API response format

JeroenDeDauw:
The new API is interesting, and perhaps we want it regardless from if we go with this exact UI approach.

tests/phpunit/EntryPoints/REST/GetPageSubjectsApiTest.php
alistair3149: Good catch, done :)

JeroenDeDauw:
A tab like I described on https://github.com/ProfessionalWiki/NeoWiki/issues/750 seems better. Unclear to me if there is also room for a dialog or what it then should contain. Let's discuss this together on the issue, exploring the design space holistically. Leaving this comment here mainly to make it explicit that I think this should not be merged and that further design conversation is needed before we add a dialog.

alistair3149:
Superseded by #755

PR #753, opened 2026-04-20
DateTime: user-local display/input with UTC storage and semantic <time>

### LINKED ISSUE #728: DateTime: correct timezone handling for Input and AttributesEditor

### LINKED ISSUE #731: DateTime: semantic <time> rendering, UTC indicator, Codex CdxTextInput
Follows-up to #685 ([review comment](https://github.com/ProfessionalWiki/NeoWiki/pull/685#issuecomment-4156019)).

> Result of extensive back-and-forth with @JeroenDeDauw on splitting the PR #685 review follow-ups into issues.
> Context: the NeoWiki codebase and the review of #685.

Three presentation-layer improvements identified in the review of #685:

## Use \`<time>\` for semantic rendering

\`DateTimeDisplay.vue:2-4\` renders into a \`<div>\`. Switch to:
\`\`\`vue
<time :datetime=\"dateString\">{{ formattedValue }}</time>
\`\`\`
Keeps the canonical ISO value machine-readable in the DOM while showing a formatted string to users. Aligns with ECHOLOT's linked-data goals (CIDOC-CRM, xsd:dateTime, RDF export) and lets assistive tech announce it as a time. Only emit the \`datetime\` attribute when the string parses as valid ISO 8601; on the invalid fallback, render a plain \`<span>\`.

## Make the UTC timezone visible

\`DateTimeDisplay.vue:30\`: formatted via \`toLocaleString( undefined, { timeZone: 'UTC' } )\`, so \"2025-06-15, 12:00:00 PM\" looks like local time with no indication. Either add a visible \`UTC\` suffix, format as ISO, or use the user's timezone. Whatever the choice, make it visible. Depends on the resolution of #728.

## Use \`CdxTextInput input-type=\"datetime-local\"\` in the inputs

\`DateTimeInput.vue\` and \`DateTimeAttributesEditor.vue\` use raw \`<input>\` instead of Codex's wrapper. Loses focus ring, validation states, and the deliberately registered \`cdxIconClock\` start-icon never appears. Three \`eslint-disable vue/html-self-closing\` comments hint at the same friction. \`CdxTextInput\` supports \`datetime-local\` as an \`input-type\`.

alistair3149:
Fixes https://github.com/ProfessionalWiki/NeoWiki/issues/728
Fixes https://github.com/ProfessionalWiki/NeoWiki/issues/731

## Summary

Three DateTime components now agree on one story: **user-local wall-clock in the UI, UTC ISO on disk**.

- **New shared utility** `dateTimeConversion.ts` (`toLocalInputValue`, `fromLocalInputValue`) uses `new Date()` + `.toISOString()` so explicit-offset ISOs (e.g. `2025-06-15T23:30:00+05:00`) round-trip without silent re-encoding.
- **`DateTimeDisplay`** renders valid ISOs inside a semantic `<time :datetime="iso">` element with the user's local wall-clock plus a visible timezone abbreviation (`UTC`, `CEST`, `PST`, ...). Invalid values render a `<span>` fallback. Validity uses the newly-exported `parseStrictDateTime` from `DateTime.ts`.
- **`DateTimeInput`** and **`DateTimeAttributesEditor`** replace raw `<input type="datetime-local">` with `CdxTextInput input-type="datetime-local"`. The Input gets a `cdxIconClock` start-icon (previously registered but never rendered). Two `eslint-disable vue/html-self-closing` comments retired.
- Tests in all three component specs rewritten around the new behavior; `FIXME(#728)` pins that locked the old buggy behavior are gone. Direct tests added for `parseStrictDateTime` since it is now part of the public module surface.

## Manual Browser Check

1. **Subject edit flow** — open a page that has a DateTime property and edit it in the subject editor.
   - Input shows a Codex clock icon on the left.
   - Typing `2025-06-15 14:00` saves as your local 14:00. Reload — the input should still show `14:00` in your host TZ.
   - If the schema declares `minimum`/`maximum`, the `min`/`max` attributes on the input reflect them in your local TZ.
2. **Subject display flow** — view the same page without editing.
   - Inspect the value: it must be a `<time>` element with a `datetime` attribute equal to the stored UTC ISO.
   - Rendered text must include a timezone abbreviation after the time (e.g. `UTC`, `CEST`, `PST`).
   - An unparseable stored value falls back to a `<span>` showing the raw string.
3. **Schema editor flow** — open the schema editor for a DateTime property.
   - Both min/max inputs are Codex-styled (focus ring, proper sizing), not bare browser inputs.
   - Entered values round-trip as the same local wall-clock after saving and reopening.

Suggested starting URL: `http://localhost:8484/index.php/Main_Page` (or any page with DateTime data).

### DIFF
--- a/resources/ext.neowiki/src/components/SchemaEditor/Property/DateTimeAttributesEditor.vue
+++ b/resources/ext.neowiki/src/components/SchemaEditor/Property/DateTimeAttributesEditor.vue
@@ -66,8 +66,13 @@ watch( () => props.property.maximum, ( newVal ) => {
 
 // datetime-local wire values are always `YYYY-MM-DDTHH:mm` in a single
 // timezone (host local), so lexicographic ordering matches chronological.
+// The regex guards against malformed values bypassing the ordering check.
+const DATETIME_LOCAL_WIRE_FORMAT = /^\d{4}-\d{2}-\d{2}T\d{2}:\d{2}$/;
+
 function minExceedsMax( min: string, max: string ): boolean {
-	return min !== '' && max !== '' && min > max;
+	return DATETIME_LOCAL_WIRE_FORMAT.test( min ) &&
+		DATETIME_LOCAL_WIRE_FORMAT.test( max ) &&
+		min > max;
 }
 
 const updateMinimum = ( value: string ): void => {
--- a/resources/ext.neowiki/src/components/SchemaEditor/Property/DateTimeAttributesEditor.vue
+++ b/resources/ext.neowiki/src/components/SchemaEditor/Property/DateTimeAttributesEditor.vue
@@ -64,13 +64,10 @@ watch( () => props.property.maximum, ( newVal ) => {
 	maximumInput.value = toLocalInputValue( newVal );
 } );
 
+// datetime-local wire values are always `YYYY-MM-DDTHH:mm` in a single
+// timezone (host local), so lexicographic ordering matches chronological.
 function minExceedsMax( min: string, max: string ): boolean {
-	if ( min === '' || max === '' ) {
-		return false;
-	}
-	const minTime = Date.parse( min );
-	const maxTime = Date.parse( max );
-	return !Number.isNaN( minTime ) && !Number.isNaN( maxTime ) && minTime > maxTime;
+	return min !== '' && max !== '' && min > max;
 }
 
 const updateMinimum = ( value: string ): void => {
--- a/resources/ext.neowiki/src/components/SchemaEditor/Property/DateTimeAttributesEditor.vue
+++ b/resources/ext.neowiki/src/components/SchemaEditor/Property/DateTimeAttributesEditor.vue
@@ -14,13 +14,11 @@
 					{{ $i18n( 'neowiki-property-editor-minimum' ).text() }}
 				</template>
 
-				<!-- eslint-disable-next-line vue/html-self-closing -->
-				<input
-					type="datetime-local"
-					class="cdx-text-input__input"
-					:value="minimumInput"
-					@input="updateMinimum"
-				>
+				<CdxTextInput
+					input-type="datetime-local"
+					:model-value="minimumInput"
+					@update:model-value="updateMinimum"
+				/>
 			</CdxField>
 
 			<CdxField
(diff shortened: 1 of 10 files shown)


malberts:
> user-local wall-clock in the UI

I am not 100% sure if this is always what we want, but likely fine to start with.

Specifically, if the timezone is an explicit part of the data (e.g. an event in Germany), then perhaps you want it to be input as CEST, and display it as CEST, irrespective of the user's locale. But it is unclear to me if this is a common, or necessary, use case.

alistair3149:
We can follow up with timezone input if needed. The display part likely goes into the display attributes of the DateTime property, which is a separate concern.

Issue #544, opened 2026-02-05
savePage() DETACH DELETEs subjects without checking incoming relations


JeroenDeDauw:
This task was generated by caude based on our issue history and has not been human-vetted yet. We should verify this is about behavior we want.

## Problem

`Neo4jQueryStore::savePage()` (line 44 of `src/Persistence/Neo4j/Neo4jQueryStore.php`) uses `DETACH DELETE subject` to remove subjects that are no longer present on a page. This silently deletes subjects that have incoming relations from other subjects, breaking those relations.

In contrast, `deleteSubject()` (line 128) correctly checks `subjectHasIncomingRelations()` before deciding whether to detach-delete or just remove the HasSubject relation. The `savePage()` code path does not use this check.

```cypher
// Delete subjects that are no longer present on the page
WITH page
MATCH (page)-[r:HasSubject]->(subject)
WHERE NOT subject.id IN $subjectIds
DETACH DELETE subject
```

## Expected behavior

`savePage()` should reuse the `deleteSubject()` logic (or equivalent) that checks for incoming relations before deciding how to handle removal. Subjects with incoming relations should only have their HasSubject relation removed, not the subject node itself.

## Origin

First flagged by Copilot code review on PR #458 (Jan 2026), which fixed the same problem in `deletePage()` but left `savePage()` unchanged. Verified on 2025-02-05 that the `DETACH DELETE` without incoming-relation check is still present on master at `Neo4jQueryStore.php:44`.

PR #755, opened 2026-04-22
Add Subject Management page action

JeroenDeDauw:
For https://github.com/ProfessionalWiki/NeoWiki/issues/750. Caters to all UC except "Choose if the subject is the main subject while adding it", which seems fine to omit, and actually unclear if it'd be net-positive to add special handling for it.

image

---

Note: this PR is one of two complementary approaches under discussion for #750. The other (#751) adds a read-only manifest dialog from the tools menu.

## What this adds

A page action `?action=subjects` that mounts a Vue management view on any wiki content page, plus a "Manage subjects" link in the NeoWiki tools-menu section (next to "Create subject") so the action is reachable without typing URL parameters.

The view supports all the use cases listed in the issue:

- **Add new subject** — opens the existing creation flow; whether the new subject becomes main or child is decided by whether the page already has a main subject
- **View subjects** — every subject on the page (main + children), regardless of whether a `#view` parser function references it
- **Delete subject** — per-row, with confirmation
- **Change main subject** — single dropdown at the top; switching demotes the previous main to a child, selecting "(none)" clears the main
- **Expand a row** to peek at its statement values (read-only), rendered via the same type-specific display components the infobox uses

Direct linking to a subject works via `?action=subjects#<subjectId>` — the matching row is scrolled into view, expanded and briefly highlighted.

Opening the row editor refetches the Subject and its Schema first, and the list is derived from the reactive store, so neither the dialog nor the row show stale data after an edit in another tab. The structural "remember to fetch before opening" contract is tracked separately in #757.

## Backend

Two new REST endpoints support the view:

| Verb | Path | Purpose |
|---|---|---|
| GET | `/neowiki/v0/page/{pageId}/subjects` | Full subjects on a page (main + children with statements) |
| PUT | `/neowiki/v0/page/{pageId}/mainSubject` | Set/clear the main subject by existing subject ID |

The existing `POST /page/{pageId}/mainSubject` (which *creates* a subject as main) is unchanged. PUT shares the path with distinct semantics.

`SetMainSubjectAction` orchestrates the swap on the in-memory `PageSubjects` aggregate before saving — promoting a child also demotes the previous main to a child rather than dropping it.

## Frontend

- New `SubjectsManagerPage.vue` and `SubjectStatementsView.vue`
- New `useSubjectPermissions` composable for best-effort hiding of Add/Edit/Delete actions and disabling the dropdown when the user lacks edit rights
- New typed `PageSubjects` view-model and `PageSubjectsDeserializer`
- `HttpClient` gains a `put` method
- `SubjectStore` gains `loadPageSubjects` and `setPageMainSubject`
- Reuses the existing `SubjectCreatorDialog` (already prop-driven for child vs main creation) and `SubjectEditorDialog` without changes

## Out of scope (deferred per issue discussion)

- Incoming-relation safety on delete (#544 follow-up)
- Ordering / drag-and-drop reordering
- Grouping by schema
- Retiring the existing tools-menu Create-subject entry — left untouched
- Structural fix for the editor-freshness contract (tracked as #757)

## Test plan

- [x] `make phpunit`, `make ts-test`, `make ts-lint`, `make cs`, `make ts-build` all green locally
- [x] On a content page (e.g. `Berlin`), confirm the NeoWiki tools menu has a "Manage subjects" link
- [x] Click the link → see header with Add button, Main Subject dropdown defaulted to the current main, and a row per subject (main first)
- [x] Click "Add subject" → schema picker → editor → save → page reloads, new row visible
- [x] Click Edit on a row → existing editor dialog opens → save → row reflects new label/values
- [x] In a second tab, edit the same subject and save; return to the manager tab, reopen the editor on that row — both the dialog and the row show the fresh values
- [x] Click Delete on a row → confirm → row disappears
- [x] Change Main Subject dropdown → previous main becomes a child row, selected one becomes main (border colour + Main badge move)
- [x] Set Main Subject to "(none)" → main is cleared, previous main appears as a child
- [x] Visit a page with no subjects yet → empty-state message + Add button still visible
- [x] Visit `?action=subjects#<subjectId>` → that row is scrolled into view, expanded, and visually highlighted
- [x] Log in as a user without edit rights → Edit / Delete / Add buttons hidden, dropdown disabled


alistair3149:
Reviewing the PR

JeroenDeDauw:
Thing local CC review indentified as "good to fix at some point" (not blocker), not evaluated by me:

  Code / design (PR-review flags, not done)                                                                                                                                                                       
                                                                                                                                                                                                                  
  1. __none__ sentinel magic for "clear main" appears 5× in SubjectsManagerPage.vue (lines ~225, 276, 288, 294, 303). Extract a const or map to null at the component edge.                                       
  2. useSubjectPermissions feeds fake IDs to the authorizer — as never cast around { text: '' } in composables/useSubjectPermissions.ts:25-26. Hinges on RightsBasedSubjectAuthorizer ignoring _subjectId (has    
  TODOs). Will bite when per-subject ACLs land.                                                                                                                                                                   
  3. hasMainSubject double-lookup in SubjectsManagerPage.vue:215. Simplify to != null, or add PageSubjects::hasMainSubject() to pair with isMain().                                                               
  4. onMainChanged dead branch — SubjectsManagerPage.vue:289 early-returns on newValue === null, but CdxSelect never emits null. Remove or trust it.                                                              
  5. SubjectsAction::getPageTitle() widened to public when parent is protected. No reason to expose.                                                                                                              
  6. SubjectStore.setPageMainSubject exposes comment but no caller passes one — SubjectsManagerPage.vue:295 never supplies it. Drop the param or add a summary dialog.                                            
  7. selectedMainId flickers to empty on load — initialize to '__none__' instead of null to avoid the cosmetic flash.                                                                                             
  8. RestSetMainSubjectPresenter silent-success default — initial state {status: 'unchanged'} + 200 means a refactor that forgets to call present* silently responds 200. Same shape in                           
  RestGetPageSubjectsPresenter::$apiResponse = [].  
  9. Silent JSON-decode in SetMainSubjectApi — malformed JSON → null → falls through to "Missing required field". Mis-attributed error. Pattern pre-exists in DeleteSubjectApi.                                   
  10. Statement array shape duplicated across GetPageSubjectsQuery::arrayifyStatements, RestGetPageSubjectsPresenter, and PageSubjectsDeserializer. Worth a shared DTO; the duplication grew in this PR.          
  11. PageSubjectsDeserializer defaults pageTitle to '' on miss (line ~46); downstream casts to SubjectWithContext. Empty string can render as a broken link. Prefer explicit nullable.                           
  12. Naming drift: all 24 new i18n keys use neowiki-managesubjects-* even though the action/class was renamed to subjects/SubjectsAction. AGENTS.md: "Use one term per concept."                                 

  Tests (review-tests flags, not done)              

  13. testIncludesSchemasWhenRequested only asserts schema names, not serialized content. Swap schemaSerializer->serialize() for an empty-string returner and nothing fails.                                      
  14. SubjectsActionTest doesn't cover onView() rendering at all — error-box fallback, addJsConfigVars, addModules, wgNeoWikiManageSubjectsPageId. Needs MediaWikiIntegrationTestCase.
  15. No REST coverage of the 'unchanged' status path — unit test covers the Action's no-op branch but not the end-to-end JSON shape.                                                                             
  16. No REST coverage of SetMainSubjectApi's 400 branches — "subjectId must be a string or null" and malformed JSON.                                                                                             
  17. REST integration tests hand-roll schema JSON where TestSchema::build() would do (GetPageSubjectsApiTest.php:32-38, 105-118).                                                                                
  18. GetPageSubjectsApiTest.php:137 uses an inline FQN — \ProfessionalWiki\NeoWiki\Tests\Data\TestRelation::build(...) should be a use.                                                                          
  19. createStub(CsrfValidator::class) in SetMainSubjectApiTest — consider a dedicated AcceptingCsrfValidator double for consistency with FailingSubjectAuthorizer etc.                                           
  20. newSpyPresenter() return type is object; intersection type for IDE/PHPStan support.                
  21. PageToolsBuilderTest decision-table gap: (isLatestRevision=false, canCreateMainSubject=true, devUiEnabled=false) not covered.                                                                               

  Frontend tests (not in this PR at all)            

  22. No Vitest for SubjectsManagerPage.vue, SubjectStatementsView.vue, useSubjectPermissions, PageSubjects, PageSubjectsDeserializer, or the new SubjectStore actions (loadPageSubjects, setPageMainSubject). Per
   AGENTS.md + project culture, not a merge blocker. Highest-value targets:                                                                                                 
    - onMainChanged revert-on-error (SubjectsManagerPage.vue:297-304)                                    
    - useSubjectPermissions error-path reset (composables/useSubjectPermissions.ts:32-38)                
    - applyHash (extract a pure function first)     
    - PageSubjectsDeserializer round-trip

alistair3149:
No blocking issue for the PR.
Follow-up PRs will be created per https://github.com/ProfessionalWiki/NeoWiki/issues/750#issuecomment-4298936011

Issue #756, opened 2026-04-22
Number input wrongly rejects any value when no maximum/minimum is set


JeroenDeDauw:
> Investigation driven by @JeroenDeDauw.
> Context: NeoWiki codebase on `master`, reproduced from a screenshot of the Subject editor showing a spurious "Maximum value is null." error on a Company schema's `founded at` field (number type) with no maximum configured.
> <sub>Written by Claude Code, `Opus 4.7`</sub>

## Symptom

When editing a Subject, entering a value into a number field whose Property Definition has no `maximum` (or a positive value when no `minimum` is set) produces a validation error:

> Maximum value is null.

This happens for any non-zero input and blocks form submission.

## Root cause

JSON round-trip converts "no limit" (`undefined` in the TS type) into an explicit `null`, and the validator's `!== undefined` guard lets `null` through.

1. PHP serializes missing bounds as explicit `null`. In `src/Domain/Schema/Property/NumberProperty.php`:

   ```php
   public function nonCoreToJson(): array {
       return [
           'precision' => $this->getPrecision(),
           'minimum' => $this->getMinimum(),
           'maximum' => $this->getMaximum(),
       ];
   }
   ```

2. The frontend copies that `null` straight onto the property. In `resources/ext.neowiki/src/domain/propertyTypes/Number.ts`:

   ```ts
   public createPropertyDefinitionFromJson( base: PropertyDefinition, json: any ): NumberProperty {
       return {
           ...base,
           precision: json.precision,
           minimum: json.minimum,
           maximum: json.maximum,
       } as NumberProperty;
   }
   ```

   The type says `maximum?: number` (implying `undefined` when absent), but runtime value is `null`.

3. The validator's guard does not reject `null`:

   ```ts
   if ( property.maximum !== undefined && value.number > property.maximum ) {
       errors.push( { code: 'max-value', args: [ property.maximum ] } );
   }
   ```

   With `property.maximum === null`, `null !== undefined` is `true`, and JS coerces `null` to `0` in the numeric comparison, so e.g. `2000 > 0` is true. The error fires with `args: [ null ]`, and the `neowiki-field-max-value` message (`"Maximum value is $1."`) renders as **"Maximum value is null."**.

The same guard pattern for `minimum` has the same bug (negative inputs would be rejected with "Minimum value is null.").

## Similar code elsewhere

[...]

PR #758, opened 2026-04-22
Ignore null bounds in NumberType validation

JeroenDeDauw:
Related to https://github.com/ProfessionalWiki/NeoWiki/issues/756

## What this changes

Hardens `NumberType` validation against `null` bounds at the JSON seam. The TS type declares `minimum?: number` / `maximum?: number` — "absent" should mean `undefined`, not `null`. Today `NumberType.createPropertyDefinitionFromJson` copies the JSON values through verbatim, so if a `null` ever arrives (e.g. from PHP's `NumberProperty::nonCoreToJson`, or a hand-edited schema page), it lands on the domain object as `null` and the validator's `!== undefined` guard lets it through. JS then coerces `null` to `0` in the numeric comparison, so any non-zero input is rejected with "Maximum value is null." (or "Minimum value is null." for negatives).

Normalize `null` to `undefined` at the JSON seam so the domain type stays honest and existing validator guards do the right thing.

Two regression tests cover the maximum and minimum paths independently (value `2000` exercises only `maximum > 0`; `-2000` exercises `minimum < 0`).

## Note on #756

I could not reproduce the user-visible "Maximum value is null." error on master today. The current `RestSchemaRepository.getSchema` fetches `/v1/page/Schema:<name>` (raw wikitext source), not the PHP-serialized JSON — and schema wikitext omits unset keys, so `minimum`/`maximum` arrive as `undefined`, not `null`, and the existing guard works. So this PR is defensive rather than a fix for a currently reachable bug via the SubjectEditor flow. Leaving #756 open — the original screenshot must have come from a different path or state that we haven't identified yet.

## Follow-up

The same null-leak pattern exists at the JSON seam for `Text.ts` (`minLength`/`maxLength`) and `DateTime.ts` (`minimum`/`maximum`). Neither triggers a user-visible bug today — `TextProperty.php` does not emit those keys, and `DateTime.ts` has a `parseStrictDateTime(null)` guard that masks the coercion — but both are latent and worth tidying as a separate change.

> Result of extensive back and forth with @JeroenDeDauw and this specific refinement of his after empirical verification in the dev environment.
> Context: the NeoWiki codebase on master, browser-based reproduction attempt against the running dev wiki, inspection of the PHP serialization and TS deserialization layers.

### DIFF
--- a/resources/ext.neowiki/src/domain/propertyTypes/Number.ts
+++ b/resources/ext.neowiki/src/domain/propertyTypes/Number.ts
@@ -28,9 +28,9 @@ export class NumberType extends BasePropertyType<NumberProperty, NumberValue> {
 	public createPropertyDefinitionFromJson( base: PropertyDefinition, json: any ): NumberProperty {
 		return {
 			...base,
-			precision: json.precision,
-			minimum: json.minimum,
-			maximum: json.maximum,
+			precision: json.precision ?? undefined,
+			minimum: json.minimum ?? undefined,
+			maximum: json.maximum ?? undefined,
 		} as NumberProperty;
 	}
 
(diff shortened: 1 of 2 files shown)


JeroenDeDauw:
Weird. I no longer get the error on master. CC says it can see the wrong code on master, but perhaps it's incorrect about that code being wrong, and the issue I ran into came from elsewhere?

PR #759, opened 2026-04-22
Rework Subject Management UI

alistair3149:
Follows-up to https://github.com/ProfessionalWiki/NeoWiki/pull/755

https://github.com/user-attachments/assets/de0e526e-eb8b-4402-ae71-fed98aa8dccc

## Summary

- Replaces the bespoke `<div>` + chevron button with a native `<details>`/`<summary>` disclosure for each Subject row. Keyboard and a11y are now native; Vue-owned `expandedIds` drives `:open` and `@click.prevent` on `<summary>` keeps programmatic opening (e.g. hash routing) working.
- Removes the global "Main subject" dropdown. The Main Subject gets its own tinted hero row with a filled-star indicator in progressive blue; child subjects live under an "Other Subjects on this Page" heading with a star button on each that promotes it to Main.
- Toolbar shows a plural-aware subject count on the left and keeps the primary `+ Add subject` button on the right. A full-width quiet `CdxButton` with a dashed border override sits at the end of the list as a secondary add affordance.
- Each row summary now reads `[chevron] [★ if main] [label / schema · N statements] [actions]`. On hover-capable desktop viewports the three action buttons hover-reveal; below the mobile breakpoint (< 640px) they collapse into a `CdxMenuButton` kebab (which is itself hover-revealed on hover-capable narrow viewports and stays visible while its menu is open).
- Expanded row footer holds a subtle monospace `ID: <id>` metadata block with copy-to-clipboard; the `<data value="">` element exposes the ID to scripts/selectors.
- Delete confirmation uses the destructive variant + trash icon via new optional `saveButtonAction`/`saveButtonIcon` props on `EditSummary.vue` (defaults preserve all other callers).
- `SubjectStatementsView` switches to semantic `<dl>`/`<dt>`/`<dd>`, adopts Infobox-style property/value layout with a continuous `border-top` divider, stacks label-over-value on narrow viewports, and flips to a 40/60 row layout at tablet+.

## Manual Browser Check

http://localhost:8484/index.php?title=Main_Page&action=neowiki-subjects

1. Toolbar shows "N subjects in total" on the left and `+ Add subject` on the right.
2. Main Subject renders as a tinted hero row with a blue filled star; its label is progressive blue and the subtle subtitle reads `[schema link] · N statements`.
3. "Other Subjects on this Page" heading appears above child rows; each has a promote-star alongside edit/delete.
4. Click or keyboard-activate any row summary — it toggles the details open/closed; Enter and Space on the summary both work.
5. Hover a row on a pointer device — the action cluster fades in. On a touch/no-hover viewport at desktop width, actions stay always visible.
6. Resize below 640px — the action cluster is replaced by a kebab button, which hover-reveals on hover-capable devices and stays visible while its menu is open.
7. Click the kebab → menu shows Set as main subject / Edit / Delete (delete in destructive style); picking one fires the correct action.
8. Star a child subject — it swaps to hero position; click Delete on any row and verify the confirm dialog's primary button is red + trash-iconed.
9. Expand a row: statements render 40/60 on desktop, stacked on narrow; footer shows "ID: s…" in monospace with the copy tooltip; clicking fires the "Copied…" toast.
10. Visit with a hash like `#s1234567890abcde` — that row opens, scrolls into view, and highlights with a progressive-subtle background.


JeroenDeDauw:
This looks a bunch better overall

Some concerns tho:
* "other subjects on this page"... as a user I would wonder "other in what sense"? If you don't know the first subject is the main subject or you're not yet familiar with that concept, then this is confusing / not self-explanatory.
* using the star icon to change something to a the main subject demotes the existing main subject. this is not obvious from the UI
* not possible any more to set there to be no main subject. If have a page where there should be none, and you click one of the stars by accident or to see what it does, you cannot really undo.

... let me wrap this up before opening other topics. Because that all hangs together. Your approach mirrors what my CC originally wanted to do, reverting the changes I deliberatly made on top: having a single list of subjects without per-subject controls for main subject and a dropdown for explicit main subject selection. That approach avoids all 3 above issues.

back to the concerns list:
* the statement list looks less nice somehow, and also seems to take up more space. The previous approach even had room to switch to a two column layout and make the info even more compact as per [comment](https://github.com/ProfessionalWiki/NeoWiki/issues/750#issuecomment-4298891364).
* subject id placement :thinking: you have to open to see/copy it. what about showing it on hover next to the action icons? in a way it is an action: "copy id". Or showing it next to the statement count.

JeroenDeDauw:
I'll make a follow-up going back to the dropdown approach

JeroenDeDauw:
Here it is:
* https://github.com/ProfessionalWiki/NeoWiki/pull/762

The other two concerns remain

alistair3149:
The section is deliberately added to separate main subject from child subjects. Compared to the previous approach, it is more obvious and easier to locate for user.

Perhaps we can enhance it further by having an animation that moves the subject up when a child subject is promoted to main subject.

I went against the main subject dropdown at the end because:

- It is more intuitive for a user to look at a specific child subject and say "make this the main one". The dropdown doesn't provide the context, and the user has to look at different places for information.
- It is a less used function, it doesn't justify having it on the top of the page, which are usually used for global actions.
- Dropdown in top toolbar usually implies some kind of filter or view switcher, not a tool to mutate data. 

> using the star icon to change something to a the main subject demotes the existing main subject. this is not obvious from the UI

That's why we also use a star icon in the main subject summary to create that connection. It's a common UI pattern.

> not possible any more to set there to be no main subject. If have a page where there should be none, and you click one of the stars by accident or to see what it does, you cannot really undo.

We can add the star button back to the main subject, I didn't consider that use case.

> The previous approach even had room to switch to a two column layout

We still have room for two column layout. However, I didn't implement here because it looks quite bad when you have statements with different number of lines on the same row. That's why I defer it from this PR. 

> subject id placement 🤔 you have to open to see/copy it. what about showing it on hover next to the action icons? in a way it is an action: "copy id". Or showing it next to the statement count.

I demoted it from the summary view because it is not useful for users who use the UI, since the ID isn't used at all in the UI. We can also move the copy ID into a menu button. There are limited width in the summary row and we have to decide on the order of importance.

JeroenDeDauw:
Both UIs, for easy comparison:

Screenshot_20260423_151439

Screenshot_20260423_151157-1

JeroenDeDauw:
Both HalloWelt and Bernhard prefer the second approach, the one with the separation and no dropdown.

Outstanding issues:
* Should be possible to not have a main subject, and to go from having one to not have one
* Star icons are easily confused with "watch" in MediaWiki
* Star icons are often used for "favourite", which suggests you can have more than one and are not replacing anything
* Make it clearer a replace happens when there is an exsiting main subject

PR #763, opened 2026-04-23
Restore Subjects tab in skin navigation

JeroenDeDauw:
Follows-up to https://github.com/ProfessionalWiki/NeoWiki/pull/755

This is somewhat obsoleting https://github.com/ProfessionalWiki/NeoWiki/pull/761, since we could just remove the sidebar link now we have the tab again. What do you think we should do with the sidebar link?

PR #755 dropped the SkinTemplateNavigation tab in favour of a
"Manage subjects" link in the NeoWiki tools-menu section. This
restores the tab (labelled "Subjects") while leaving the tools-menu
link in place, so the Subject management page is reachable from
both surfaces.

The tab is rendered only on content pages where
`SubjectsAction::isEligibleTitle()` returns true, and is marked
selected when the current request uses `?action=subjects`.

## Test plan

- [x] Visit a content page (e.g. `Main_Page`) and confirm a "Subjects" tab appears next to the standard action tabs
- [x] Click the Subjects tab → lands on `?action=subjects` and the tab is marked selected
- [x] Confirm the tab does not appear on non-content namespaces (e.g. Schema, Layout, Special pages)
- [x] Confirm the NeoWiki tools-menu still exposes the "Manage subjects" link

### DIFF
--- a/i18n/en.json
+++ b/i18n/en.json
@@ -222,6 +222,7 @@
 	"neowiki-layout-delete-success": "Deleted $1 layout",
 	"neowiki-layout-delete-error": "Failed to delete $1 layout.",
 
+	"neowiki-managesubjects-tab": "Subjects",
 	"neowiki-managesubjects-title": "Subjects on $1",
 	"neowiki-managesubjects-not-applicable": "Subject management is not available on this page.",
 	"neowiki-managesubjects-add-button": "Add subject",
--- a/extension.json
+++ b/extension.json
@@ -54,6 +54,7 @@
 		"ContentModelCanBeUsedOn": "ProfessionalWiki\\NeoWiki\\EntryPoints\\NeoWikiHooks::onContentModelCanBeUsedOn",
 
 		"SidebarBeforeOutput": "ProfessionalWiki\\NeoWiki\\EntryPoints\\NeoWikiHooks::onSidebarBeforeOutput",
+		"SkinTemplateNavigation::Universal": "ProfessionalWiki\\NeoWiki\\EntryPoints\\NeoWikiHooks::onSkinTemplateNavigationUniversal",
 		"ScribuntoExternalLibraries": "ProfessionalWiki\\NeoWiki\\EntryPoints\\NeoWikiHooks::onScribuntoExternalLibraries"
 	},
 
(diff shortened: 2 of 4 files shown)


JeroenDeDauw:
> This is somewhat obsoleting https://github.com/ProfessionalWiki/NeoWiki/pull/761, since we could just remove the sidebar link now we have the tab again. What do you think we should do with the sidebar link?

Actually, I reconsidered:

While having both at the same time is not desirable in most cases, wiki admins will have different preferences wrt how prominent the Subjects UI is.

Levels of prominence:
* Subjects UI is default rather than the "View tab", possibly with "View tab" hidden/disabled
* (only) Tab is shown (seems sensible default)
* (only) Sidebar link is shown
* neither is shown but action still exists
* neither is shown and action is not registered

Having a boolean setting for each could make sense, something like: showSubjectTab, showSubjectsToolLink

alistair3149:
A bit late to the topic, perhaps we can just make it into a wiki config on where the manage subject link is shown.

It might be desirable because skins handle the "view" menu differently:
- In Vector and Vector 2022, it is shown as tab on top right corner.
- In MinervaNeue, it is shown as a button in the center.
- In Timeless, it is shown as a button on the top right corner.
- In Chameleon, it varies.
- In Medik, it is rendered in the actions dropdown.

While the while the "tools" menu are handled consistently as a link menu in most if not all skins.

JeroenDeDauw:
Recorded prominense configurability via https://github.com/ProfessionalWiki/NeoWiki/issues/765

@alistair3149 did I record your input sufficiently with the "Note" on there?

i18n/en.json
```diff
@@ -222,6 +222,7 @@
 	"neowiki-layout-delete-success": "Deleted $1 layout",
 	"neowiki-layout-delete-error": "Failed to delete $1 layout.",
 
+	"neowiki-managesubjects-tab": "Subjects",
```
alistair3149, line 225: I am conflict about the label since we are adding the manage subject link to the action menu.
All labels in the action menu uses a verb because it is an action. However, if going with "Manage subjects" can take up too much space.

The alternative is to use the associated pages menu instead (e.g. `Page`, `Discussion` in Vector), that would probably be a more appropriate place to be (e.g. the homepage feature in user namespace from the GrowthExperiments extension). However, this is technically an action, not a separate page...

PR #770, opened 2026-04-23
DateTimeInput: format min/max error bounds as host-local wall-clock

malberts:
Follows-up to https://github.com/ProfessionalWiki/NeoWiki/pull/753

Validation error args were passed to the i18n message as the raw UTC ISO (e.g. `Maximum value is 2025-12-31T15:45:00.000Z.`), while the rest of the edit dialog shows host-local wall-clock. The user had to do the timezone math to reconcile the two.

The error now reads e.g. `Maximum value is 31 Dec 2025, 17:45:00 GMT+2.`, consistent with the subject display.

Extracts the `toLocaleString` formatting into a shared `formatDateTimeForDisplay` helper and uses it from both `DateTimeInput` (for error args) and `DateTimeDisplay` (for the rendered value), so the two surfaces cannot drift.

## Manual browser check

Open a subject page whose value is outside the schema's `minimum`/`maximum`, click edit, and verify the error message shows the bound in host-local format with a TZ abbreviation — matching the format used by the subject display infobox.


resources/ext.neowiki/src/domain/propertyTypes/dateTimeFormat.ts
```diff
@@ -0,0 +1,25 @@
+/**
+ * Formats a UTC ISO 8601 string as a human-readable host-local wall-clock
+ * with timezone abbreviation, using the user's browser locale.
+ *
+ * Falls back to the raw input when the ISO cannot be parsed, so malformed
+ * values surface verbatim in the UI rather than as `Invalid Date`.
+ */
+export function formatDateTimeForDisplay( iso: string ): string {
```
JeroenDeDauw: Why not in `DateTime.ts` in the same folder?

alistair3149:
PR LGTM 🎉

resources/ext.neowiki/src/domain/propertyTypes/dateTimeFormat.ts
malberts: I prompted this with https://github.com/ProfessionalWiki/NeoWiki/pull/753 being the main context where @alistair3149 created a separate utility file (dateTimeConversion.ts), so it followed that convention here. But checking the other property types, we actually keep all these random functions in the main file, so I think we should collapse both files back into the main file.

@alistair3149 was there a strong reason why you made it separate? Or think that it should stay separate?

I am aware that both `dateTimeConversion` and `dateTimeFormat` are slightly more generic helpers and not strictly only for consumption by the property type (e.g. you could arguably use these functions to convert and format a date like the NeoWiki git commit date to show the version somewhere, although this is YAGNI). But in that case, the code should probably not exist in `/propertyTypes/` and rather move to a more general helpers location.

alistair3149: I separated the logic because it is a generic helper function, and to make it more testable as a standalone "class".
Yes it should be moved to a more general helper location, do we have an existing one?

malberts: We don't. I'm moving the format function in this PR back into the main file, because that is the pattern we have in Url.ts, and then we can figure out a direction afterwards.

PR #771, opened 2026-04-23
Subject management UI follow-ups

alistair3149:
Follows-up to https://github.com/ProfessionalWiki/NeoWiki/pull/759

> Result of an extensive back and forth with @alistair3149.

https://github.com/user-attachments/assets/79bf4096-7831-4659-9b59-7450100fa71e

## Summary

- Swaps the star used to mark/promote the Main Subject for a pushpin (`cdxIconPushPin`). Stars were easily confused with MediaWiki's watch icon and with "favourite" affordances.
- Makes the pin indicator on the Main Subject row clickable to demote the subject, so the page can go from having a Main Subject to not having one. Read-only users keep a static pin as a status badge.
- Adds three empty-state branches:
  - Zero subjects: centered dashed panel with title, description, and a primary "Add subject" button; toolbar CTA hides.
  - No Main Subject but has children: same panel in the hero slot with a pin icon and instructions.
  - Has main but no children: heading + dashed add button are sufficient; no extra placeholder.
- Introduces `@row-height: 70px` so the end-of-list add-subject button matches the placeholder row height rather than relying on fragile text-height math.
- After promote/demote, the affected row gets a 2s Codex-style focus ring (`@box-shadow-outset-small` + `@box-shadow-color-progressive--focus`) and auto-scrolls into view when off-screen. The ring fades at `@transition-duration-medium`; an explicit transparent baseline shadow keeps the fade-out smooth.
- Respects `prefers-reduced-motion`: scroll behavior drops to `'auto'`, row transitions drop to `0s`.
- Copy pass on notifications and empty states: splits `main-subject-changed` into `-set`/`-cleared`, drops redundant "subject" wording, switches to contractions, quotes the label (`Deleted "$1".` etc.). Error toasts restructured from `title + error.message` into a single user-facing line; technical errors go to `console.error` for devs.
- Reloads the page after a successful subject deletion (mirroring the create flow) so the list reflects the deletion. Without the reload, the just-deleted subject stayed visible because of a pre-existing duplication between the store's by-ID map and `pageSubjects`. A proper reactive fix for both create and delete is tracked in #772.

## Manual Browser Check

http://localhost:8484/index.php?title=Main_Page&action=neowiki-subjects

1. Page with zero subjects -> centered dashed panel shows; toolbar `+ Add subject` is hidden. "Add subject" opens the creator.
2. Page with only a Main Subject -> hero + "Other Subjects on this Page" heading + dashed add button; heading stays visible as a structural cue.
3. Page with children but no Main Subject -> hero slot shows a dashed panel with a pin icon and "No main subject" heading; list and add-more still show.
4. Main Subject indicator is a pushpin (not a star). On the main row, clicking the left pin demotes the subject. On child rows the right-cluster pin promotes.
5. Promote a child -> success toast `Set "..." as the main subject.`; row jumps to the hero slot with a 2s blue focus ring, scrolling into view if off-screen.
6. Click the main row's left pin -> toast `Main subject cleared.`; the row lands in the children list with the same focus ring.
7. Delete a subject -> page reloads; after the reload the row is gone and a `Subject deleted.` toast appears. On failure the error toast shows without reloading.


JeroenDeDauw:
> Subjects capture structured data about this page

That's not right. Subjects contain structured data stored on the page. The main subject has structured data about the same concept as the page. (Structured data about the page is not in subjects as per earlier call, where we talked about accessing page properties stored on page nodes)

Perhaps something like

> A subject describes a concept with structured data

JeroenDeDauw:
image

This UI makes me want to drag the subject to the main subject area.

"or add a new one" :thinking: does adding a new one make it the main subject? or why is that part of the sentence there?

I'm unsure about the "pin a subject to make it main subject" wording. It seems inelegant to use the "pin" concept in the phrase. I don't have a better suggestion though. Also idk about the pin icon in that box. It looks good. But maybe its better more cohesive with that sentence, and to have the box take up less space. One concern I have is that it might look like there _should_ be a main subject, while it's fine to not have one.

alistair3149:
> This UI makes me want to drag the subject to the main subject area.

You are exactly right. That is the intention since drag to reorder is one of the follow-up items.
It is more complex to implement so I deferred it from this PR. That is also why the pin icon is used instead of other icons (I considered bookmark as well, but that also clash with reading list).

> "or add a new one" 🤔 does adding a new one make it the main subject? or why is that part of the sentence there?

The SubjectCreator by default adds a subject to the main subject if there are no main subject. We need to add the option to SubjectCreator to allow user picking whether they want to create a main subject or not. That's why the label is there now.

> One concern I have is that it might look like there should be a main subject, while it's fine to not have one.

I agree, I will look into how to address this.

alistair3149:
> Result of back and forth with @alistair3149.
> Context: the PR discussion thread and the NeoWiki Glossary for domain terms.

Pushed copy changes in response to the feedback.

**Empty-state description**: `"Subjects capture structured data about this page."` → `"A subject describes one thing with structured data."` Agreed with @JeroenDeDauw that the previous phrasing wrongly implied subjects describe the page itself; subjects describe one thing (per Glossary). "Concept" from Jeroen's suggested phrasing was swapped for "one thing" to stay concrete for an empty-state explainer.

**No-main placeholder** (addressing "might look like there should be a main subject, while it's fine to not have one"):
- Title: `"No main subject"` → `"No main subject set"` (states a fact instead of implying a deficiency).
- Description: `"Pin a subject below to make it the main subject, or add a new one."` → `"Main subjects are optional. Pin a subject below to set one."` The first sentence makes optionality explicit; dropped the "or add a new one" tail since it muddled the message and will be fully redundant once SubjectCreator gets an explicit main/child toggle.

Kept the hero-sized dashed panel as-is — it'll double as the drop target for the upcoming drag-to-promote behavior.

JeroenDeDauw:
> Main subject**s** are optional. Pin a subject below to set one.

This is potentially confusing as it can be read as a page supporting multiple main subjects.

That also reads like a really AI-style fix. It address my earlier concern with explicit text about, which seems inappropriate. 

I'll make a FU

PR #775, opened 2026-04-24
Simplify "no main subject" empty-state copy

JeroenDeDauw:
This is more concise and fits on a single line. It is not quite technically correct in the sense that the page is not about the subject, they are both about the same concept, but its my read that we can ignore that distinction for this help text which is aimed at people unfaimiar with the system and unlikley to be refined enough to differentiate those things anyway. 

Follows-up to https://github.com/ProfessionalWiki/NeoWiki/pull/773

Shortens the subtitle from 19 words to 15 and uses plainer English.
"Pin any subject below to make it the main subject, indicating it
represents the same topic as this page." becomes "Pin a subject to make
it the main subject — the one this page is about." — dropping "below"
(redundant when the subject list is the only place to pin from),
replacing the formal "indicating it represents the same topic as" with
an em-dash and appositive in natural English.

> Result of a back-and-forth with @JeroenDeDauw refining the wording after PR #773 shipped.
> Context: the NeoWiki codebase, the Glossary's Main Subject definition, and PR #773's final copy.

### DIFF
--- a/i18n/en.json
+++ b/i18n/en.json
@@ -229,7 +229,7 @@
 	"neowiki-managesubjects-empty-title": "No subjects yet",
 	"neowiki-managesubjects-empty-description": "A subject describes one thing with structured data.",
 	"neowiki-managesubjects-no-main-title": "No main subject set",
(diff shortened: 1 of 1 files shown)


JeroenDeDauw:
image

PR #777, opened 2026-04-24
[REFERENCE] Move DateTime frontend from NeoWiki core to RedHerb

malberts:
**Reference PR. Not intended to merge.**

Stacks on top of #754. Demonstrates what migrating an existing core property type into an extension looks like using the extension mechanism introduced in that PR.

Kept open while the team decides what permanent shape the "extension owns a real property type" story should take. The actual permanent forcing-function will be a **Color** property type in a separate PR (stacked on #754, not on this one). DateTime itself should stay in NeoWiki core long-term because:

- It's broadly useful across NeoWiki deployments.
- Disabling the owning extension would leave stored DateTime values unrenderable (silent UI failure demonstrated at https://github.com/ProfessionalWiki/NeoWiki/pull/754).
- Ongoing DateTime UX work (#770, #753) lands cleaner on master than on RedHerb.

## What this PR does

Single commit on top of #754:
- Deletes core DateTime code: `DateTime.ts`, `dateTimeConversion.ts`, the three `.vue` SFCs, their Vitest specs.
- Adds the equivalent as plain-JS MW-native SFCs in `tests/RedHerb/resources/`, with the same runtime behaviour:
  - `dateTimeConversion.js` — host-TZ-aware `toLocalInputValue` / `fromLocalInputValue` / `formatDateTimeForDisplay`.
  - `DateTimeDisplay.vue` — semantic `<time>` element, TZ-suffixed format.
  - `DateTimeInput.vue` — `cdx-text-input` with clock start-icon, live validation with host-local bound formatting (mirroring #770).
  - `DateTimeAttributesEditor.vue` — Codex text inputs, cross-validation (`min > max` rejected), wire-format guard.
  - `init.js` — subscribes to `mw.hook('neowiki.registration')` and registers `dateTime` with the full validate/createPropertyDefinitionFromJson/getExampleValue implementation.
  - `RedHerbFrontendHook.php` — `NeoWikiGetFrontendModules` handler adding `ext.redherb`.
- Removes DateTime exports from the public-API barrel (now that the files are gone).
- `extension.json` on RedHerb gains the ext.redherb RL module with the full DateTime packageFiles set; NeoWiki core drops the `cdxIconClock` icon and `neowiki-property-type-datetime` message (still owned by core's `neowiki-field-*` contract, but `property-type-datetime` belongs with the property-type's label owner).

## Status

- [x] `make tsci` green (688 tests)
- [x] `make phpunit filter=RedHerb` green
- [x] `make cs` phpcs + phpstan clean
- [x] Runtime-verified in the browser: DateTime property registers via the hook, bounds persist, validation renders

## Why reference-only

Moving DateTime out and then immediately back would be pointless churn. The preferred path is:

1. **Merge #754** (core mechanism only — barrel + hook + adapter + registrar + broad public-API). No property-type migration.
2. **Merge the Color PR** (permanent RedHerb example type). Stacked on #754 directly, not on this PR.
3. **Close this PR** once the above have merged, leaving the commit SHA reachable for anyone wanting to see what a real core→extension migration looks like.


malberts:
Moved fully to core in https://github.com/ProfessionalWiki/NeoWiki/pull/805

Issue #779, opened 2026-04-25
Schema content validator's type enum blocks extension-defined property types


malberts:
The frontend extensibility mechanism (`mw.hook('neowiki.registration')`, `FrontendRegistrar`, `PropertyTypeAdapter`) lets extensions register new property types end-to-end on the JS side, and the PHP `NeoWikiRegistrar->addPropertyType()` accepts arbitrary `PropertyType` implementations. But the persistence layer rejects any type the validator doesn't statically know about, so saves fail.

This affects normal users of the schema editor as well as direct API callers — it is not an API-only issue.

## Reproduction via the UI

With any extension that registers a new property type (e.g. RedHerb's `color`, see #780):

1. Go to `Special:Schemas`.
2. Click **Create** (or open an existing schema in **Edit schema**).
3. Add a property definition and pick the extension-defined type from the **Type** picker (e.g. "Color").
4. Click **Create schema** (or **Save**).

The dialog reports a save error and the schema is not created/updated. Devtools shows the underlying `PUT /rest.php/v1/page/Schema:Foo` returning `400 hookaborted`.

The same path is used both for new schemas (via `Special:Schemas` → Create) and for edits to existing schemas (via the Edit schema dialog). Both fail identically.

## Reproduction via the REST API

```http
PUT /rest.php/v1/page/Schema:Foo
{
  "source": "{\"description\":\"\",\"propertyDefinitions\":{\"Foo\":{\"type\":\"color\"}}}",
  "content_model": "NeoWikiSchema",
  ...
}
→ 400 hookaborted
```

Same shape with `"type":"text"` returns `201 Created`. Identical 400 for an unrecognized type like `"thisIsNotAType"` — confirming the gate is the type enum, not the content shape.

## Root cause

`src/Persistence/MediaWiki/schemaContentSchema.json` declares:

```json
"type": {
  "type": "string",
  "enum": [
    "text", "email", "url", "phoneNumber", "date", "time", "dateTime",
    "duration", "number", "currency", "progress", "checkbox", "select", "relation"
  ]
}
```

`SchemaContentValidator::validate()` runs against this on every save via the `EditFilter` hook (registered in `extension.json`). The hook fires for the REST `v1/page` PUT endpoint (used by the UI dialog) and for the legacy edit form. Any `type` not in the enum → validation error → hook sets `\$error` → MediaWiki aborts the edit → REST handler reports `hookaborted`.

## Fix options

[...]

PR #780, opened 2026-04-25
Add Color property type frontend in RedHerb

malberts:
First extension-provided property-type frontend built on the `neowiki.registration` hook surface. For #686. Stacks on #754.

This is an **alternative path** to PR #777, not a stack on it. PR #777 is a reference-only example that *moves* the existing core DateTime frontend into RedHerb to demonstrate the migration path; the intent for DateTime itself is to remain in NeoWiki core as a built-in. This PR introduces Color as a permanent example of an extension-defined property type — i.e. the shape any third-party extension would take — without disturbing core's built-in types.

## What it adds

* `ColorDisplay.vue` — renders a swatch + hex via a format-only check (intentionally tolerant of input-time constraints like `allowedColors` so previously-valid values keep rendering); falls back through `I18nSlot` for unparseable values.
* `ColorInput.vue` — `cdx-text-input` plus a live preview swatch and the type's start icon. Mirrors `useStringValueInput`'s convention: invalid mid-typing stays visible to the user but does not propagate through `update:modelValue` or `getCurrentValue()`.
* `ColorAttributesEditor.vue` — wraps an optional `allowedColors` palette in `NeoNestedField`, with a drag-reorderable list using `useSortable` and a dirty marker driven by `useChangeDetection`.
* `RedHerbFrontendHook.php` — implements `NeoWikiGetFrontendModules` so the new `ext.redherb` ResourceLoader module loads alongside `ext.neowiki`.
* `ColorProperty` gains an optional `allowedColors` attribute (validated as 6-digit hex strings), with round-trip coverage in `tests/phpunit/RedHerb/ColorPropertyTest.php`.

## Integration points exercised

* `mw.hook('neowiki.registration')` — registration via `PropertyTypeRegistration` as a second consumer alongside core property types.
* `NeoWikiGetFrontendModules` PHP hook.
* `NeoWikiServices.getPropertyTypeRegistry()` and `NeoWikiServices.getComponentRegistry()` (the latter for `getIcon('color')`).
* Composables: `useValueValidation`, `useChangeDetection`, `useSortable`.
* Vue components from the public-API barrel: `NeoNestedField`, `I18nSlot`.

## Known gaps

* **Persistence-side schema validation rejects extension-defined type names.** A schema with `"type": "color"` round-trips through `?action=raw` only after the JSON Schema in `src/Persistence/MediaWiki/schemaContentSchema.json` is loosened or built from the registry — see #779. This is the only thing blocking an end-to-end UI round-trip; everything else works.
* **Pinia store sharing not directly probed in this PR.** The integration assumes one Vue app + one Pinia, and the public-API barrel re-exports the same `useSubjectStore` symbol NeoWiki uses internally. A Vitest integration test that asserts cross-mount state visibility belongs in NeoWiki core's own test suite, not in RedHerb. Filed as a follow-up.
* **SchemaDeserializer / RestSchemaRepository / RestLayoutRepository** — no natural fit in the Color flow; left for a dedicated extension-integration-tests PR.

## Verification

* `make tsci` — 771 Vitest tests pass; lint clean.
* `make phpunit filter=ColorProperty` — 14 tests, 21 assertions, green.
* `make cs` — phpcs and phpstan clean.
* `curl 'http://localhost:8484/load.php?modules=ext.redherb&only=scripts&lang=en&debug=true'` — 200, ext.redherb status `ready`, all three Color SFCs in payload.
* Browser: in `Special:Schemas` → Create, the Color type appears in the property-type picker with a highlighter icon. Selecting it renders ColorAttributesEditor with a sortable palette (rows have drag handle with "Drag to reorder" tooltip, swatch, hex input, "Remove color" delete button; "+" adds entries; dirty marker `*` on the label). The Initial value `ColorInput` shows the type's highlighter start-icon. Save round-trip blocked at the persistence layer per #779.

### DIFF
--- a/tests/RedHerb/extension.json
+++ b/tests/RedHerb/extension.json
@@ -131,6 +131,7 @@
 			],
 			"packageFiles": [
 				"init.js",
+				"hexRegex.js",
 				"ColorDisplay.vue",
 				"ColorInput.vue",
 				"ColorAttributesEditor.vue",
--- a/tests/RedHerb/extension.json
+++ b/tests/RedHerb/extension.json
@@ -21,15 +21,26 @@
 		"ProfessionalWiki\\RedHerb\\": "src/"
 	},
 
+	"MessagesDirs": {
+		"RedHerb": [ "i18n" ]
+	},
+
+	"ExtensionMessagesFiles": {
+		"RedHerbAliases": "i18n/_Aliases.php"
+	},
+
 	"Hooks": {
 		"NeoWikiRegistration": "ProfessionalWiki\\RedHerb\\RedHerbHooks::onNeoWikiRegistration",
-		"NeoWikiGetFrontendModules": "RedHerbFrontendModulesHook",
+		"NeoWikiGetFrontendModules": [ "RedHerbFrontendModulesHook", "RedHerbFrontendHook" ],
 		"SidebarBeforeOutput": "RedHerbSidebarHook"
 	},
 
 	"HookHandlers": {
 		"RedHerbFrontendModulesHook": {
 			"class": "ProfessionalWiki\\RedHerb\\RedHerbFrontendModulesHook"
+		},
+		"RedHerbFrontendHook": {
+			"class": "ProfessionalWiki\\RedHerb\\RedHerbFrontendHook"
 		},
 		"RedHerbSidebarHook": {
 			"class": "ProfessionalWiki\\RedHerb\\RedHerbSidebarHook"
@@ -40,14 +51,6 @@
 		"RedHerbSubjectFinder": {
 			"class": "ProfessionalWiki\\RedHerb\\Specials\\SpecialRedHerbSubjectFinder"
 		}
-	},
-
-	"MessagesDirs": {
-		"RedHerb": [ "i18n" ]
-	},
-
-	"ExtensionMessagesFiles": {
-		"RedHerbAliases": "i18n/_Aliases.php"
 	},
 
 	"ResourceModules": {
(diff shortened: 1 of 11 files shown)


malberts:
Screenshot_20260425_224546

malberts:
I did not manually review the code here. This PR is meant for proving that we can reuse some core NeoWiki code in an extension. This might not necessarily be the best way of writing the extension logic, though.

alistair3149:
> Result of extensive back-and-forth with @alistair3149 about RedHerb as a reference example.
> Context: the PR diff, `useStringValueInput` in `resources/ext.neowiki/src/composables/`, and the `frontend-extensibility` base. Empirical check done in a local working tree.

Did you consider building `ColorInput` on top of `nw.useStringValueInput` instead of reimplementing the parts/validation/emit/initialization plumbing? Color is structurally a string-with-format property like URL, and the composable already handles the `userInputValues` separation, deep-watching property changes, and routing validation through the registered PropertyType's `validate()` callback.

Quick local check that it works from RedHerb's plain-JS context:

- Vite lib build (`vite.config.ts:15`) entries on `public-api.ts` with named CJS exports; `vue` and `pinia` are externalized so refs/Pinia identity is shared across modules.
- Plain-JS call from RedHerb's `<script>`:
  ```js
  var stringInput = nw.useStringValueInput(
      vue.toRef( props, 'modelValue' ),
      vue.toRef( props, 'property' ),
      ctx.emit,
      propertyType
  );
  ```
- Browser test in `Special:Schemas` → Create → Color: start icon shows, `#ff5733` renders the swatch, `#zzz` keeps the bad text visible while the error message routes through `validate()` → `neowiki-field-invalid-hex`.
- 773/773 vitest pass; lint clean. LOC: 194 → ~120 (just the swatch preview + LESS).

If there's a reason to keep the from-scratch implementation (e.g. deliberate pedagogical contrast), it's probably worth a code comment so future authors don't copy this as the canonical pattern. Otherwise consuming the composable looks like the more useful best-practice example for extension authors.

JeroenDeDauw:
> Did you consider building ColorInput on top of nw.useStringValueInput instead of reimplementing the parts/validation/emit/initialization plumbing? Color is structurally a string-with-format property like URL, and the composable already handles the userInputValues separation, deep-watching property changes, and routing validation through the registered PropertyType's validate() callback.

:+1: Same question

malberts:
I did not specifically ask it to do or not do that, only to come up with some expansion of the Color property that tested similar integration points as https://github.com/ProfessionalWiki/NeoWiki/pull/777.

I'll simplify as per the comments.

malberts:
> Result of back-and-forth with @malberts after the #786 merge of the schema-validator fix unblocked the end-to-end test path.
> Context: the NeoWiki codebase, Alistair's review feedback on the previous revision, and a browser walkthrough on the dev environment.

Rebased onto the updated `frontend-extensibility` and tightened a few things while there:

- **`ColorInput.vue` now consumes `nw.useStringValueInput`** instead of reimplementing the same plumbing inline (per Alistair's earlier comment). Color is structurally a single-string-with-format property, so it fits the canonical `UrlInput` pattern. Side effect: stored values that don't pass the registered `validate()` (non-hex, or outside an `allowedColors` palette) now render with the user-visible error rather than being silently blanked.
- **Dropped the dirty `*` marker from `ColorAttributesEditor`.** It duplicated the indicator `SchemaEditorDialog` already shows at the dialog level, and no core `*AttributesEditor` consumes `useChangeDetection` — only the six Dialog components do. The "extension consumes a NeoWiki composable from the public-API barrel" integration point stays covered by `useStringValueInput`, `useSortable`, and the `useSchemaStore`/`useSubjectStore` consumers in the merged sidebar probes.
- **Two small tidies:** renamed `HEX_PREVIEW_REGEX` to `HEX_REGEX` so all four consumers of `hexRegex.js` use the same name, and extracted a private `isValidHex()` helper in `ColorProperty.php` that's called from both the `allowedColors` and default-value checks in the constructor.

Verified end-to-end via Schema → Subject round-trip in the browser: invalid hex surfaces `neowiki-field-invalid-hex`, empty-and-required surfaces `neowiki-field-required`, valid input emits a `StringValue` and round-trips back through `ColorDisplay` as a swatch.

PR #794, opened 2026-04-28
Move page-subject predicates off ViewHtmlBuilder

### LINKED ISSUE #789: Establish public PHP API surface for extensions consuming NeoWiki services
> Result of code review of #781 with @alistair3149, reframed once we treated RedHerb as a true third-party extension.
> Context: the NeoWiki codebase and the `frontend-extensibility` branch.

Surfaced by #781. RedHerb — which exists to model a separate MediaWiki extension consuming NeoWiki's extension points — has to reach into NeoWiki internals from PHP because there is no documented public PHP API. The TS side has `resources/ext.neowiki/src/public-api.ts` (broad, explicitly alpha-stable barrel). The PHP side has no equivalent.

## Concrete breach points in #781

1. **`tests/RedHerb/src/RedHerbSidebarHook.php:18-19`** — calls `NeoWikiExtension::getInstance()->newViewHtmlBuilder()->pageHasMainSubject( $title )` to gate a sidebar link. Traverses the singleton, then a factory named after view-HTML construction, to reach a domain predicate.

2. **`tests/RedHerb/src/Specials/SpecialRedHerbSubjectFinder.php:25`** — calls `NeoWikiHooks::addNeoWikiModules( $out, $this->getSkin() )` to bootstrap NeoWiki's frontend on a non-NeoWiki special page. `NeoWikiHooks` lives in `EntryPoints/`, which is the namespace for MW hook entry points, not a public API.

Every extension needing either capability would copy these patterns and bind to internals.

## Design options to react to

**(A) Service-container facade.** Register a `NeoWiki.PublicApi` MW service exposing a small, documented method set. Extensions resolve via `MediaWikiServices::getInstance()->get( 'NeoWiki.PublicApi' )`. Mockable, version-able, plays with existing MW patterns.

**(B) Static facade class.** New top-level `ProfessionalWiki\NeoWiki\NeoWikiPublicApi` (avoiding the name `NeoWikiServices`, already used on the TS side). Easiest to call from anywhere, harder to test cleanly, encourages global state.

**(C) Promote `NeoWikiExtension` selectively.** Mark specific methods `@public-api`, leave the rest internal. Lowest churn, but `::getInstance()` traversal stays painful and the "internal vs. public" contract lives in docblocks rather than structure.

## Open questions

- Stability contract: do we want a PHP analogue of the TS barrel's "0.x alpha — anything can change" note, or a stricter promise from day one?
- Should the PHP surface narrow before production, paralleling the TS narrowing follow-up referenced at `public-api.ts:5`?
- Initial scope: just `pageHasMainSubject` + frontend bootstrap, or sweep wider (subject CRUD, schema lookup, etc.) while we're designing the shape?

Related: #686 is the TS-side property-type extension mechanism; this is the PHP counterpart.
Related: #782

JeroenDeDauw:
For https://github.com/ProfessionalWiki/NeoWiki/issues/789

`pageHasMainSubject` and `pageHasSubjects` are pure subject lookups that don't render any HTML. Reaching them via `newViewHtmlBuilder()` forced consumers (internal hook code; #781 RedHerb sidebar) to traverse a factory named after view-HTML construction to ask a domain question.

Move both predicates to `SubjectContentRepository`, where the underlying `getSubjectContentByPageTitle` lookup already lives. `ViewHtmlBuilder` now only renders HTML.

Addresses one of the two breach points raised in #789. The other (`NeoWikiHooks::addNeoWikiModules` called from a foreign extension's SpecialPage) lives on the `frontend-extensibility` branch and will follow up after #754 lands.

Doesn't claim to close #789 — that issue's broader question of whether NeoWiki should commit to a documented PHP API surface is unresolved on the issue thread.

### DIFF
--- a/src/EntryPoints/NeoWikiHooks.php
+++ b/src/EntryPoints/NeoWikiHooks.php
@@ -75,7 +75,7 @@ class NeoWikiHooks {
 		if ( self::shouldShowSubjectCreator( $out ) ) {
 			$attrs['data-mw-neowiki-create-subject'] = 'true';
 			$attrs['data-mw-neowiki-page-has-main-subject'] =
-				NeoWikiExtension::getInstance()->newViewHtmlBuilder()->pageHasMainSubject( $out->getTitle() )
+				NeoWikiExtension::getInstance()->newSubjectContentRepository()->pageHasMainSubject( $out->getTitle() )
 					? 'true'
 					: 'false';
 		}
--- a/src/Persistence/MediaWiki/Subject/SubjectContentRepository.php
+++ b/src/Persistence/MediaWiki/Subject/SubjectContentRepository.php
@@ -35,6 +35,16 @@ class SubjectContentRepository {
 		return $this->getSubjectContentFromWikiPage( $this->wikiPageFactory->newFromTitle( $pageIdentity ) );
 	}
 
+	public function pageHasSubjects( PageIdentity $pageIdentity ): bool {
+		return $this->getSubjectContentByPageTitle( $pageIdentity )?->hasSubjects() === true;
+	}
+
+	public function pageHasMainSubject( PageIdentity $pageIdentity ): bool {
+		return $this->getSubjectContentByPageTitle( $pageIdentity )
+			?->getPageSubjects()
+			->getMainSubject() !== null;
+	}
+
 	public function getSubjectContentByRevisionId( int $revisionId ): ?SubjectContent {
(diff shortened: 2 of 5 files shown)


JeroenDeDauw:
Superseded — placement was wrong (predicates lying about being SubjectContent-shaped). Will reopen with a dedicated Application/PageSubjectsLookup using domain types.

PR #801, opened 2026-04-30
Ensure PHP 8.1 compatibility

HamishSlater:
Remove readonly modifier
Remove class constants with native types

ERM20742

### DIFF
--- a/.github/workflows/ci-php.yml
+++ b/.github/workflows/ci-php.yml
@@ -24,7 +24,7 @@ jobs:
       matrix:
         include:
           - mw: 'REL1_43'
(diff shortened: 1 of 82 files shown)


malberts:
If we do this we should also lower the version in composer.json and CI.

I'm aware that MW 1.43 was released with PHP 8.1 support, but that PHP version is EOL by now. In which environment are you running where a newer version is unavailable?

JeroenDeDauw:
@HamishSlater can we close this?

HamishSlater:
Obsolete; will be released with REL1_47

PR #804, opened 2026-04-30
Centralize neowiki.registration fire on ext.neowiki module load

malberts:
> Result of extensive back and forth with @malberts.
> Context: the NeoWiki codebase, PR #803 (the just-merged targeted fix), and the project's `mw.hook` replay-on-late-subscribe semantics covered by `HookRegistration.spec.ts`.

Follows-up to https://github.com/ProfessionalWiki/NeoWiki/pull/803

Related to https://github.com/ProfessionalWiki/NeoWiki/issues/686

The previous setup duplicated `mw.hook( 'neowiki.registration' ).fire()` in each of the six mount-point initializers in `neowiki.ts`, plus a seventh in RedHerb's SubjectFinder init added by PR #803. Each fire constructed a fresh `FrontendRegistrar` over the same singleton registries — pure boilerplate that any new custom mount point would have to remember to replicate or risk extension property types silently failing to register.

Extract a single `fireRegistrationHook()` and call it once at `ext.neowiki` module load, before the initializers. Subscribers that load before or after benefit equally thanks to `mw.hook`'s documented replay-on-late-subscribe behavior, already covered by `HookRegistration.spec.ts`. The seven redundant per-mount fires are removed.

A new boot-time test asserts that loading `@/neowiki` fires the hook with a registrar wired to the live `NeoWikiExtension` registries — registering a fake type via the fired registrar must be observable on the singleton's `PropertyTypeRegistry` *and* `TypeSpecificComponentRegistry`. This catches a missing fire, a fully detached registrar, and a half-detached registrar (where one registry is fresh and the other live — would only surface as broken Vue rendering in the browser).

### DIFF
--- a/resources/ext.neowiki/src/neowiki.ts
+++ b/resources/ext.neowiki/src/neowiki.ts
@@ -34,6 +34,13 @@ export function registerSubjectCreatorClickHandler( pinia: Pinia, signal?: Abort
 	}, { signal } );
 }
 
+function registerExtensions(): void {
+	const ext = NeoWikiExtension.getInstance();
+	mw.hook( 'neowiki.registration' ).fire(
+		new FrontendRegistrar( ext.getTypeSpecificComponentRegistry(), ext.getPropertyTypeRegistry() ),
+	);
+}
+
 function initializeNeoWikiApp(): void {
 	queueMicrotask( () => {
 		const neowikiApp = document.querySelector( '#mw-content-text > #ext-neowiki-app' );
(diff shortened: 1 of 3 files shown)


resources/ext.neowiki/src/neowiki.ts
```diff
@@ -196,6 +182,7 @@ const isTestEnvironment = typeof window !== 'undefined' &&
 	( window as unknown as { neoWikiTestMode?: boolean } ).neoWikiTestMode === true;
 
 if ( !isTestEnvironment ) {
+	registerExtensions();
```
malberts: I considered matching the PHP `ensureExtensionsRegistered()`, but the TS does not have a guard, and this is the point where extensions get registered.

resources/ext.neowiki/src/neowiki.ts
```diff
@@ -34,6 +34,13 @@ export function registerSubjectCreatorClickHandler( pinia: Pinia, signal?: Abort
 	}, { signal } );
 }
 
+function registerExtensions(): void {
+	const ext = NeoWikiExtension.getInstance();
+	mw.hook( 'neowiki.registration' ).fire(
```
JeroenDeDauw, line 39: regsiterExtensions? that seems not quite right. It fires a hook which extensions often register handlers for. So `triggerRegistrationHooks`, `fireRegistrationEvent`, or similar, seems better

malberts: I went back to the original name that literally matches what happens: `fireRegistrationHook`.

PR #805, opened 2026-05-01
Promote DateTime property type to NeoWiki core

malberts:
> Result of back-and-forth with @malberts.
> Context: the NeoWiki codebase, the existing built-in property types
> (Text/Url/Number/Select/Relation), and the in-tree RedHerb test extension
> where DateTime previously lived.

Move the `dateTime` PropertyType backend from the RedHerb test extension into
NeoWiki core, alongside the other built-in types. The frontend already lived
in NeoWiki, so this completes the migration.

- `DateTimeType` and `DateTimeProperty` move to `src/Domain/PropertyType/Types/`
  and `src/Domain/Schema/Property/`.
- Registered via `PropertyTypeRegistry::withCoreTypes()` and
  `Neo4jValueBuilderRegistry::withCoreBuilders()`, using the shared `$toScalars`
  builder — matching what RedHerb did.
- RedHerb keeps Color as the canonical extension-point example.
- The two RedHerb-specific DateTime assertions in `NeoWikiRegistrationHookTest`
  are dropped; equivalent core coverage is added via the moved
  `DateTimePropertyTest`, a new `DateTimeTypeTest`, and an extra
  `hasBuilder( 'dateTime' )` assertion in `Neo4jValueBuilderRegistryTest`.

No behaviour changes.

### DIFF
--- a/src/Domain/PropertyType/PropertyTypeRegistry.php
+++ b/src/Domain/PropertyType/PropertyTypeRegistry.php
@@ -5,6 +5,7 @@ declare( strict_types = 1 );
 namespace ProfessionalWiki\NeoWiki\Domain\PropertyType;
 
 use OutOfBoundsException;
+use ProfessionalWiki\NeoWiki\Domain\PropertyType\Types\DateTimeType;
 use ProfessionalWiki\NeoWiki\Domain\PropertyType\Types\NumberType;
 use ProfessionalWiki\NeoWiki\Domain\PropertyType\Types\RelationType;
 use ProfessionalWiki\NeoWiki\Domain\PropertyType\Types\SelectType;
@@ -25,6 +26,7 @@ class PropertyTypeRegistry implements PropertyTypeLookup {
 		$registry->registerType( new NumberType() );
 		$registry->registerType( new SelectType() );
 		$registry->registerType( new RelationType() );
+		$registry->registerType( new DateTimeType() );
 		return $registry;
 	}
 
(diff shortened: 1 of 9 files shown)


alistair3149:
> Result of code review with @alistair3149.
> Context: the NeoWiki codebase, specifically the sibling property type registrations and test patterns under `src/Domain/PropertyType/Types/` and `tests/phpunit/Domain/...`.

Reviewed the diff against the surrounding code. No blockers — the move is mechanically correct, behaviour-preserving (`$toScalars` matches the prior `fn ( $value ) => $value->toScalars()`), namespaces line up with sibling types, the frontend `typeName = 'dateTime'` matches `DateTimeType::NAME`, and no DateTime references are left behind in `tests/RedHerb/`.

A couple of minor test-shape items worth tracking, both introduced by this PR:

1. **Duplicate test** — `DateTimeTypeTest::testBuildPropertyDefinitionFromJsonReturnsDateTimeProperty` is functionally identical to `DateTimePropertyTest::testBuildPropertyDefinitionFromJsonViaType` (same setup, same `'minimum'` value, same assertions). Sibling `*TypeTest` classes (`NumberTypeTest`, `UrlTypeTest`, `RelationTypeTest`, `TextTypeTest`) are deliberately tiny and only cover `getDisplayAttributeNames()`. Suggest dropping the duplicate from `DateTimeTypeTest` to match.
2. **`testValueTypeIsString`** in `DateTimeTypeTest` is uneven with siblings — none of the other `*TypeTest`s assert `getValueType()`. Either drop it for consistency, or extend the same assertion to siblings in a follow-up.

Optional follow-up (not specific to this PR): `DateTimePropertyTest` could adopt the `assertSerializationDoesNotChange` round-trip pattern used by `NumberPropertyTest` / `TextPropertyTest` via `PropertyTestCase`, to exercise the registry-driven `PropertyDefinition::fromJson()` path.

PR #811, opened 2026-05-05
Self-bootstrapping bundled dev env at the extension root

alistair3149:
Converts `Docker/` from a try-it-out-only stack into the canonical NeoWiki
dev environment, with a single `make` entry point at the extension root.
A fresh clone now goes from zero to a running wiki with `make dev`.

## Summary

- New `dev-mw` Dockerfile stage and `docker-compose.dev.yml` overlay with a
  bind-mounted NeoWiki source, `MW_MODE=dev`-aware `SettingsTemplate.php`,
  and an opt-in `LocalSettings.local.php` hook.
- `make dev`, `make phpunit`, `make cs`, `make tsci`, `make bash`,
  `make reset`, etc. live at the extension root in a single dual-mode
  Makefile (proxies via `docker compose exec` from the host; runs binaries
  directly inside the container).
- `make bootstrap` clones MW core into the gitignored `Docker/mediawiki/`,
  so a public clone is self-bootstrapping.
- Auto-allocated host ports from documented ranges: `MW_SERVER_PORT`
  8484-8499, `MAILCATCHER_PORT` 8025-8040. `make dev-tools` opt-in adds
  Neo4j Browser/Bolt at 7474/7687.
- `Docker/.env` now ships as `Docker/.env.dist` and is auto-copied on
  first run; runtime port allocation no longer dirties the working tree.
- `opcache.validate_timestamps=1` re-enabled in the dev image so PHP
  edits (including `LocalSettings.local.php` overrides) land on the next
  request without a container restart.

## Test plan

- [x] Fresh clone, no prerequisites: `make dev` succeeds end-to-end and
      `http://localhost:8484/index.php/Main_Page` renders.
- [x] `make phpunit filter=SchemaNameTest`, `make phpcs`, `make ts-lint`
      pass.
- [x] `make dev-tools` exposes Neo4j Browser at `http://localhost:7474`.
- [x] Override `port=8488` correctly allocates `MW_SERVER_PORT=8488` while
      keeping `MAILCATCHER_PORT` in its 8025-8040 range.
- [x] Edits to `Docker/LocalSettings.local.php` propagate to the running
      container within ~2s, no restart needed.
- [x] `make stop` / `make down` / `make reset` work.


JeroenDeDauw:
Interesting approach

Seems like a mediawiki clone ends up being on the host fs, which is great for IDE autocomplete and AI grep. However composer dependencies added to `vendor/` with `composer install` in the Docker image build won't be there. This adds some friction. Not a deal breaker for MVP.

JeroenDeDauw:
PR LGTM, though I have not tried it yet.

---

> Result of back and forth with @JeroenDeDauw.
> Context: this PR diff plus an inspection of the MW `REL1_43` submodule layout and the dev-entrypoint flow.

**Goal: a fully-populated MW `vendor/` on the host filesystem.**

Today the relationship between host and container is one-way for MW core: `Docker/mediawiki/` is only a build context. The Dockerfile's `composer install` writes into the image, and the dev-entrypoint's `composer update` (which merges NeoWiki deps via composer-merge-plugin) writes into the container's writable layer. Neither flows back to host. The host's `mediawiki/vendor/` stays at whatever the WMF deployment submodule provides — production deps only, no dev deps, no post-merge state.

This matters for IDE autocomplete and for AI tooling that reads what's on disk. Both work much better when host vendor matches what the container actually runs.

**Possible fix, kept consistent with the self-bootstrapping design:**

1. Add a host-side composer install to \`make bootstrap\`, run via \`docker run composer:2\` so no host PHP is required. ~30s one-shot, idempotent. Gives host the full vendor including dev deps.
2. Optional follow-up for drift: a \`make sync-mw-vendor\` target that re-runs (or \`docker cp\`s) when an MW-root dep changes. Or bind-mount \`Docker/mediawiki/vendor/\` into the container so subsequent composer ops write through to host — bidirectional for vendor only, MW core PHP stays image-baked.

Both are small and orthogonal to the rest of the Makefile. Happy to do as a follow-up rather than block this PR.

alistair3149:
> Back and forth with @alistair3149 → simplified to bind-mount only.
> <sub>Claude Code, `Opus 4.7`</sub>

Done in 32decc07: bind-mount `Docker/mediawiki/vendor/` to host so the entrypoint's existing `composer update` writes through. Dropped the dev-mw `composer install` from the Dockerfile (bind mount shadows it). chown/chmod for host-deletability, matching the NeoWiki-vendor pattern.

Verified after a clean first boot: 52 packages on host vendor, including `laudis/neo4j-php-client` and `jeroen/file-fetcher`. `composer.lock` stays in-image (MW REL1_43 doesn't track one).

JeroenDeDauw:
Works on my machine (SELinux + rootless podman)

JeroenDeDauw:
Well, apparently, it works because CC decided to apply a bunch of hacky one-time fixes. Let me follow-up with proper fixes

JeroenDeDauw:
> Result of @JeroenDeDauw locally testing the PR on rootless podman + SELinux (Kinoite), then directing both fixes after I flagged the issues. Context: this PR's diff and commit history, the NeoWiki codebase, local end-to-end testing of `make dev` from a fresh clone in a default-labeled directory, and Docker's podman/SELinux semantics for `:z`.

Pushed [2a14b2ee](https://github.com/ProfessionalWiki/NeoWiki/commit/2a14b2ee) with two fixes that surfaced when running `make dev` end-to-end on rootless podman + SELinux enforcing.

### 1. Healthcheck always reports `unhealthy`

The previous test was `curl -fsSL http://localhost:80/index.php/Special:Version | grep 'Jeroen De Dauw'`. The `-L` follows redirects: `/index.php/Special:Version` issues a 301 toward the canonical short URL, and the chain ultimately resolves through `$wgServer` (`http://localhost:8484` by default) — the **host-side** URL, which isn't reachable from inside the container. Curl exits non-zero before grep ever runs.

New check: `curl -fs http://localhost:80/wiki/Special:Version | grep -q 'Jeroen De Dauw'`. The canonical `/wiki/` path doesn't redirect, so no `-L` is needed. Kept the `Jeroen De Dauw` grep — substituting a more generic string like `MediaWiki` would risk false positives on error pages, and the personal name is a stronger signal that Special:Version actually rendered. Verified the mediawiki container transitions to `(healthy)` after one healthcheck interval.

### 2. Bind mounts need `:z` for SELinux-enforcing rootless podman

On Fedora / RHEL / Kinoite / Silverblue with rootless podman, an unlabeled bind mount inherits the host directory's SELinux type (typically `user_home_t`), which container processes can't read. Without the label the dev stack errors out with permission-denied on the bind-mounted NeoWiki source — the workaround is a manual `chcon -R -t container_file_t .` on the worktree.

Adding `:z` (added `,z` for the read-only one) tells the container engine to relabel the source dir to `container_file_t` at mount time. **No-op on Docker Engine without SELinux**, so plain Docker users see no change.

Verified by running `make dev` from a fresh clone in a `user_home_t`-labeled directory: podman auto-relabels the worktree to `container_file_t`, the stack comes up, the wiki responds, and no `chcon` was needed.

alistair3149:
@malberts are you able to check if you can boot up the dev environment?
We all use different local setups so I wanted to make sure that this work across everyone's setup.

malberts:
`make dev` works for me, but I still need to review.

malberts:
`Docker/scripts/set-port.sh` skips the host-port-free check whenever `MW_SERVER_PORT` is non-empty. The Makefile does `include Docker/.env` + `export` and then passes `"$${MW_SERVER_PORT:-}"` to `set-port.sh`. Since `.env.dist` ships with `MW_SERVER_PORT=8484`, every fresh clone hits the "explicit request" branch on first run, so `is_port_free` is never called for the MW port.

Reproduction (fresh clone, host already has 8484 bound):

```
$ make dev
...
Error response from daemon: failed to set up container networking: ...
Bind for 127.0.0.1:8484 failed: port is already allocated
```

This contradicts the `Docker/README.md` claim that "auto-allocation skips ports that are already bound on the host" and breaks the multi-worktree side-by-side flow that is this PR's central pitch.

> Result of hands-on testing of this PR by @malberts, then this analysis.
> Context: this PR's diff and a live `make dev` run from a fresh `/tmp/neowiki-pr811` clone.

malberts:
Summary of issues found while reviewing this PR, in addition to the port-allocator one I [noted above](https://github.com/ProfessionalWiki/NeoWiki/pull/811#issuecomment-4429783096).

The main thing I'll follow up on is the production VPS upgrade path:

- **Production VPS upgrade path is gone.** `Docker/Makefile` is removed, and the new top-level Makefile's maintenance targets (`install-db`, `load-neo4j-users`, `import-demo-data`, `update-dot-php`) all route through `$(DC_DEV)` (`-f docker-compose.dev.yml --profile dev`). The dev overlay rebuilds the image locally and bind-mounts the parent NeoWiki repo plus `Docker/mediawiki/vendor` — neither fits a `Docker/`-only VPS running off the prebuilt ghcr image. `make up` happens to use the GH image but is positioned as a try-it-out target, omits `--profile server`, and pairs with the same dev-overlay maintenance commands. Raw `docker compose --profile server` from `Docker/` still works as a stopgap. I'll handle this in a follow-up PR.

Other findings:

- **`_wait-mw` may attach to the wrong wiki.** Polls `http://localhost:$$MW_SERVER_PORT/` from the host, while `_first-run-seed` runs against this project's containers via `docker compose exec`. If anything else already binds the port, the wait gate reports "Reachable" against an unrelated wiki and the seed proceeds. The check also only verifies HTTP reachability, not MediaWiki readiness — any response counts as "ready". The port-allocator bug above makes both materially more likely.

- **`Docker/LocalSettings.local.php` directory-bind footgun.** `bootstrap` `touch`es the host file so `make dev` is safe, but anyone bringing the dev stack up via `docker compose` directly (bypassing `make`/`bootstrap`) will see Docker create a directory in its place and silently break PHP loading. Observed the directory form left behind from a prior session.

- **`MW_BRANCH` / `MW_GIT_URL` undocumented.** Referenced in `bootstrap` but absent from `.env.dist` and the READMEs.

- **`make reset` skips `composer-install`** (which `_first-run-seed` includes). Asymmetry that only bites when `composer.json` changes.

> Result of hands-on testing of this PR by @malberts, then this analysis.
> Context: this PR's diff and live `make dev` runs from `/tmp/neowiki-pr811`.

malberts:
I'm merging so we can unblock. The main issue for me is how to run/update the demo wiki now. Ideally, we do not run it in dev mode, so it at least performs like a production wiki.

malberts:
This breaks the image building CI. I am investigating alongside the prod env issue.

Makefile
```diff
@@ -1,70 +1,343 @@
...
+# Worktree:      mediawiki/extensions/NeoWiki-feature-x/  -> neowiki-neowiki-feature-x
+PROJECT_NAME := $(shell echo "neowiki-$(notdir $(CURDIR))" | tr A-Z a-z)
+
+PORT_RANGE_START := 8484
+PORT_RANGE_END := 8499
 
-IS_PODMAN := $(shell docker info 2>&1 | grep -qi podman && echo 1 || echo 0)
+# ---- Compose invocations -----------------------------------------------------
 
+DC := docker compose -p $(PROJECT_NAME) --project-directory Docker
```
malberts, line 25: `--project-directory` is not a valid argument when using Podman.

@JeroenDeDauw did CC leave an uncommitted local modification on your system? I am running into this on a Podman system now.

@alistair3149 was there a specific issue you ran into that required this?

alistair3149: It can be dropped, it was just safety against older Docker compose that uses current working directory for `.env`.

PR #812, opened 2026-05-06
Store dateTime statements as Neo4j datetime values

malberts:
For https://github.com/ProfessionalWiki/NeoWiki/issues/831

Statement values for the dateTime property type were written to Neo4j as
plain ISO 8601 strings, leaving them un-queryable with Cypher's datetime
functions (comparisons, ordering, range filtering).

Value builders registered with `Neo4jValueBuilderRegistry` can now return
typed objects: the Neo4j driver converts `DateTimeInterface` instances
nested in the props map to native Bolt temporal values, so the single-map
`SET n = $props` write stays intact and no Cypher constructor expressions
or query-string assembly are needed. The dateTime builder parses statement
strings with `DateTimeProperty::parseStrictDateTime()`; malformed values
are omitted from the graph projection (the revision slot stays
authoritative). Extension-defined property types get the same capability
by returning typed objects (e.g. Laudis `Date`, `Duration`, points) from
their builders.

The result normalizer converts temporal values in query results to
ISO 8601 strings, and other Cypher property objects to arrays, instead of
throwing `Unsupported Cypher value type` (previously reachable via
`RETURN datetime()`).

Existing graph data is re-typed by the standard rebuild
(`maintenance/RebuildGraphDatabases.php`); until then, pre-existing
string-stored values silently don't match temporal comparisons.

The page-side typed path (`PageValue` / `Neo4jQueryStore::extractTypedValues()`)
is deliberately left untouched.

## Follow-ups (out of scope, filed separately)

- The `date` property type still stores plain strings (same un-queryability this
  fixes for `dateTime`): https://github.com/ProfessionalWiki/NeoWiki/issues/989
- The page-facts datetime write path emits zone-less strings that shift under a
  non-UTC Neo4j timezone: https://github.com/ProfessionalWiki/NeoWiki/issues/990

> First iteration (Cypher constructor wrapping) written by Claude Code
> `Opus 4.7 (1M context)` with @malberts. Reworked to driver-native typed
> parameters with @JeroenDeDauw, following an adversarial multi-agent
> design review of #831 against the codebase and the Laudis client.


JeroenDeDauw:
@malberts ready?

malberts:
@JeroenDeDauw Ready after https://github.com/ProfessionalWiki/NeoWiki/pull/1004. There are some minor follow-ups to do here, which CC got too excited to post already.

PR #818, opened 2026-05-12
Fix prod/CI Makefile targets and run image build on PRs

malberts:
PR #811 made `install-db`, `load-neo4j-users`, and `wait-for-neo4j` depend on
`test_neo`, which is profile-gated to dev. Both CI (against `docker-compose.ci.yml`)
and the VPS upgrade flow (against `docker-compose.yml`) now fail at the 60s
`test_neo` wait. CI on master has been red since the merge.

Decouple `test_neo` from the shared targets:

- `wait-for-neo4j` and `load-neo4j-users` now wait/configure only `neo`.
- A new dev-only `setup-test-neo` target waits for and seeds `test_neo`. It is
  explicitly chained by `_first-run-seed` and `reset` (the only dev call sites).
- `EXEC_MW_ROOT`/`EXEC_MW` switch from `$(DC_DEV)` to `$(DC)`. `docker compose exec`
  attaches by project+service to the already-running `mediawiki` container, so
  this works in both dev and prod without overlay coupling.

Add two new targets:

- `update` for VPS upgrades: pull, up -d with `--profile server`, wait for db
  and neo readiness, then update.php.
- `smoke-test` wrapping `Docker/test.sh` for CI.

Update `build-docker.yml`: drop `working-directory: Docker` from the make steps
so they pick up the top-level Makefile, switch the start step to `make up` so
the docker compose project name matches subsequent make invocations, rename
the test step to `make smoke-test`. Add `pull_request` trigger so the build
runs (without ghcr push) on every PR and catches regressions like this one
before merge.

## Test plan

- [x] Full `make dev` cycle from a fresh `/tmp` checkout (port 8500). Wiki
      responds 200; `setup-test-neo` seeded test_neo.
- [x] Full prod sequence from a separate fresh `/tmp` checkout (port 8501):
      `make up && make install-db && make load-neo4j-users && make smoke-test
      && make import-demo-data`. Wiki responds 200.
- [x] `make reset` after a dev setup. Chain completes including the new
      `setup-test-neo` step.
- [x] `make update` end-to-end against a fresh prod stack (with Caddy ports
      remapped to 8080/8443 to avoid host conflicts on a dev machine).
      Pull + up + db/neo readiness wait + update.php all succeed.
- [ ] CI (this PR) goes green.

### DIFF
--- a/.github/workflows/build-docker.yml
+++ b/.github/workflows/build-docker.yml
@@ -4,12 +4,18 @@ on:
   push:
     branches:
       - master
+  pull_request:
   workflow_dispatch:
+
+concurrency:
+  group: build-docker-${{ github.ref }}
+  cancel-in-progress: true
 
 jobs:
 
   build_and_deploy_mediawiki:
     runs-on: ubuntu-latest
+    timeout-minutes: 20
     permissions:
       contents: read
       packages: write
@@ -47,15 +53,13 @@ jobs:
         run: rsync -a --exclude='Docker/' ./ Docker/mediawiki/extensions/NeoWiki/
 
       - run: make wiki-production-image
-        working-directory: Docker
 
       - name: Use CI docker-compose
         run: rm docker-compose.yml && mv docker-compose.ci.yml docker-compose.yml
         working-directory: Docker
 
       - name: Start containers
-        run: docker compose up -d
-        working-directory: Docker
+        run: make up
 
       - name: Wait for Neo4j
         uses: iFaxity/wait-on-action@v1.1.0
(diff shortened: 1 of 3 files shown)


alistair3149:
Heads-up — PR #820 renames `Docker/test.sh` → `Docker/tests/smoke.sh` (consolidating shell tests into `Docker/tests/`). The planned `smoke-test` target in this PR needs its path updated to match before / after merging #820, depending on which lands first.

> Coordination note from work on #820.
> <sub>Claude Code, `Opus 4.7`</sub>

malberts:
CI already fixed in https://github.com/ProfessionalWiki/NeoWiki/pull/838.
Prod is no longer a required target, since the live demo wiki ("prod") moved out of this repo. Mixing prod and dev into the same environment just complicates both.

Issue #591, opened 2026-02-14
REST API endpoint for updating a single Statement


JeroenDeDauw:
_Written by Claude Code, Opus 4.6. Based on discussion with @JeroenDeDauw. Context: the NeoWiki codebase, deliverables, and feature request_

A dedicated endpoint for setting or removing a single Statement of a Subject, without requiring a full Subject replace.

## Motivation

The full-replace `PUT /subject/{subjectId}` (see #280) is sufficient for the NeoWiki editing UI, which always has the complete Subject in memory. However, third-party integrators and programmatic consumers (bots, external systems) benefit from being able to update a single property without fetching and re-sending the entire Subject.

This also reduces the conflict surface: two systems updating different properties of the same Subject won't overwrite each other.

## Proposed endpoints

Since property names are unique within a Subject, a Statement is naturally addressable as a sub-resource:

**Set/replace a statement:**
```
PUT /neowiki/v0/subject/{subjectId}/statements/{propertyName}
```
```json
{
  "propertyType": "url",
  "value": ["https://example.com"]
}
```

**Remove a statement:**
```
DELETE /neowiki/v0/subject/{subjectId}/statements/{propertyName}
```

### Open questions

- Should `propertyType` be required, or can it be looked up from the Schema? The current design stores the "writer's schema" on each Statement, which argues for requiring it. But defaulting to the current Schema type would simplify the client experience.
- Property names with special characters need URL encoding. Worth noting in API docs.

## Context

- Follows up on the `PUT` refactor in #280
- "Web APIs for third-party usage, including queries and CRUD" is part of the BlueSpice deliverables

JeroenDeDauw:
Part of an ECHOLOT requirement: the API-connector one; this adds statement-level programmatic access: [FR4.4 — provide connectors (APIs) to the ECHOLOT system so that users can connect their own applications or AI-agents to ECHOLOT instances programmatically](https://github.com/ECHOLOT-ECCCH/echolot-WP2/issues/38).

PR #822, opened 2026-05-14
Add backend validation foundation

alistair3149:
Implements the foundation tier of [ADR 21: Add Backend Validation](https://github.com/ProfessionalWiki/NeoWiki/blob/master/docs/adr/021_Add_Backend_Validation.md). Adds a `SubjectValidator` service, extends the `PropertyType` plugin contract with a `validate()` method, and exposes two new REST endpoints that return constraint violations without performing a write.

Follows-up to #819 (re-uses the post-PUT-refactor patterns for `Application/` use cases and `EntryPoints/REST/` adapters).

## Summary

- Two new endpoints: `POST /neowiki/v0/subject/validate` (create-shape body) and `POST /neowiki/v0/subject/{subjectId}/validate` (update-shape body). Both return `200 {violations: [...]}` on well-formed input — violations are not HTTP errors.
- Plugin-contract change: `PropertyType` interface now requires `validate(NeoValue, PropertyDefinition): Violation[]`. Implemented on all six built-in types and on RedHerb's `ColorType` (canonical extension example).
- `docs/ValidationCodes.md` is the shared-codes contract between PHP and TS, including known gaps.

## Scope

**In:** PHP-side validator + endpoints + plugin contract change. Foundation tier only.

**Out (deferred):** Edit endpoints unchanged (no validator on `POST`/`PUT /subject`); no enforcement config; no pre-existing-vs-new violation differentiation; TS top-level validator still returns boolean (not violations); no UI surface for the new endpoints.

### `POST /neowiki/v0/subject/validate`

Body matches `POST /subject` create:

```json
{ "schema": "Company", "label": "ACME", "statements": { ... } }
```

`200 {violations: [...]}` on well-formed input; `400` for missing/malformed body fields; `404` for non-existent `schema`.

### `POST /neowiki/v0/subject/{subjectId}/validate`

Body matches `PUT /subject/{subjectId}`:

```json
{ "label": "New", "statements": { ... }, "comment": "ignored" }
```

`200 {violations: [...]}` on well-formed input; `400` for invalid subject-id format; `404` for non-existent Subject; `200` with a `schema-not-found` violation when the Subject's stored Schema has gone missing (asymmetry with the create endpoint is intentional and explained in the handler).

### Response shape

```json
{
  "violations": [
    { "propertyName": "Websites", "code": "invalid-url", "args": [], "valuePartIndex": 1 },
    { "propertyName": null, "code": "label-required", "args": [] }
  ]
}
```

`propertyName` is the string form (or `null` for Subject-level violations). `args` is always present. `valuePartIndex` is omitted when not applicable. The stable code set is documented in `docs/ValidationCodes.md`.

## Code structure

- `src/Domain/Validation/Violation.php` — immutable VO with `withPropertyName()`. `SubjectValidator` uses the builder to attach the property name (which PropertyType plugins don't know themselves).
- `src/Application/Validation/SubjectValidator.php` — pure orchestration. Algorithm mirrors the TS `SubjectValidator`: label check → iterate statements → skip unknown-to-schema properties (schema-drift accommodation) → delegate to `PropertyType::validate()`.
- `src/Domain/PropertyType/PropertyType.php` — interface gains `validate(NeoValue, PropertyDefinition): Violation[]`. Built-in implementations port the rules from `resources/ext.neowiki/src/domain/propertyTypes/*.ts`.
- `src/Domain/Subject/SubjectLabel.php` — gains `createForValidation(string)` static that bypasses the empty-rejection constructor via reflection. The closure-`->call($instance)` alternative was rejected by PHPCS; reflection is the equivalent that passes the standard. Used only by the validate path; the regular constructor still rejects empty labels for write paths.
- `src/Domain/Schema/Property/DateTimeProperty.php` — gains public static `parseStrictDateTime(string): ?DateTimeImmutable` so `DateTimeType::validate` can reuse the existing strict-ISO-8601 + calendar-overflow logic.
- `src/Application/SelectStatementResolver.php` — gains `resolveOrLeave()`, a non-throwing variant used only by the validate handlers. Unresolvable select values pass through as-is; `SelectType::validate` then surfaces them as `invalid-option` violations rather than the handler erroring out. The throwing `resolve()` is unchanged for write paths.
- `src/EntryPoints/REST/ValidateSubjectApi.php` + `ValidateSubjectUpdateApi.php` — thin adapters. `needsWriteAccess()` returns false. No CSRF (read auth).
- `extension.json` — two new `RestRoutes` entries; `ModuleSpecHandler` picks them up into the OpenAPI spec automatically.
- `tests/RedHerb/src/ColorType.php` — real hex-color validation replacing the stub, as the canonical plugin-contract example.

Pre-1.0 stability — `docs/RestApi.md` already disclaims that `/neowiki/v0/*` is not stable. No back-compat shims.

## Test plan

- [x] All new unit tests pass: `ViolationTest`, `SubjectValidatorTest`, the six `*TypeValidateTest` classes, `DateTimePropertyTest`, `SubjectLabelTest`, `ColorTypeValidateTest`, `SelectStatementResolverTest`.
- [x] `make phpunit filter=ValidateSubjectApiTest` — 9/9.
- [x] `make phpunit filter=ValidateSubjectUpdateApiTest` — 9/10 (1 skipped: `testSubjectWithMissingSchemaReturnsSchemaNotFoundViolation` — fixture cannot create a findable Subject with a never-existing Schema; the defensive code path is documented).
- [x] `make phpunit filter=EntryPoints` — 249/249 (1 skip).
- [x] `make phpunit filter=ModuleSpecHandlerNeoWikiTest` — both new routes appear in the OpenAPI spec (assertion count rose from 376 to 407).
- [x] `make cs` clean (phpcs + phpstan).
- [x] **Live API smoke** — both endpoints exercised against demo data; one bug was caught and fixed mid-review (`c734dd96`).

## Manual API Check

Demo data needs to be present first: `make import-demo-data`. Replace `localhost:8484` with your wiki URL.

### Create-shape endpoint

```sh
# Happy path
curl -sS -w '\nHTTP %{http_code}\n' -X POST "http://localhost:8484/rest.php/neowiki/v0/subject/validate" \
  -H 'Content-Type: application/json' \
  -d '{"schema":"Company","label":"ACME","statements":{"Status":{"propertyType":"select","value":["Active"]}}}'
# Expect: 200 {"violations":[]}

# Empty label
curl -sS -w '\nHTTP %{http_code}\n' -X POST "http://localhost:8484/rest.php/neowiki/v0/subject/validate" \
  -H 'Content-Type: application/json' \
  -d '{"schema":"Company","label":"","statements":{"Status":{"propertyType":"select","value":["Active"]}}}'
# Expect: 200 with a single label-required violation

# Invalid URLs at indices 1 and 2
curl -sS -w '\nHTTP %{http_code}\n' -X POST "http://localhost:8484/rest.php/neowiki/v0/subject/validate" \
  -H 'Content-Type: application/json' \
  -d '{"schema":"Company","label":"ACME","statements":{"Status":{"propertyType":"select","value":["Active"]},"Websites":{"propertyType":"url","value":["https://example.com","not a url","ftp://nope"]}}}'
# Expect: 200 with two invalid-url violations at valuePartIndex 1 and 2

# Invalid select option (regression for the 500 bug fixed in c734dd96)
curl -sS -w '\nHTTP %{http_code}\n' -X POST "http://localhost:8484/rest.php/neowiki/v0/subject/validate" \
  -H 'Content-Type: application/json' \
  -d '{"schema":"Company","label":"ACME","statements":{"Status":{"propertyType":"select","value":["bogus-id"]}}}'
# Expect: 200 with invalid-option violation

# Non-existent schema
curl -sS -w '\nHTTP %{http_code}\n' -X POST "http://localhost:8484/rest.php/neowiki/v0/subject/validate" \
  -H 'Content-Type: application/json' \
  -d '{"schema":"NoSuchSchema","label":"X","statements":{}}'
# Expect: 404
```

### Update-shape endpoint

Find a subject ID first (e.g. via `curl http://localhost:8484/rest.php/neowiki/v0/page/{pageId}/subjects`), then:

```sh
SID='sFromYourWiki00'

# Happy path
curl -sS -w '\nHTTP %{http_code}\n' -X POST "http://localhost:8484/rest.php/neowiki/v0/subject/$SID/validate" \
  -H 'Content-Type: application/json' \
  -d '{"label":"New label","statements":{}}'
# Expect: 200 with violations matching the schema/statements

# Empty label
curl -sS -w '\nHTTP %{http_code}\n' -X POST "http://localhost:8484/rest.php/neowiki/v0/subject/$SID/validate" \
  -H 'Content-Type: application/json' \
  -d '{"label":"","statements":{}}'
# Expect: 200 with label-required violation

# Non-existent subject (valid-format ID)
curl -sS -w '\nHTTP %{http_code}\n' -X POST "http://localhost:8484/rest.php/neowiki/v0/subject/s1aaaaaaaaaaaaa/validate" \
  -H 'Content-Type: application/json' \
  -d '{"label":"X","statements":{}}'
# Expect: 404

# No write should occur: re-read the subject after the calls above and confirm
# the label is still its original value.
curl -sS "http://localhost:8484/rest.php/neowiki/v0/subject/$SID"
```

## Plugin-contract breaking change

The `PropertyType` interface gains a required `validate()` method. Extension authors maintaining custom Property Types must implement it. The RedHerb `ColorType` is updated in this PR as the canonical example.

## Known limitations

Documented in `docs/ValidationCodes.md`. Intentional gaps in the foundation tier:

- **Missing-required.** Schema properties with no Statement at all are not flagged with `required` — mirrors the TS behavior. To be fixed when the TS top-level validator is upgraded.
- **Empty-Statement-required.** `StatementListBuilder` drops empty-valued Statements before validation; same fix as above.
- **`TextType` min-length / max-length.** PHP `TextProperty` doesn't yet expose those fields; TS does.
- **`RelationType` invalid-subject-id.** PHP `SubjectId` validates at construction, so `RelationValue` cannot hold an invalid target; PHP doesn't emit this code.

> Result of extensive collaborative work with @alistair3149 — brainstorming (where we considered #591 vs. backend validation vs. conditional writes and settled on the latter being a precursor to all three), design, spec, plan, execution, review feedback, and live smoke testing that caught one real defect.
> Context: NeoWiki codebase, ADR 21, the TS PropertyType implementations and tests, and PR #819 conventions. Working design + plan at `docs/planning/2026-05-14-backend-validation-foundation-{design,plan}.md` (uncommitted per project policy).


JeroenDeDauw:
@alistair3149 I did not verify these points yet beyond reading them and thinking they sound sensible.

These are notable findings, I wonder why you did not catch them. This was my prompt:

> Read everything in neowikis docs/, esp the adrs, and everything in the project context/. Then review https://github.com/ProfessionalWiki/NeoWiki/pull/822 by posting here on cli, focusing on notable issues and design/architecture.                                                                                                                                                

---

> Posting at @JeroenDeDauw's request — he directed this review and selected these three points from a longer analysis to surface here.
> Context: the NeoWiki codebase and ADRs (esp. 1, 11, 12, 21), the repo's reference docs, and a side-by-side comparison of this PHP validation against the existing TS validators.

Three design/architecture points, roughly in priority order.

## A. Orchestration lives in the REST handlers rather than an Application action

`ValidateSubjectApi::run()` and `ValidateSubjectUpdateApi::run()` do the full schema-lookup → `selectStatementResolver->resolveOrLeave()` → `statementListBuilder->build()` → assemble/mutate a `Subject` → validate flow inline. Compare `ReplaceSubjectApi` (`src/EntryPoints/REST/ReplaceSubjectApi.php:33-39`), which stays thin and delegates all of that to `ReplaceSubjectAction` in the Application layer.

So these handlers carry application-layer orchestration in `EntryPoints/`, which cuts against the domain-centric layering in ADR 1 and is inconsistent with the sibling write endpoints. A `ValidateSubjectAction` / `ValidateSubjectUpdateAction` returning `Violation[]`, with the handlers reduced to body-parsing + serialization + error mapping, would restore the pattern and remove the duplicated resolve→build dance between the two handlers. It would also retire the `PLACEHOLDER_SUBJECT_ID` workaround, which only exists because the create path validates a fully-constructed `Subject` rather than the proposed input.

## B. `required` is largely unreachable through the endpoints

`SubjectValidator` only iterates statements present on the Subject; it does not iterate the Schema's properties. And `StatementListBuilder::build()` drops any statement whose value `isEmpty()` before the validator runs (`src/Application/StatementListBuilder.php:45-47`), where `StringValue::isEmpty()` is `strings === []`. Net effect through the API:

- absent required property → no violation
- required property sent as `value: []` → dropped by the builder → no violation
- `required` only actually fires for a statement that survives the builder but is still content-empty (e.g. `[""]` or whitespace-only text/url, or a wrong-typed value)

That last window is not what a consumer sends on purpose, so the common case ("I omitted a required field") silently passes. `ValidationCodes.md` does document the missing-required and empty-statement gaps — the ask here is mostly about prominence. Given ADR 21's goal of letting the API reject edits that would make Subjects invalid, a `required` code that is present-but-mostly-unreachable can create false confidence. Worth surfacing the limitation in the endpoint docs themselves (not only in `ValidationCodes.md`), or reconsidering whether `required` should ship before the schema-iteration round.

## C. Type-mismatch invalidity is silently treated as valid (and isn't in the known-limitations list)

`SubjectValidator` resolves the PropertyType from the **writer's** type (`$statement->getPropertyType()`) but passes the **current** `PropertyDefinition`, and every `validate()` opens with `if ( !$definition instanceof XProperty ) return [];`. So when a Schema property's type changes (say `url` → `number`), the stored url-typed statement is handed to `UrlType::validate()` against a `NumberProperty`, hits the guard, and returns no violations.

That is the exact scenario ADR 11 describes (a property changing to a type the existing value cannot satisfy should surface as invalid), and ADR 12 explicitly named type mismatch as a form of invalidity. Unlike the other gaps, this one is not in `ValidationCodes.md`'s known-limitations. Suggest either documenting it as a deliberate foundation-tier gap, or deciding whether a dedicated `type-mismatch` signal belongs here. Related: out-of-schema statements are `continue`-skipped and likewise produce no violation — defensible as schema-drift tolerance, but worth a line given ADR 8.

src/Application/SelectStatementResolver.php
```diff
@@ -95,6 +95,71 @@ private function resolveSingle( string $propertyName, SelectProperty $property,
...
+	 * Like resolve() but never throws. When a select value cannot be resolved to an
+	 * option ID (unknown ID and unknown label), the original value is left in place
+	 * so downstream validation can flag it as an invalid-option violation rather
+	 * than aborting the entire request.
+	 *
+	 * @param array<string, mixed> $statements
+	 *
+	 * @return array<string, mixed>
+	 */
+	public function resolveOrLeave( Schema $schema, array $statements ): array {
```
JeroenDeDauw, line 126: Do we need both the throwing and non-throwing forms?

src/Domain/PropertyType/Types/DateTimeType.php
```diff
@@ -29,4 +33,67 @@ public function buildPropertyDefinitionFromJson( PropertyCore $core, array $prop
...
+			if ( $max !== null && $parsed > $max ) {
+				$violations[] = new Violation(
+					propertyName: null,
+					code: 'max-value',
+					args: [ $maximumString ],
+				);
+			}
+		}
+
+		return $violations;
```
JeroenDeDauw: This function is rather big

src/Application/SelectStatementResolver.php
alistair3149: Yes, for now. Different error policies for two genuinely different needs:

- `resolve()` throws → write paths (`Create`/`ReplaceSubjectAction`) wrap it to a 400.
- `resolveOrLeave()` is silent → validate paths leave unresolvable values in place so `SelectType::validate()` surfaces them as `invalid-option` violations with the correct `valuePartIndex` (which an exception can't carry).

The duplication is the foundation-tier cost of not running the validator on write paths. Once the enforcement tier does, the throwing form becomes redundant (validator's `invalid-option` check produces the same rejection) and the two collapse to one. Documented the rationale on the class in 69cf8b02.

src/Domain/PropertyType/Types/DateTimeType.php
alistair3149: Done in https://github.com/ProfessionalWiki/NeoWiki/pull/822/commits/06d2965209377fc4171c16c68c8942840ee40d96

PR #823, opened 2026-05-14
Hide RedHerb subject-editing sidebar links from users without permission

### LINKED ISSUE #821: "Edit main subject" tool link should only show to those with edit subject permissions
Image

### LINKED ISSUE #824: RedHerb "Create child Company" sidebar link should respect create-child-subject permission
Follow-up to #821 / #823.

The RedHerb sidebar's `Create child Company` link (`RedHerbSidebarHook`) is rendered on every existing page regardless of whether the current user may create child subjects. This is the same class of problem fixed for `Edit main subject` in #823, which gated that link on `SubjectAuthorizer::canEditSubject()`.

`Create child Company` should be gated on `SubjectAuthorizer::canCreateChildSubject()`.

RedHerb is the canonical example extension demonstrating NeoWiki extension points, so it should model correct permission handling rather than the bug.

Identified with Claude Code

malberts:
Fixes https://github.com/ProfessionalWiki/NeoWiki/issues/821
Fixes https://github.com/ProfessionalWiki/NeoWiki/issues/824

The "Edit main subject" and "Create child Company" sidebar links were
rendered without checking whether the current user could perform the
action. On wikis where anonymous editing is disabled, anonymous users
saw tools they could not use.

Gate "Edit main subject" on `SubjectAuthorizer::canEditSubject()` and
"Create child Company" on `SubjectAuthorizer::canCreateChildSubject()`,
so each link only shows when the current user can perform that action.

### DIFF
--- a/tests/RedHerb/src/RedHerbSidebarHook.php
+++ b/tests/RedHerb/src/RedHerbSidebarHook.php
@@ -6,6 +6,7 @@ namespace ProfessionalWiki\RedHerb;
 
 use Closure;
 use MediaWiki\Hook\SidebarBeforeOutputHook;
+use MediaWiki\Permissions\Authority;
 use MediaWiki\Title\Title;
 use ProfessionalWiki\NeoWiki\Domain\Page\PageId;
 use ProfessionalWiki\NeoWiki\NeoWikiExtension;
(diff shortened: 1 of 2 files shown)


malberts:
Superseded by #828, which implements the same fixes for #821/#824 without the closure-injection idiom (used nowhere else in the codebase) and additionally removes the pre-existing closure. Closing in favour of #828.

PR #837, opened 2026-05-16
Add Boolean property type

### LINKED ISSUE #703: Add Boolean property type

malberts:
Fixes https://github.com/ProfessionalWiki/NeoWiki/issues/703

## Summary

Adds a new `boolean` property type for true/false values, following the established core property type pattern (most recently the select type, #711). `BooleanValue` and the `ValueType::Boolean` deserialization path already existed in both PHP and TS; this PR adds the property-type wrapper, definition, UI, and registration.

Schema JSON format: `{ "type": "boolean" }`. No type-specific attributes; `default` may be `true`, `false`, or `null`.

## Screenshots

| | |
|---|---|
| **Schema editor: type dropdown.** "Boolean" appears in the type-picker dropdown with its check icon, alongside Text/URL/Number/Select/Relation/Date & Time/Date. | 703-4-type-dropdown |
| **Schema editor: Boolean property attributes.** `BooleanAttributesEditor` is intentionally empty (nothing renders between "Require a value" and "Initial value"); the initial-value control is the same `CdxToggleSwitch` used in the subject editor. | 703-5-boolean-property-editor |
| **Schema read view.** The `Is public · Boolean · No · Yes` row confirms PHP serialization of the Boolean property and its default round-trip and render. | 703-3-schema-table |
| **Subject edit: `BooleanInput` value editor.** "Is public" rendered as `CdxToggleSwitch` on the ACME Inc subject form, currently on. | 703-2-subject-editor-toggle |
| **Subject edit: full dialog.** The whole subject editor dialog with the toggle visible at the bottom of the form. | 703-2b-subject-editor-full |
| **Subject display: `BooleanDisplay`.** The infobox shows **Is public: Yes** after toggling and saving (full PHP/Neo4j round-trip). | 703-1-infobox-display-yes |

## Design decisions

- Type name is `boolean` (not `checkbox`/`toggle`): it names the data, not a widget, so the UI can change without a schema migration.
- Editor uses an inline toggle switch (label + switch in a single row); display renders the value as "Yes" / "No" text.
- `BooleanInput` always emits a `BooleanValue`, so a boolean is never "unset"; this matches checkbox/toggle semantics, and the input renders as a single inline `label [switch]` row matching the schema editor's `Require a value` row. One residual cosmetic: the schema editor's **Require a value** switch still renders for Boolean properties because it's owned by the shared `PropertyDefinitionEditor` chrome rather than gated per type. It has no effect on Boolean validation (the toggle always satisfies `required`), so it reads oddly but is harmless. Hiding it cleanly would need a small per-type opt-out hook on `TypeSpecificComponentRegistry` or `BasePropertyType`; deferred as a follow-up so the same hook can serve any future always-has-a-value types.

## Test plan

- [x] PHP tests pass (`BooleanPropertyTest`: serialization round-trips, non-boolean default rejected); full suite **1032 tests** green.
- [x] TS tests pass (`Boolean.spec.ts`: type, factory, validation); full suite **875 tests** green.
- [x] PHPCS + PHPStan clean; ESLint + Stylelint + build clean.
- [x] New tests fail when the production code is reverted.
- [x] Test review (mutation testing): **6/6 mutations caught** across both test files (default-validation removal/inversion, wrong type name, required-error removal, example-value flip, `===` to `!==`).
- [x] UI (Playwright), schema editor: "Boolean" in type selector; no spurious attributes; Initial value is a toggle; saved schema persists and the schema table shows "Yes".
- [x] UI (Playwright), subject editor: "Is public" renders as a toggle on the ACME Inc Company subject; toggled on, saved, and the infobox now displays "Is public: Yes" (full PHP/Neo4j round-trip). 0 console errors.


malberts:
I went with the toggle, because that's what we use elsewhere for boolean inputs. However, I have a feeling there's a slight difference between "toggling config" and "setting a yes/no value". With the latter falling more into checkbox category for me. On the other hand, mixing toggles and checkboxes might look odd, and in this case the checkbox would not be for selecting items in a list.

In either case, for displaying the value I went with "yes/no" text, because the rest of the infobox is also text. We could go with some kind of checkmark/cross icon, but that would be inconsistent with what we have now.

JeroenDeDauw:
Sidenote: every time i see those codex toggles with the text on the left and the toggle on the far right i think they look bad

image

alistair3149:
> Sidenote: every time i see those codex toggles with the text on the left and the toggle on the far right i think they look bad

There's an option to have it next to text. They look nicer when there is a description field.

On a form with multiple toggle switches, it's more scannable and less ugly if they are all aligned at the end.

alistair3149:
> I went with the toggle, because that's what we use elsewhere for boolean inputs. However, I have a feeling there's a slight difference between "toggling config" and "setting a yes/no value". With the latter falling more into checkbox category for me. On the other hand, mixing toggles and checkboxes might look odd, and in this case the checkbox would not be for selecting items in a list.
> 
> In either case, for displaying the value I went with "yes/no" text, because the rest of the infobox is also text. We could go with some kind of checkmark/cross icon, but that would be inconsistent with what we have now.

I had a deep dive on the Codex docs, and you are right. It is recommended to use checkbox over toggle switch when the option does not trigger immediate response (e.g. having a submit button): https://doc.wikimedia.org/codex/main/style-guide/constructing-forms.html#field-type-usage

In that case, we should switch to checkbox for forms. I can file up a follow-up issue for adapting checkbox. This is not a blocker for this PR.

alistair3149:
Code review PR: #848

malberts:
I'll do a follow-up checkbox version for this property so we can compare.

I suppose it is generally a good idea to follow the Codex usage guidelines. But my understanding of that explanation then is that our Schema Editor should also use checkboxes, since the changes are non-instant and require a save button. My impression is also that "non-instant" includes dynamic form element changes, but which do not actually persist and update the global UI (as opposed to the form). Not 100% confident, though.

Swapping all toggle switches to checkboxes might be trivial, but also seems like unnecessary change right now.

PR #844, opened 2026-05-20
Add drag to promote/demote/reorder subjects in the Subject Management UI

### LINKED ISSUE #783: Drag to reorder subjects in the Subject Management UI
> Drafted by Claude Code based on the design described by @alistair3149.
> Context: the NeoWiki codebase, the Subject Management UI from #750, and the
> existing drag-to-reorder pattern in `SchemaEditor` / `PropertyList` /
> `useSortable`.

Follows-up to #750.

The Subject Management UI today lets users add, view, edit, delete, promote, and
demote subjects, but the order of child subjects is fixed by insertion order. Users
should be able to drag a row to reorder it, and the new order should be persisted
on the page's subject data.

## Capabilities

- **Reorder child subjects.** Drag a child row to a new position in the child list;
  the new ordering is saved to the page's subject data.
- **Promote via drag onto the main slot.** Dragging a child row onto the main
  subject slot:
  - If there is no main subject, the dragged child is promoted to main.
  - If a main subject already exists, the dragged child is promoted to main *and*
    the previous main is demoted into the child list at the slot the dragged child
    came from (i.e. they swap roles).
- **Demote via drag off the main slot.** Dragging the main row into the child list
  demotes it to a child at the drop position. (Mirror of the promote case; include
  for symmetry — confirm whether we want this in scope.)

## Implementation notes

- Reuse the existing drag pattern from `SchemaEditor` / `PropertyList`: the
  `useSortable` composable (wrapping `sortablejs`), `cdxIconDraggable` as the
  handle, and the existing ghost-class styling. Same look and feel as schema
  property reordering.
- `useSortable` currently wraps a single container. The promote/demote-via-drag
  cases cross between the main slot and the child list, so the composable will
  need to gain a `group` option (SortableJS supports cross-container drag via
  groups) before this can ship. Single-list reorder works without that change.

## Out of scope / open questions

- **Touch and keyboard parity.** Drag-and-drop alone isn't accessible. We need an
  equivalent keyboard affordance (e.g. "move up / move down" in the row's overflow
  menu) and the drag handle should be reachable on touch. Confirm whether that
  ships in this issue or as a follow-up.
- **Backend.** `PageSubjects` has no reorder operation today (`SubjectMap` keeps
  insertion order, and the REST surface only supports create/update/delete and
  setting the main). A new endpoint or an extension of an existing one is needed
  to persist a new child order in a single request. Concurrent-edit semantics
  should match the rest of the subject endpoints.
- **Edit summary.** Reordering generates a revision; we need a default edit
  summary message and decide whether the promote-via-drag case should reuse the
  existing main-subject-change summary.

## Acceptance

- Reordering child subjects persists across page reload.
- Promote-via-drag onto an occupied main slot results in a swap, not a lost
  subject.
- Visual feedback matches the existing schema property reordering pattern.
- Works on mouse; touch and keyboard parity is at minimum tracked as a follow-up.

alistair3149:
Fixes https://github.com/ProfessionalWiki/NeoWiki/issues/783.

Follows-up to #750.

https://github.com/user-attachments/assets/1a0e86e6-831a-4c5c-b25a-a07919cc08d3

## Summary

Adds drag affordances to the Subject Management UI for promoting,
demoting, and reordering subjects, alongside the existing pin button:

- Drop a child onto the main slot to promote it. The previous main
  lands in the dragged child's original slot (swap-into-position).
- Drop the main row into the child list at any position to demote it.
- Drag a child within the child list to reorder.

All three interactions describe a desired end-state ordering of the
page's subjects and route through one new state-replacement REST
endpoint. One user action equals one revision.

## New endpoint: `PUT /page/{pageId}/subjectsOrdering`

Body: `{ mainSubjectId: string | null, childSubjectIds: string[], comment?: string }`.

Strict membership validation — the union of `mainSubjectId` (if
non-null) and `childSubjectIds` must equal the current set of subject
IDs on the page, no duplicates. Mismatches return 400. The endpoint is
for ordering only, not membership.

No-change detection — identical state returns 200 `unchanged`, no
new revision.

Same CSRF / auth as the existing v0 endpoints. The existing
`PUT /mainSubject` stays for the pin button.

## Implementation notes

- `SubjectMap::withOrdering( SubjectId[] )` and
  `PageSubjects::setOrdering( ?SubjectId, SubjectId[] )` are new
  domain operations. The serializer already preserves order through
  the JSON-object-property-order pipeline; no schema change.
- `useSortable` gains `group` / `sort` / `onDropIn` options for
  cross-container drag, and attaches via a reactive watcher so
  containers rendered behind a v-if (after an async load) get
  sortablejs wired up when they appear. Within-container reorder
  stays in `onEnd` after the DOM revert, so the consumer's reactive
  update lands on a clean DOM.
- New `useSubjectDrag` composable owns the cross-container wiring
  between the main slot and child list (shared group name,
  `sort: false` on the main slot). Exports a shared
  `subjectRowDomId( idText )` helper used by both the template's
  row id and the parser.
- `SubjectsManagerPage` computes the full desired ordering on the
  client for every drag and calls a single
  `SubjectStore.setPageSubjectsOrdering`. Swap-into-position splices
  the previous main into the dragged child's original slot; demote
  inserts the previous main at the drop position.

## Manual Browser Check

Page: any page with a main subject and 2+ child subjects. Demo data
ships with single-subject pages; create children via "Add subject" to
set up. Step 6 covers the empty / no-main cases.

1. **Drag-to-reorder.** Drag a child within the child list. Order
   persists across reload. "Subjects reordered" toast fires.
2. **Drag-to-promote with swap-into-position.** Drag a child at
   index N onto the main slot. The child becomes main; the previous
   main lands at index N in the child list.
3. **Drag-to-demote at position.** Drag the main row into the child
   list at index N. Page has no main; previous main appears at index N.
4. **Existing pin / edit / delete buttons.** Unchanged.
5. **SchemaEditor reorder.** Drag a property in any Schema's edit
   dialog. Order updates, Save button enables.
6. **Empty / no-main cases.** Drag-promote into empty main slot.
   Drag-demote into empty child list. Drag handles hidden for users
   without edit permission.

> <sub>Written by Claude Code, `Opus 4.7 (1M context)`.</sub>


malberts:
> Scope cut: no reorder within child list

I think this blocks it for me. Dragging UIs in lists generally imply this kind of behavior. At least, all our other implementations do that. Right now the dragging does exactly the same thing as the pin button, and then does not do anything else associated with dragging, which makes it look a bit broken. It also seems like more effort for the user, when there is a single click button next to it.

I am not suggesting we implement the reordering, though, as that is indeed not currently meaningful in our data model.

Since @JeroenDeDauw mentioned drag reordering in https://github.com/ProfessionalWiki/NeoWiki/issues/750#issuecomment-4298891364, I'll wait for his feedback first.

JeroenDeDauw:
Agree this is net-negative due to implied reorderability of child subjects which isn't supported. 

Now that we have the subject managment UI, I think ability to reorder has some value for some use cases. If people mainly look at and/or manage the data via this UI, then the order can matter. Allowing it via the UI does not seem harmful, people can alread do it via the API.

Very YAGNI: Later we might want predefined ordering abilities. If we have page-level schemas. (Related: https://www.wikibase.consulting/sorting-wikibase-statements-by-property/)

Simplest approach seems to be to add support for reordering and keep the UI as it already is in this PR

My main concern is people creating a bunch of small edits by reodering, but thisIsFine.png for MVP

alistair3149:
Updated the PR with reordering implemented.

JeroenDeDauw:
Video looks good. I checked some tests and they seem OK. Did not review further. Am wondering about 1200 LOC added

JeroenDeDauw:
@malberts

JeroenDeDauw:
@malberts you did ... something? ready to merge?

JeroenDeDauw:
@malberts

malberts:
The UI seems a bit glitchy. When I drop a row, it snaps back to where it was before it actually moves to the new position. I would expect the row to stay in the right place immediately. This happens on Firefox and Brave. However, this can be addressed in a follow-up.

[Screencast_20260619_234016.webm](https://github.com/user-attachments/assets/257a4772-b20d-4677-8044-1615f1244470)

PR #845, opened 2026-05-20
Fetch latest schema before opening subject editor

### LINKED ISSUE #713: Fetch latest schema before Subject edit
Follow up to
* https://github.com/ProfessionalWiki/NeoWiki/pull/708

alistair3149:
Tested manually in browser.

Fixes https://github.com/ProfessionalWiki/NeoWiki/issues/713
Follows-up to https://github.com/ProfessionalWiki/NeoWiki/pull/708

PR #708 made the Infobox fetch the latest subject before opening the
SubjectEditorDialog, but the editor also depends on the schema to
decide which property fields to render. If the schema was modified
in another tab, the editor still showed stale fields.

SubjectEditorDialog's watch on `props.subject` runs `loadSchema()` via
`schemaStore.getOrFetchSchema(...)` — a cache-aware read. Fetching the
schema in parallel with the subject and awaiting both before opening
the dialog keeps the cache fresh by the time the watch fires, so the
dialog renders against the latest schema.

## Manual Browser Check

- Open a subject page with an infobox in two browser tabs
- In tab 2, open the subject's schema and add a new property, then save
- In tab 1, click the edit button on the infobox
- Verify the new property field appears in the editor
- Repeat with property removal: in tab 2, delete a property from the
  schema and save; in tab 1, click edit and verify the removed field
  is gone

> Result of a clear spec follow-up to @JeroenDeDauw's PR #708.
> Context: the NeoWiki codebase and the parent PR.

### DIFF
--- a/resources/ext.neowiki/src/components/Views/Infobox.vue
+++ b/resources/ext.neowiki/src/components/Views/Infobox.vue
@@ -98,7 +98,10 @@ const schema = computed( () => schemaStore.getSchema( subject.value.getSchemaNam
 
 async function openEditor(): Promise<void> {
 	try {
-		await subjectStore.fetchSubject( props.subjectId );
+		await Promise.all( [
+			schemaStore.fetchSchema( subject.value.getSchemaName() ),
+			subjectStore.fetchSubject( props.subjectId )
+		] );
 		isEditorOpen.value = true;
 	} catch ( error ) {
 		mw.notify(
--- a/resources/ext.neowiki/tests/components/Views/Infobox.spec.ts
+++ b/resources/ext.neowiki/tests/components/Views/Infobox.spec.ts
@@ -32,7 +32,7 @@ describe( 'Infobox', () => {
 	} );
 
 	let pinia: ReturnType<typeof createPinia>;
-	let schemaStore;
+	let schemaStore: any;
 	let subjectStore: any;
 
 	const mockSchema = new Schema(
@@ -150,7 +150,8 @@ describe( 'Infobox', () => {
 		expect( editButton.exists() ).toBe( true );
 	} );
 
-	it( 'fetches latest subject and opens dialog when edit button is clicked', async () => {
+	it( 'fetches latest schema and subject before opening dialog when edit button is clicked', async () => {
+		schemaStore.fetchSchema = vi.fn().mockResolvedValue( undefined );
 		subjectStore.fetchSubject = vi.fn().mockResolvedValue( undefined );
 
 		const wrapper = mountComponent( mockSubject, true );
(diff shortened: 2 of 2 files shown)


JeroenDeDauw:
Can't we do those requets in parallel? Current code looks serial

alistair3149:
> Can't we do those requests in parallel? Current code looks serial

Yeah you are right. Updating the PR.

malberts (commented):
## Review

Approving. The fix is small, well-targeted, and I verified the user-visible behavior end-to-end with two-tab UI testing on a local NeoWiki dev stack.

### What it does
`Infobox.vue#openEditor` now warms **both** the schema and subject caches in parallel before flipping `isEditorOpen`. `schemaStore.fetchSchema()` unconditionally refetches and `setSchema()`s, so when `SubjectEditorDialog`'s `watch(() => props.subject, …, { immediate: true })` fires `loadSchema()` → `schemaStore.getOrFetchSchema()`, the cache is already fresh and returns the latest schema synchronously — no extra round-trip, no race.

### Correctness
- `subject.value.getSchemaName()` in `openEditor` is safe: the entire infobox (and therefore the edit button) is wrapped in `v-if="subject !== null"`, so `openEditor` cannot be invoked when `subject` is null.
- Schema-rename edge case (schema-name on the subject changes between fetches): the parallel `fetchSchema(oldName)` becomes a wasted fetch, but the dialog's own watcher then calls `getOrFetchSchema(newName)` on the freshly-refetched subject, which falls through to a network fetch on cache-miss — so the dialog still ends up rendering against the correct schema. Safe.
- `Promise.all` short-circuits on either rejection and the `try/catch` surfaces the error via `mw.notify`, keeping the dialog closed. Good failure mode.

### Tests
- Updated `Infobox.spec.ts` test is meaningful: I reverted just the production change and confirmed the new assertion fails with `Number of calls: 0` against the spy. So it would have caught any regression of the parent PR's pattern.
- Full vitest suite: **778/782** pass on PR; the 3 failures (`DateTimeInput`, `DateTimeDisplay`) reproduce identically on `origin/master`, are host-timezone-sensitive, and unrelated.
- `eslint` + `vue-tsc -b` clean on the changed files.

### Local UI verification (NeoWiki-845 worktree, `make dev`)
On `Sandbox Subject` (schema `Sandbox`, 5 properties), in a single browser context with two tabs:

1. Tab 2 → edit `Schema:Sandbox` → add a new property `PR845StaleSchemaCheck` → save.
2. Tab 1 (still showing the stale infobox with 5 props) → click the infobox edit pencil → dialog opens with **6 fields including `PR845StaleSchemaCheck`** ✅
3. Tab 2 → delete `PR845StaleSchemaCheck` → save.
4. Tab 1 → click edit again → dialog opens with **5 fields, the removed prop gone** ✅

### Nits (non-blocking)
- `let schemaStore: any` in the test — same shape as `subjectStore: any` two lines down, fine. A `ReturnType<typeof useSchemaStore>` would be tidier, but matching the existing style is the right call for this PR.
- `SchemaStore.fetchSchema`/`SubjectStore.fetchSubject` still don't dedupe concurrent in-flight requests. Not introduced by this PR and not a problem here (each `openEditor` click triggers exactly one of each), but worth keeping in mind if these are ever called from other reactive paths.

LGTM. 👍

PR #846, opened 2026-05-21
Switch #view to named-args with single positional Subject

### LINKED ISSUE #739: #view parser function: named arguments, only one positional
> Result of back-and-forth with @JeroenDeDauw exploring ergonomics of the `#view` parser function.
> Context: `src/EntryPoints/ViewParserFunction.php`, ADR 018 (Views), and the demo wikitext under `DemoData/Page/`.

## Problem

`{{#view}}` currently takes two positional arguments: Subject ID, then Layout name. This has two ergonomic issues:

1. **Awkward empty-arg form** for the common canonical-page pattern of "render my page's Main Subject with a specific Layout":
   ```
   {{#view: | Finances}}
   ```
   The leading pipe with nothing before it reads badly and is easy to get wrong.

2. **No room to grow.** ADR 018 anticipates per-View configuration beyond Layout (e.g. View Type overrides, per-instance display attribute overrides). Positional-only locks us in.

## Proposal

Support named arguments and reduce positional to a single slot for Subject ID.

### New signatures

```
{{#view:}}                                — Main Subject, default display
{{#view: s123}}                           — explicit Subject, default display (backwards compat)
{{#view: layout=Finances}}                — Main Subject with Layout
{{#view: s123 | layout=Finances}}         — explicit Subject with Layout
```

### Rationale

- **Backwards compatible** with the common `{{#view: SubjectId}}` form used across `DemoData/Page/`.
- **Fixes the awkward empty-arg case**: `{{#view: | Finances}}` becomes `{{#view: layout=Finances}}`.
- **No ambiguity**: the first positional is always a Subject ID; Layout is always flagged with `layout=`.
- **Reader-friendly**: `layout=` is self-documenting when someone opens the wikitext cold.
- **Room to grow**: future named args (`viewType=`, per-View settings) slot in without further signature churn.

### Alternatives considered

- **Remove all positional.** Cleaner, but throws away the terse `{{#view: s123}}` form for no real gain.
- **Make Layout the positional arg instead of Subject.** Optimizes the canonical-page pattern but breaks every existing `{{#view: SubjectId}}` call and creates ambiguity with Subject-ID-shaped strings.
- **Keep both positional forms alongside named.** Two ways to say the same thing; more to document and test.

## Scope

- Accept `subject=` and `layout=` as named arguments in `ViewParserFunction`.
- Keep first positional as Subject ID.
- Decide: hard-break the old `{{#view: SubjectId | LayoutName}}` form, or accept both during a transition. (Given NeoWiki's pre-release status and that the only in-tree callers are demo pages, hard-break seems fine — update the demo wikitext in the same PR.)
- Reject unknown named args with a parser-level error so typos fail loudly.
- Update `ViewParserFunctionTest` to cover the new forms.
- Update demo wikitext under `DemoData/Page/` to the new form (no `{{#view: | Foo}}` in tree currently, but worth a grep before merging).
- Update any user-facing docs that show the old form.

alistair3149:
Fixes https://github.com/ProfessionalWiki/NeoWiki/issues/739

https://github.com/user-attachments/assets/8467d953-a678-4bdc-86c2-2b49f389d734

`{{#view}}` now takes one positional Subject ID plus named arguments
(`subject=`, `layout=`). The old `{{#view: id | LayoutName}}` positional-Layout
form is hard-broken and renders a visible parser error, as do unknown named
arguments, conflicting subject specifications (positional + `subject=`),
and extra positional args.

## Accepted forms

| Wikitext | Result |
|---|---|
| `{{#view:}}` | Main Subject of page (existing) |
| `{{#view: s123}}` | explicit Subject, positional (existing) |
| `{{#view: subject=s123}}` | explicit Subject, named |
| `{{#view: layout=Finances}}` | Main Subject + Layout |
| `{{#view: s123 \| layout=Finances}}` | explicit + Layout |
| `{{#view: subject=s123 \| layout=Finances}}` | fully named |
| `{{#view: layout=Finances \| s123}}` | order-free |

Errors (rendered as `<div class="error">…</div>`):
* Two or more positional args (also catches the old positional-Layout form).
* Subject specified both positionally and via `subject=`.
* Unknown named arg (e.g. `layuot=Finances`).
* Argument with empty name (e.g. `=Finances`).

## Hard break

NeoWiki is pre-release; the only in-tree caller of the old positional-Layout
form was `DemoData/Module/SubjectRow.lua`, updated in the same PR. Once the
wiki re-imports that module, demo pages using it (e.g. `Research_catalog`,
`Subject_views`) will render correctly; until then those pages will display
the new parser errors for any subject-with-Layout entry.

## Notes for review

* Live smoke testing through `action=parse` against a running dev wiki
  surfaced one production bug the unit tests had missed: `<code>...</code>`
  markup in error message text was HTML-escaped by `Message::escaped()`,
  leaking literal `&lt;code&gt;` to users. Fixed by going plain-text-only
  (commit `6650c530`), matching the precedent in `CypherRawParserFunction`.
* New `ViewParserFunctionI18nTest` (commit `19650fbb`) locks this down:
  it asserts every `neowiki-view-error-*` message contains no `<`.
  Sanity-checked: temporarily reintroducing `<code>` makes it fail.
* `make cs` green (phpcs 359/359, phpstan 176/176 [OK]),
  `make phpunit filter=ViewParserFunction` green (17 tests, 54 assertions).

### DIFF
--- a/DemoData/AGENTS.md
+++ b/DemoData/AGENTS.md
@@ -54,8 +54,9 @@ Subject, relation, and option IDs:
    `NeoWikiHooks::handleContentPage` (`BeforePageDisplay`). Don't `{{#view}}` the same subject in
    the page's wikitext, or you get two infoboxes. Use `{{#view}}` only on plain `Page/` hubs (which
    have no Main Subject) or to embed a different subject.
-4. **`{{#view}}` uses named args for Layout.** Write `{{#view:id|layout=LayoutName}}`. The
-   old positional form `{{#view:id|LayoutName}}` is gone and now renders a visible parser error.
+4. **`{{#view}}` takes one positional Subject ID plus named args.** Write
+   `{{#view:id|layout=LayoutName}}` or `{{#view:subject=id|layout=LayoutName}}`. The old
+   positional form `{{#view:id|LayoutName}}` is gone and now renders a visible parser error.
 
 ## Cypher gotchas
 
--- a/DemoData/AGENTS.md
+++ b/DemoData/AGENTS.md
@@ -54,8 +54,8 @@ Subject, relation, and option IDs:
    `NeoWikiHooks::handleContentPage` (`BeforePageDisplay`). Don't `{{#view}}` the same subject in
    the page's wikitext, or you get two infoboxes. Use `{{#view}}` only on plain `Page/` hubs (which
    have no Main Subject) or to embed a different subject.
-4. **`{{#view:id|LayoutName}}` is positional.** The named form `{{#view:id|layout=LayoutName}}` is
-   silently ignored.
+4. **`{{#view}}` uses named args for Layout.** Write `{{#view:id|layout=LayoutName}}`. The
+   old positional form `{{#view:id|LayoutName}}` is gone and now renders a visible parser error.
 
 ## Cypher gotchas
 
--- a/DemoData/Module/SubjectRow.lua
+++ b/DemoData/Module/SubjectRow.lua
@@ -33,7 +33,7 @@ function p.render( frame )
 
 		local view
 		if viewLayout and viewLayout ~= '' then
-			view = frame:preprocess( '{{#view:' .. id .. '|' .. viewLayout .. '}}' )
+			view = frame:preprocess( '{{#view:' .. id .. '|layout=' .. viewLayout .. '}}' )
 		else
 			view = frame:preprocess( '{{#view:' .. id .. '}}' )
 		end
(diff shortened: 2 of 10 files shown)


JeroenDeDauw:
Looks like some things should be extracted to private methods in the tests. Certainly some assertions. Right now there is repeating of code that binds to production details, so if those details change all the tests would need to change

alistair3149:
> Looks like some things should be extracted to private methods in the tests. Certainly some assertions. Right now there is repeating of code that binds to production details, so if those details change all the tests would need to change

Updated. Also filed #847 as a separate follow-up issue because it is out of scope in this PR.

PR #850, opened 2026-05-23
Boolean property: use CdxCheckbox instead of CdxToggleSwitch

### LINKED ISSUE #849: Boolean property editor: use CdxCheckbox instead of CdxToggleSwitch
PR #837 introduced the Boolean property type. Its value editor (`BooleanInput.vue`) currently renders as a `CdxToggleSwitch`.

[Codex usage guidelines for ToggleSwitch](https://doc.wikimedia.org/codex/latest/components/demos/toggle-switch.html#when-to-use-togglesswitch) state:

> Use the ToggleSwitch component only where an instant change in the user-interface based on their assigned action is intended. For non-instant selections and selection groups, use Checkbox instead.

`BooleanInput` is used in two non-instant contexts:

- Subject editor: the value lands in a form that requires Save
- Schema editor "Initial value" control: also lives inside the property-definition form, gated by Save

Both should be `CdxCheckbox` per the Codex guideline.

Scope intentionally limited to `BooleanInput.vue`. Other `CdxToggleSwitch` usages in NeoWiki (Require a value, Text/URL/Select per-type attributes, etc.) are out of scope of this issue (see PR #837 thread for discussion).

Identified with Claude Code

malberts:
Closes #849.

## Context

PR #837 introduced the Boolean property type with `BooleanInput.vue` rendering as `CdxToggleSwitch`. Codex's [ToggleSwitch usage guideline](https://doc.wikimedia.org/codex/latest/components/demos/toggle-switch.html#when-to-use-toggleswitch) states:

> Use the ToggleSwitch component only where an instant change in the user-interface based on their assigned action is intended. For non-instant selections and selection groups, use Checkbox instead.

`BooleanInput` is used in two non-instant contexts:

- Subject editor (form gated by Save).
- Schema editor's "Initial value" control (also gated by Save inside the schema-editor dialog).

Both fit the "non-instant selection" case, so Checkbox is the guideline-correct widget.

## Change

Swap `CdxToggleSwitch` for `CdxCheckbox` in `BooleanInput.vue`. Same `:model-value` and `@update:model-value` wiring; no change to the emitted `BooleanValue(true/false)` shape, the checked-iff-`modelValue.boolean === true` rule, or the default-to-`false` behaviour for `undefined` / wrong-type input.

### Why the field labelling differs across call sites

`BooleanInput` renders in two places that semantically show different things:

- **Subject editor.** The checkbox *is* the property's value for this subject. The property name (e.g. `Is public`) is the natural label for the checkbox itself; `[ ] Is public` reads as "this subject's `Is public` is / isn't set". No separate top-of-field heading is needed.
- **Schema editor (Initial value field).** Here the Boolean isn't a value of the property; it's the *initial value* that will be used when a new subject is created. The field renders the way every other property type's initial-value widget renders: an "Initial value" heading at the top, the type's value widget (the checkbox) below.

So the same `BooleanInput` carries an inline checkbox label *always* (the property name) and a CdxField `#label` heading *conditionally*: shown when the caller-supplied `props.label` differs from the property name (the schema-editor case, where `props.label` is "Initial value"); suppressed when they match (the subject-editor case, where the caller passes the property name as the label and the heading would just repeat the inline checkbox label).

### Why CdxToggleSwitch didn't need this machinery

The toggle implementation in PR #837 had no CdxField heading. `align-switch="true"` rendered the toggle's slot content to the *left* of the switch, giving every row a `Label [O ]` shape. The schema-editor row read `Initial value [O ]` and visually rhymed with `Require a value [O ]` right above it; the widget itself handled the label-first layout, so no separate heading slot was needed.

`CdxCheckbox` has no equivalent of `align-switch`. It always renders as `[ ] Label`, box on the left and label to its right. A bare toggle→checkbox swap therefore makes the schema-editor row read `[ ] Initial value` hugging the left edge, with no top heading and no visual continuity with `Name` / `Type` / `Description` above. To restore a "heading on top, control below" rhythm in the schema editor, the `Initial value` text has to live somewhere other than the checkbox's own slot. That somewhere is the CdxField `#label` slot, conditionally shown as described above.

`:optional` is intentionally not set on the CdxField: a Boolean value is always defined ("never unset" per PR #837), so an "(optional)" affordance would be misleading.

## Screenshots

| | |
|---|---|
| **Schema editor: Boolean property editor.** The property-definition form shows the Boolean property's row with "Initial value" as a top heading (matching Name / Type / Description above) and `[ ] Is public` as the checkbox below. | 849-9-schema-editor-property-only |
| **Subject editor: full dialog.** ACME Inc edit dialog with `Is public` as the only boolean property; rendered as a checkbox. | 849-4-subject-editor-full |

## Test plan

- [x] TS unit tests for `BooleanInput` updated to assert `CdxCheckbox` (not `CdxToggleSwitch`), the inline checkbox label is the property name, and the heading-shown / heading-hidden branches both behave correctly. The behavioural assertions (modelValue prop binding, emit on update, BooleanValue emission, default-to-false, type-mismatch fallback, external-modelValue reactivity, `getCurrentValue`) are unchanged.
- [x] Updated tests fail against the toggle implementation and pass against the checkbox implementation.
- [x] Full TS suite passes.
- [x] Build clean; ESLint + Stylelint clean.
- [x] UI (Playwright), schema editor: "Is public" property's Initial value renders as `CdxCheckbox`; "Require a value" still renders as `CdxToggleSwitch` (unchanged).
- [x] UI (Playwright), subject editor (ACME Inc): "Is public" renders as `CdxCheckbox`; modelValue=true → checkbox checked; 0 console errors during interaction.

### DIFF
--- a/resources/ext.neowiki/src/components/Value/BooleanInput.vue
+++ b/resources/ext.neowiki/src/components/Value/BooleanInput.vue
@@ -2,13 +2,19 @@
 	<CdxField
 		:status="validationError === null ? 'default' : 'error'"
 		:messages="validationError === null ? {} : { error: validationError }"
-		:hide-label="true"
+		:hide-label="!showFieldHeading"
 	>
+		<template
+			v-if="showFieldHeading"
+			#label
+		>
+			{{ props.label }}
+		</template>
 		<CdxCheckbox
 			:model-value="internalValue"
 			@update:model-value="onInput"
 		>
-			{{ props.label }}
+			{{ props.property.name.toString() }}
 			<CdxIcon
 				v-if="props.property.description"
 				v-tooltip="props.property.description"
@@ -25,7 +31,7 @@ import type { Value } from '@/domain/Value';
 </script>
 
 <script setup lang="ts">
-import { ref, watch } from 'vue';
+import { computed, ref, watch } from 'vue';
 import { CdxCheckbox, CdxField, CdxIcon } from '@wikimedia/codex';
 import { cdxIconInfo } from '@wikimedia/codex-icons';
 import { newBooleanValue, BooleanValue, ValueType } from '@/domain/Value';
@@ -44,6 +50,10 @@ const props = withDefaults(
 const emit = defineEmits<ValueInputEmits>();
 
 const validationError = ref<string | null>( null );
+
+// Hide the heading when it would duplicate the inline checkbox label
+// (subject-editor case: the caller passes the property name as the label).
+const showFieldHeading = computed( () => props.label !== props.property.name.toString() );
 
 const toBoolean = ( value: Value | undefined ): boolean =>
 	value !== undefined && value.type === ValueType.Boolean ? ( value as BooleanValue ).boolean : false;
--- a/resources/ext.neowiki/src/components/Value/BooleanInput.vue
+++ b/resources/ext.neowiki/src/components/Value/BooleanInput.vue
@@ -4,10 +4,8 @@
 		:messages="validationError === null ? {} : { error: validationError }"
 		:hide-label="true"
 	>
-		<CdxToggleSwitch
+		<CdxCheckbox
 			:model-value="internalValue"
-			:align-switch="true"
-			:label="props.label"
 			@update:model-value="onInput"
 		>
 			{{ props.label }}
@@ -18,7 +16,7 @@
 				class="ext-neowiki-value-input__description-icon"
 				size="small"
 			/>
-		</CdxToggleSwitch>
+		</CdxCheckbox>
 	</CdxField>
 </template>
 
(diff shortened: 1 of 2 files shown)


JeroenDeDauw:
Should we also replace the toggles in the property definition editor with checkboxes (in a seperate PR)?

malberts:
> Should we also replace the toggles in the property definition editor with checkboxes (in a seperate PR)?

I increasingly lean towards doing that:
* The combination of toggle and checkbox on the same form looks inconsistent. It also seems like slight negative UX when the same form makes you move to the left and the right for the same "on/off" interaction. Granted, the toggle could be moved to the left, but still leaves the form looking inconsistent.
* As per [Codex docs](https://doc.wikimedia.org/codex/latest/components/demos/toggle-switch.html#when-to-use-toggleswitch), and reading up on this more widely now, toggles should update something immediately (whether UI or something in the system). This means we should not be using them on forms with submit buttons.
* We used only checkboxes in the past, so for me that was a more typical UI approach anyway.

JeroenDeDauw:
Sounds good to me. Don't forget follow up issue or PR

Issue #1, opened 2024-09-07
Add GH CI


malberts:
Run:
* [x] PHP tests
* [ ] https://github.com/ProfessionalWiki/NeoExtension/issues/12
* [ ] https://github.com/ProfessionalWiki/NeoExtension/issues/11

PR #855, opened 2026-05-27
Surface violations on Subject write endpoints

alistair3149:
Follows-up to https://github.com/ProfessionalWiki/NeoWiki/pull/822.

Implements step 1 of [ADR 21 backend validation](../tree/master/docs/adr/021-add-backend-validation.md): the two write endpoints — `POST /subject` and `PUT /subject/{id}` — now run `SubjectValidator` against the proposed Subject before persisting and include the resulting `Violation[]` in the success response body. Validation does not block the write; enforcement is a future round.

## What changed

- `CreateSubjectAction` and `ReplaceSubjectAction` each gain a `SubjectValidator` constructor dependency and run validation immediately before the repository write.
- `CreateSubjectPresenter::presentCreated` signature extends to `presentCreated(string $subjectId, array $violations)`.
- `ReplaceSubjectAction::replace` returns `Violation[]` instead of `void`.
- Both REST handlers emit `violations` (top-level array) in their `201`/`200` response bodies.
- Missing schema returns empty violations and proceeds with the write — preserves ADR 21's "invalid Subjects must stay editable" requirement.
- Create's `subject-already-exists` branch omits the `violations` key entirely (validation never ran).
- 14 new PHPUnit tests across Action and REST integration levels.

## Out of scope (future rounds)

- Enforcement / `$wgNeoWikiValidationEnforced` config.
- Pre-existing-vs-newly-introduced violation differentiation.
- TS `SubjectValidator` upgrade to return `Violation[]` at the top level.
- Frontend consumption.

## Test plan

- [x] `make phpunit filter=ReplaceSubject` — 23/23 green.
- [x] `make phpunit filter=CreateSubject` — 23/23 green (Action + REST).
- [x] `make phpunit filter=Application` — 270/270 green (no Application-layer regressions).
- [x] `make cs` — phpcs 392 files + phpstan 187 files, no errors.
- [x] Four Playwright smoke tests against the live demo wiki (POST + PUT × clean/violating). Raw response bodies captured during development — all four returned the expected shape:
  - `PUT /subject/s1demo1aaaaaaa1` (Status omitted): `200 {"status":"updated","subjectId":"s1demo1aaaaaaa1","violations":[{"propertyName":"Status","code":"required","args":[]}]}`
  - `PUT /subject/s1demo1aaaaaaa1` (Status set): `200 {"status":"updated","subjectId":"s1demo1aaaaaaa1","violations":[]}`
  - `POST /page/{id}/mainSubject` (Status omitted): `201 {"status":"created","subjectId":"...","violations":[{"propertyName":"Status","code":"required","args":[]}]}`
  - `POST /page/{id}/mainSubject` (Status set): `201 {"status":"created","subjectId":"...","violations":[]}`

### DIFF
--- a/docs/reference/validation-codes.md
+++ b/docs/reference/validation-codes.md
@@ -4,15 +4,19 @@ Constraint violations returned by NeoWiki's backend validation API use stable `c
 This document is the authoritative reference, shared between the PHP backend implementation and
 (in a future round) the TS frontend implementation.
 
-The API endpoints are:
+Violations are returned by:
 
-- `POST /neowiki/v0/subject/validate` — validates a proposed create-shape body.
-- `POST /neowiki/v0/subject/{subjectId}/validate` — validates a proposed update-shape body
-  against an existing Subject's Schema.
+- `POST /neowiki/v0/subject/validate` — dry-run validation of a proposed create-shape body.
+- `POST /neowiki/v0/subject/{subjectId}/validate` — dry-run validation of a proposed update-shape
+  body against an existing Subject's Schema.
+- `POST /neowiki/v0/subject` and `PUT /neowiki/v0/subject/{subjectId}` — the write endpoints include
+  the resulting `violations` array in their `201`/`200` success body. Validation does not block the
+  write (see [ADR 21](../adr/021-add-backend-validation.md)).
 
-Both endpoints return `200 OK` with a `{violations: [...]}` body whenever the request is
-well-formed. Violations present in the body do not change the HTTP status. `400` is reserved for
-malformed input; `404` for a missing Schema or Subject.
+The `/validate` endpoints return `200 OK` with a `{violations: [...]}` body whenever the request is
+well-formed; violations in the body do not change the HTTP status. `400` is reserved for malformed
+input, `404` for a missing Subject (update dry-run). A missing Schema is reported differently per
+endpoint — see [`schema-not-found`](#schema-not-found).
 
 Each violation in the response has this shape:
 
--- a/docs/reference/validation-codes.md
+++ b/docs/reference/validation-codes.md
@@ -166,14 +166,23 @@ A part does not match the 6-digit hex-color pattern (`/^#[0-9a-fA-F]{6}$/`).
 
 ### `schema-not-found`
 
-Emitted by `ValidateSubjectUpdateApi`.
+Emitted by `ValidateSubjectUpdateApi` and by the Subject write endpoints
+(`POST /subject`, `PUT /subject/{id}`) via `ProposedSubjectValidator`.
 `args`: `[schemaName]`.
 `valuePartIndex`: never set.
 `propertyName`: always `null` (Subject-level).
 
-The existing Subject's Schema cannot be loaded — usually because the Schema page was deleted or
-renamed since the Subject was created. Surfaced as a violation rather than a 404 because the
-Subject itself does exist; the caller can fix this by re-creating or renaming the Schema page.
+The Subject's Schema cannot be loaded — usually because the Schema page was deleted or renamed
+since the Subject was created, or because a Subject is created/imported referencing a Schema that
+does not (yet) exist. On the write endpoints it is non-blocking: the write proceeds and the
+(unvalidatable) Subject stays editable per ADR 21, but the response reports `schema-not-found`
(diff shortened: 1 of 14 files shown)


JeroenDeDauw:
> Posted at @JeroenDeDauw's request, following a first-principles review he directed.
> Context: the NeoWiki codebase, ADR 21 + (superseded) ADR 12, `docs/reference/validation-codes.md`, and PR #822.

Mechanically this is clean, in-scope for ADR 21, and well-tested (the always-persist behavior is correctly pinned on both paths). Two things I'd address before merge, plus one sequencing question.

## 1. The dry-run and the write now disagree (the main one)

The whole point of the `/validate` endpoints (ADR 21: get violations *before attempting a save*) is to predict what a write will report. Now that the write path also reports violations, the two pipelines are observably different on at least two cases:

**Missing schema**
- `POST /subject/{id}/validate` → `200` with a `schema-not-found` violation (a documented code in `validation-codes.md`)
- `PUT /subject/{id}` (this PR) → `200`, `violations: []`
- Worse on create: `POST /subject/validate` → **404** (`SchemaNotFoundException`), while `POST /subject` → **201 persisted, `violations: []`** (pinned by `testCreateWithMissingSchemaProducesEmptyViolations`). So the create dry-run is *stricter* than the create write — a client that validates-before-writing gets blocked on a case the write accepts.

**Unresolvable select option**
- validate path uses `resolveOrLeave()` → surfaces `invalid-option` (`200`)
- write path uses the throwing `resolve()` → **`400`**

The select divergence is already documented as a known, temporary foundation-tier cost in `SelectStatementResolver`'s docblock — credit for flagging it. But that docblock says the two resolvers collapse "once the enforcement tier runs the validator on write paths too." This PR *is* the moment the validator starts running on write paths, yet it keeps the throwing `resolve()` and doesn't collapse them. The precondition named there has now arrived.

A dry-run that doesn't faithfully predict the write undermines the reason the dry-run exists.

## 2. One pipeline instead of four copies

`validateProposedSubject()` is byte-identical in `CreateSubjectAction` and `ReplaceSubjectAction`, and it's a fourth near-copy of the "look up schema → validate" flow already in `ValidateSubjectQuery` and `ValidateSubjectUpdateQuery`. The copies aren't even faithful — they differ precisely on the null-schema branch (`[]` vs `404` vs `schema-not-found` violation), which **is** the drift in #1.

Extracting a single collaborator (e.g. `ProposedSubjectValidator::validate( Subject ): Violation[]`) that owns the missing-schema decision once, used by both actions and both queries, collapses the duplication *and* removes the drift by construction. Not coincidentally, that's also the single seam where enforcement and pre-existing-vs-new differentiation will need to live. Fixing #1 and #2 is one refactor.

## 3. Sequencing / response shape (worth a sanity check, not a blocker)

Stepping back from "step 1 of ADR 21": #822 already shipped dedicated `/validate` endpoints, so the marginal value here is narrow (saves a round-trip for optimistic writers; lays plumbing). The half stakeholders actually asked for — *rejecting* edits that introduce violations — is the hard part, and ADR 21 couples it to a hard requirement: distinguishing newly-introduced violations from pre-existing ones. That's fully deferred.

The concern: this PR freezes a flat `violations: []` shape on the edit endpoints before the design that should determine that shape exists. Under enforcement, a consumer will need to know *which* violations are new (i.e. which would have blocked) — likely per-violation classification or a split list. Pre-1.0 (`/neowiki/v0/*` is explicitly unstable) makes this "you'll break the shape later" rather than fatal, but it's worth sketching the differentiation/enforcement response shape now so this shape is chosen with it in mind, rather than forward from "easiest first."

## Recommendation

Preferred: unify on the shared collaborator (fixes #1 + #2), make a deliberate, documented call on missing-schema and unresolvable-select parity between validate and write, and sketch the enforcement response shape before locking `violations`.

Acceptable if consciously incremental: merge as-is *only* if the next PR is the unification + differentiation + enforcement, with a note here that the validate/write drift is known and temporary. The risk is shipping the easy slice, stalling on differentiation, and leaving a permanently-divergent dry-run plus a shape that has to break later.

Minor: the `\RuntimeException`→import and `} catch` reformatting are nice but unrelated to surfacing violations — only worth splitting out if you're strict about it.

JeroenDeDauw:
Review-PR: https://github.com/ProfessionalWiki/NeoWiki/pull/864

PR #856, opened 2026-05-27
Enforce validation under NeoWikiValidationEnforced

alistair3149:
Follows-up to https://github.com/ProfessionalWiki/NeoWiki/pull/855.

Blocked by #865 since that should land first.

Implements step 2 of [ADR 21 backend validation](../tree/master/docs/adr/021-add-backend-validation.md): when an admin opts in via `$wgNeoWikiValidationEnforced = true`, the two write endpoints reject edits that introduce *newly-introduced* constraint violations. Pre-existing violations remain tolerated, so editing an already-invalid Subject is still possible (per ADR 21's invalid-Subjects-must-stay-editable requirement). Default is `false`; enforcement-off behavior is unchanged from step 1.

## What changed

- New `NeoWikiValidationEnforced` boolean config (default `false`) in `extension.json`.
- New `Domain/Validation/ViolationDiff::newViolations(proposed, prior): array` helper. Identity: `(propertyName, code)`.
- New `Application/Subject/Exception/ValidationFailedException` carrying `Violation[]`.
- `ReplaceSubjectAction` validates the prior persisted Subject against the *current* Schema, validates the proposed Subject, diffs. Throws `ValidationFailedException` when enforcement is on AND new violations are non-empty.
- `CreateSubjectAction` throws when enforcement is on AND the proposed Subject has any violation (no prior on Create). The `presentSubjectAlreadyExists` branch still short-circuits first.
- Both REST handlers catch `ValidationFailedException` → HTTP 422 with `{status, message, violations}` body.
- Step 1's success-response contract (`violations: [...]` on 200/201) is preserved.

## Wire shape on enforcement block

```json
HTTP 422 Unprocessable Entity
{
  "status": "error",
  "message": "Validation failed",
  "violations": [
    { "propertyName": "Status", "code": "required", "args": [] }
  ]
}
```

`violations` carries the *full* proposed violation list, not just the newly-introduced subset — the new-vs-pre-existing split is server-internal.

## Out of scope (future rounds)

- Severity (error/warning) tiering. Wire format stays additive (a future `severity` field per violation can land without breaking step 2).
- Per-Schema or per-Property enforcement overrides (ADR 21 lists this as a possible future direction).
- TS-side validator upgrade.
- `DeleteSubject` and `SetMainSubject` — neither changes Subject data.

## Test plan

- [x] `make phpunit filter=ViolationDiff` — 10/10 green.
- [x] `make phpunit filter=Replace` — 26/26 green.
- [x] `make phpunit filter=Create` — 26/26 green.
- [x] `make phpunit filter=Application` — 276/276 green (1 expected skip).
- [x] `make cs` — phpcs (392 files) + phpstan (189 files), no errors.
- [x] Live smoke against the running wiki, raw response bodies captured:
  - `PUT /subject/s1demo1aaaaaaa1` enforcement on, no Status → `422 {"status":"error","message":"Validation failed","violations":[{"propertyName":"Status","code":"required","args":[]}]}`
  - `PUT /subject/s1demo1aaaaaaa1` enforcement on, Status set → `200 {"status":"updated","subjectId":"s1demo1aaaaaaa1","violations":[]}`
  - `POST /page/91/mainSubject` enforcement on, no Status → `422 {"status":"error","message":"Validation failed","violations":[{"propertyName":"Status","code":"required","args":[]}]}`
  - `POST /page/92/mainSubject` enforcement on, Status set → `201 {"status":"created","subjectId":"sEu8VkDNrkdFxWz","violations":[]}`
  - `PUT /subject/s1demo1aaaaaaa1` enforcement off, no Status → `200 {"status":"updated","subjectId":"s1demo1aaaaaaa1","violations":[{"propertyName":"Status","code":"required","args":[]}]}` (step 1 contract restored)

### DIFF
--- a/docs/reference/validation-codes.md
+++ b/docs/reference/validation-codes.md
@@ -10,8 +10,14 @@ Violations are returned by:
 - `POST /neowiki/v0/subject/{subjectId}/validate` — dry-run validation of a proposed update-shape
   body against an existing Subject's Schema.
 - `POST /neowiki/v0/subject` and `PUT /neowiki/v0/subject/{subjectId}` — the write endpoints include
-  the resulting `violations` array in their `201`/`200` success body. Validation does not block the
-  write (see [ADR 21](../adr/021-add-backend-validation.md)).
+  the resulting `violations` array in their `201`/`200` success body. By default
+  (`$wgNeoWikiEnforceValidation = false`) the write always persists. When an admin enables
+  enforcement (`$wgNeoWikiEnforceValidation = true`), a write that introduces *new* blocking
+  violations is rejected with `422 Unprocessable Entity` and an `{ status, message, violations }`
+  body; `violations` carries the full proposed list, not just the newly-introduced subset.
+  Pre-existing violations (already present on the stored Subject under the current Schema) and
+  non-blocking codes such as [`schema-not-found`](#schema-not-found) never block, so editing an
+  already-invalid Subject stays possible. See [ADR 21](../adr/021-add-backend-validation.md).
 
 The `/validate` endpoints return `200 OK` with a `{violations: [...]}` body whenever the request is
 well-formed; violations in the body do not change the HTTP status. `400` is reserved for malformed
--- a/extension.json
+++ b/extension.json
@@ -140,6 +140,10 @@
 			"description": "Whether development UIs should be enabled",
 			"value": false
 		},
+		"NeoWikiValidationEnforced": {
+			"description": "Whether write endpoints reject edits that introduce new constraint violations. When false, violations are surfaced in responses but writes always persist.",
+			"value": false
+		},
 		"NeoWikiQueryLimits": {
 			"description": "Per-tier per-query resource caps. Keys: 'default' (users without apihighlimits) and 'expensive' (users with apihighlimits). Each value: { timeoutSeconds, maxRows }.",
 			"value": {
--- a/extension.json
+++ b/extension.json
@@ -140,7 +140,7 @@
 			"description": "Whether development UIs should be enabled",
 			"value": false
 		},
-		"NeoWikiValidationEnforced": {
+		"NeoWikiEnforceValidation": {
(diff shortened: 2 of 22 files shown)


JeroenDeDauw:
<sub>Question surfaced during an AI-assisted review (Claude Code, `Opus 4.8`); posted by @JeroenDeDauw.</sub>

One question on the enforcement granularity before this merges:

`ViolationDiff` keys violations on `(propertyName, code)` only — it ignores `valuePartIndex` and the violation *count*. So an edit that adds *more* violations of an already-violated `(property, code)` isn't treated as new and passes enforcement. Concretely, on a `url` property that accepts multiple values and already holds one invalid URL (so `(Website, invalid-url)` pre-exists), changing it to two invalid URLs makes `newViolations()` return `[]` → the write is accepted even with enforcement on.

That's consistent with ADR 21's "tolerate pre-existing violations / keep invalid Subjects editable" stance, so it may well be intentional — but it's a touch weaker than the PR description's "reject edits that introduce newly-introduced constraint violations", and nothing currently pins it. Is the `(propertyName, code)` granularity the intended contract? If so, a one-line note on `ViolationDiff` plus a test locking it in would help; if not, folding `valuePartIndex`/count into the identity would tighten it.

JeroenDeDauw:
@alistair3149 what do you think about the above question? I didn't think it through yet and the answer (optimal behavior) seems non-obvious

alistair3149:
I agree with increasing the granularity. Including `valuePartIndex` or the violation count ensures we only block truly "new" violations while keeping invalid subjects editable per ADR 21. I'll adjust the identity logic accordingly.

JeroenDeDauw:
Did you test how that interacts with the frontend editing UIs?

alistair3149:
> Did you test how that interacts with the frontend editing UIs?

Yes. Currently the frontend would throw a 422 error when enforcement is enabled. Handling that in the frontend is a follow-up to this PR, which is only scoped to enforce validation in the backend.

image

PR #817, opened 2026-05-11
Add public Cypher query REST API + close #549

alistair3149:
Fixes https://github.com/ProfessionalWiki/NeoWiki/issues/549

## Summary

- Adds `POST /neowiki/v0/query/cypher` — public read-only Cypher endpoint with stable JSON envelope (`columns`, `rows`, `truncated`, `resultCount`, `durationMs`) and stable `errorType` strings keyed to documented HTTP statuses.
- Funnels all three Cypher surfaces (REST, `nw.query` Lua, `{{#cypher_raw}}` parser fn) through a shared `QueryService` so validation, transaction timeout, result-row cap, and exception translation live in one place.
- Closes #549 (no transaction timeout, no LIMIT enforcement, no rate limiting on user-supplied Cypher) by applying the safety caps in `QueryService` and rate-limiting at the REST surface only (parser/Lua skip `pingLimiter` because their output is parser-cached).

Per-surface user controls:
- `neowiki-query` user right (default `*`) gates whether anyone can call the endpoint at all. Sysadmin lockdown: `$wgGroupPermissions['*']['neowiki-query'] = false;`
- Core's `apihighlimits` (default `bot`/`sysop`) lifts limits to the `expensive` tier (300s / 50,000 rows vs default 30s / 5,000 rows).
- `$wgRateLimits['neowiki-query']` ships defaults: anon 10/min, user 60/min, bot 1000/min. Sysop/bureaucrat exempt via core `noratelimit`.
- `$wgNeoWikiQueryLimits` overrides per-query timeout and row cap per tier.

Architecture follows ADR 19 (no NeoWiki-specific query abstraction language; per-backend honesty). The future SPARQL endpoint will sit alongside as `/neowiki/v0/query/sparql` with its own service.

## Files

- New: `src/Application/Query/{QueryService,QueryRequest,QueryResult,QueryLimits,QueryResultNormalizer}.php` + 7 typed exception classes
- New: `src/EntryPoints/REST/QueryCypherApi.php`
- New: `docs/QueryApi.md` (discovery flow, contract, Neo4j-side deployment recommendations)
- Refactored: `CypherQueryRunner` (Lua) and `CypherRawParserFunction` (parser fn) into thin adapters over `QueryService`
- Modified: `extension.json` (right, RateLimits, config, route), `Neo4jQueryStore::runReadQuery()` (timeout)
- Tests: 80+ new test methods across unit, integration, and end-to-end-against-real-Neo4j layers

## Markup change

[...]

PR #861, opened 2026-05-29
Localize Cypher error messages on wikitext surfaces

### LINKED ISSUE #858: Cypher error messages are no longer localized in {{#cypher_raw}} and nw.query()
> Result of back and forth with @JeroenDeDauw while reviewing #817.
> Context: the post-#817 `Neo4jQueryService` / `CypherRawParserFunction` / `CypherQueryRunner` code paths, the removed i18n keys, and the diff vs the pre-#817 behavior.

## Goal

Restore localized error messages on the wikitext-facing Cypher surfaces (`{{#cypher_raw}}` and `nw.query()`). End users of a non-English wiki should not see hardcoded English when they trigger an error on these surfaces.

The REST endpoint (`POST /neowiki/v0/query/cypher`) is a separate concern — for an API surface, hardcoded English plus a stable `errorType` is arguably the right contract, so it's fine to leave that alone unless you decide otherwise.

## Background

#817 removed five i18n keys:
- `neowiki-cypher-raw-error-empty-query`
- `neowiki-cypher-raw-error-write-query`
- `neowiki-cypher-raw-error-query-failed`
- `neowiki-lua-query-error-empty-query`
- `neowiki-lua-query-error-write-query`

and replaced them with hardcoded English strings thrown from exceptions:
- `Neo4jQueryService.php:34` — `throw new EmptyQueryException( 'Query is empty.' )`
- `Neo4jQueryService.php:49` — `throw new WriteQueryRejectedException( 'Query is not read-only.' )`

`CypherRawParserFunction::handle()` and `CypherQueryRunner::run()` now surface these via `$e->getMessage()`, so a non-English wiki user sees `Query is empty.` (or the raw Neo4j-supplied string for syntax errors etc.) regardless of their interface language.

`docs/QueryApi.md` also claims `message` is translated, which is currently inaccurate for the REST surface too — see the "if we keep the REST behavior" note below.

## Considerations (recommendations, not prescriptions)

The shape of a fix depends on where you want localization to live. A few directions worth weighing:

- **Translate at the adapter boundary.** `QueryException::errorType()` already gives a stable key. The parser function and Lua adapter could map `errorType` → message key and call `wfMessage`. Keeps the service layer pure and English-message-free, makes the REST surface unaffected. Probably the smallest change.
- **Carry a message key on the exception.** Exceptions get a `messageKey()` alongside `errorType()` and a default English message. Adapters call `wfMessage($e->messageKey())`. More invasive but lets future Cypher surfaces (e.g. a Special page) reuse the same mapping without re-doing the switch.
- **Localize in the REST output too.** If you want the REST `message` field to match the docs' claim, the request's `Accept-Language` would have to drive translation. I'd argue against this for an API — it makes integration testing harder and `errorType` already gives clients what they need — but it's a defensible direction if there's a real consumer who'd benefit.

Whichever direction, the docs claim at `docs/QueryApi.md:111` should be made accurate (either localize the REST output, or remove "translated" from the docs).

Push back on any of these framings if you see a better goal. For example, if Neo4j-supplied error messages (syntax errors, etc.) shouldn't be localized at all, that's worth discussing — those carry diagnostic detail that a translator probably can't usefully restate.

## Done when

- A `{{#cypher_raw}}` empty-query / write-query / syntax-error / timeout error on a non-English wiki shows a localized message.
- Same for `nw.query()` errors in Lua.
- `docs/QueryApi.md` accurately describes whether REST `message` is translated.

alistair3149:
> Direction and the key architecture decision — a shared mapper in the EntryPoints layer rather than carrying message keys on the exception — by @alistair3149.
> Context: the NeoWiki codebase, issue #858, the post-#817 Cypher query code paths, and MediaWiki/Scribunto message-handling internals.

Fixes https://github.com/ProfessionalWiki/NeoWiki/issues/858

## Screenshots

00-overview

<details>
<summary>Per case</summary>

| Surface / input | Screenshot |
|---|---|
| `{{#cypher_raw:}}` (empty) | 01-empty |
| `{{#cypher_raw:CREATE (n) RETURN n}}` (write) | 02-write |
| `{{#cypher_raw:MATCH (n) RETURN}}` (syntax - Neo4j detail forwarded as `$1`) | 03-syntax |
| `nw.query( "" )` (Lua) | 04-lua |

</details>

#817 replaced the localized i18n error messages on the wikitext-facing Cypher surfaces (`{{#cypher_raw}}` and `nw.query()`) with hardcoded English thrown from exceptions, so non-English wikis saw English error text.

A shared `CypherErrorMessage` mapper in the Neo4j EntryPoints layer turns a `QueryException` into a message key plus parameters (not rendered text), so each surface renders it in its own content-language context (parser-cache safe). The four Neo4j-diagnostic error types embed the verbatim, untranslatable Neo4j detail as `$1`; the semantic types (empty, write-rejected, backend-unavailable) are fully localized. `CypherQueryRunner` no longer flattens `QueryException` into a bare `RuntimeException`, letting the typed exception reach the Lua surface for localization. The seven `cypher-raw`/`lua` error keys are unified into one `neowiki-cypher-error-*` family so both surfaces show identical wording.

The REST API keeps its English message plus stable `errorType` by design; only the inaccurate "message is translated" claim in `docs/reference/query-api.md` is corrected.

## Manual Browser Check

These surfaces render server-side, so no TypeScript build is needed. The wiki runs at http://localhost:8484.

1. **Empty query** — create a page containing `{{#cypher_raw: }}` and view it. The error box should read *"The Cypher query must not be empty."* (previously the hardcoded *"Query is empty."*).
2. **Write query** — `{{#cypher_raw: CREATE (n) RETURN n}}` → *"Only read-only Cypher queries are allowed."*
3. **Syntax error** (Neo4j must be running) — `{{#cypher_raw: MATCH (n) RETURN }}` → *"The Cypher query has a syntax error: &lt;verbatim Neo4j detail&gt;"* — a localized frame wrapping the untranslatable Neo4j text.
4. **Proves it goes through i18n, not a hardcoded string** — edit `MediaWiki:Neowiki-cypher-error-empty-query` to any custom text, then reload the page from step 1; the error should now show your custom text. Revert afterwards.
5. **Lua surface** — create a module with `return mw.neowiki.query( '' )` (or call it from a template) and invoke it; the Lua error should carry the same *"The Cypher query must not be empty."* message instead of the old raw string.

### DIFF
--- a/i18n/en.json
+++ b/i18n/en.json
@@ -115,6 +115,14 @@
 
 	"neowiki-cypher-raw-error-json-encode": "Failed to encode query result as JSON",
 
+	"neowiki-cypher-error-empty-query": "The Cypher query must not be empty.",
+	"neowiki-cypher-error-write-query": "Only read-only Cypher queries are allowed.",
+	"neowiki-cypher-error-syntax": "The Cypher query has a syntax error: $1",
+	"neowiki-cypher-error-parameter-missing": "The Cypher query is missing a required parameter: $1",
+	"neowiki-cypher-error-timeout": "The Cypher query exceeded the allowed time limit: $1",
+	"neowiki-cypher-error-backend-unavailable": "The graph database is currently unavailable.",
+	"neowiki-cypher-error-internal": "The Cypher query could not be completed: $1",
+
 	"neowiki-view-error-extra-positional": "Unexpected extra positional argument: $1",
 	"neowiki-view-error-conflicting-subject": "Subject specified both positionally and as a named subject argument.",
 	"neowiki-view-error-unknown-arg": "Unknown argument: $1",
--- a/docs/reference/query-api.md
+++ b/docs/reference/query-api.md
@@ -103,7 +103,10 @@ Temporal and spatial values (e.g. `datetime()`, `point()`) are not supported. Ca
 | `internalError` | 500 | Anything else. |
 
 `errorType` strings are stable across releases. Clients and LLMs should branch on `errorType`, not on `message`
-text — `message` is translated and may change.
+text. The REST `message` is always English and may change between releases; it is intended for logging and
+debugging, not display. User-facing localization applies only to the wikitext surfaces — the `{{#cypher_raw}}`
+parser function and `nw.query()` in Lua — which render translated messages derived from the same `errorType`
+classification.
 
 ## Example walkthrough
 
--- a/i18n/qqq.json
+++ b/i18n/qqq.json
@@ -14,6 +14,14 @@
 	"neowiki-schema-label": "Label used to identify the schema. $1 is the schema name or link.",
 	"neowiki-cypher-raw-error-json-encode": "Error message shown when JSON encoding of query results fails.",
 
+	"neowiki-cypher-error-empty-query": "Error shown on the <code>{{#cypher_raw}}</code> parser function and the <code>nw.query()</code> Lua function when the supplied Cypher query is empty or only whitespace.",
+	"neowiki-cypher-error-write-query": "Error shown on the <code>{{#cypher_raw}}</code> parser function and the <code>nw.query()</code> Lua function when the query contains a write operation. Cypher queries from wikitext are restricted to read-only.",
+	"neowiki-cypher-error-syntax": "Error shown when Neo4j reports a syntax error for a Cypher query run from wikitext. $1 is the verbatim Neo4j-supplied syntax detail and is not translatable.",
+	"neowiki-cypher-error-parameter-missing": "Error shown when a Cypher query run from wikitext references a parameter that was not supplied. $1 is the verbatim Neo4j-supplied detail and is not translatable.",
+	"neowiki-cypher-error-timeout": "Error shown when a Cypher query run from wikitext exceeds the allowed time limit. $1 is the verbatim Neo4j-supplied detail and is not translatable.",
+	"neowiki-cypher-error-backend-unavailable": "Error shown when the graph database backend cannot be reached for a Cypher query run from wikitext. The underlying connection detail is intentionally omitted.",
+	"neowiki-cypher-error-internal": "Generic fallback error shown for a Cypher query run from wikitext that fails for a reason not covered by the more specific cases. $1 is the verbatim underlying detail and is not translatable.",
+
 	"neowiki-view-error-extra-positional": "Error shown when {{#view}} is called with more than one positional argument. $1 is the offending extra argument value.",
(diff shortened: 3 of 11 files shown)


i18n/en.json
```diff
@@ -115,6 +115,14 @@
 
 	"neowiki-cypher-raw-error-json-encode": "Failed to encode query result as JSON",
 
+	"neowiki-cypher-error-empty-query": "The Cypher query must not be empty.",
+	"neowiki-cypher-error-write-query": "Only read-only Cypher queries are allowed.",
+	"neowiki-cypher-error-syntax": "The Cypher query has a syntax error: $1",
+	"neowiki-cypher-error-parameter-missing": "The Cypher query is missing a required parameter: $1",
+	"neowiki-cypher-error-timeout": "The Cypher query exceeded the allowed time limit: $1",
+	"neowiki-cypher-error-backend-unavailable": "The graph database is currently unavailable.",
+	"neowiki-cypher-error-internal": "The Cypher query could not be completed: $1",
```
malberts, line 124: The three parameterless messages in the new `neowiki-cypher-error-*` family end with a full stop:

- `neowiki-cypher-error-empty-query` → "The Cypher query must not be empty."
- `neowiki-cypher-error-write-query` → "Only read-only Cypher queries are allowed."
- `neowiki-cypher-error-backend-unavailable` → "The graph database is currently unavailable."

Each renders on both surfaces that `CypherErrorMessage` feeds:

- `{{#cypher_raw}}` shows the message standalone in an error box, where the trailing period reads correctly.
- `nw.query()` throws a `LuaError`, which Scribunto wraps as `Lua error: $1.`, so the trailing period becomes a double period:
  `Lua error: The Cypher query must not be empty..`
  This is visible in the PR's own `nw.query( "" )` (Lua) screenshot above.

This is pre-existing (the strings #817 introduced ended in a period too) and purely cosmetic, so not a blocker. It's just the first time I noticed it, and we don't necessarily have to decide here.

There is no period choice that is clean on both surfaces, since the only one that adds the extra period is Scribunto's `$1.` wrapper, which we don't control. Dropping the period from the three strings fixes the Lua surface but leaves the `{{#cypher_raw}}` box without a terminal period, and is slightly inconsistent with the sibling `neowiki-view-error-*` full-sentence messages, which keep their periods and only ever render standalone.

The four `$1` messages (syntax, parameter-missing, timeout, internal) are unaffected: their tail is the Neo4j-supplied detail, not an authored period.

PR #862, opened 2026-06-01
CI: run PHPUnit against MediaWiki master (1.47) on PHP 8.5

malberts:
Adds a MediaWiki `master` (1.47) / PHP 8.5 row to the PHPUnit matrix, so the extension is exercised against the development branch in addition to the current LTS (`REL1_43` / PHP 8.3). The `master` row is marked `experimental` and runs `continue-on-error`, so breakage originating in MediaWiki core does not block PRs while still surfacing the failure. This follows the `continue-on-error: ${{ matrix.experimental }}` convention already used across the other ProfessionalWiki extensions (e.g. Network, WikibaseFacetedSearch, Chameleon).

Adding the row required migrating the test runner, since MediaWiki 1.46 removed the `tests/phpunit/phpunit.php` entry point the workflow invoked.

## One runner for every version

The replacement is a `phpunit` composer script in the extension, called by both CI and the Makefile so the two cannot drift. It runs unchanged on `REL1_43` and `master`: `composer phpunit:config` generates the config MediaWiki 1.46 and later require, and is tolerated as a no-op where that script does not exist. No version gate is needed, and 1.43's own entry point already printed a deprecation notice pointing at `composer phpunit`.

The script keeps `-c extensions/NeoWiki/phpunit.xml.dist` and passes core's bootstrap explicitly, which the removed entry point used to inject on its own. The alternative, selecting tests by path and inheriting core's config, would have silently adopted its `forceCoversAnnotation`, `failOnWarning`, `failOnRisky` and strictness settings, and would have dropped the `<env>` block that points the Neo4j and QLever tests at the right endpoints.

## Cloning instead of unpacking

MediaWiki is now cloned rather than fetched as an archive. `phpunit.xml.template` is `export-ignore`d in MediaWiki's `.gitattributes`, so it is absent from the tarball while the generator that reads it is present, and on `master` the bootstrap refuses to run without the generated config.

The MediaWiki cache key is bumped alongside it. That cache is keyed only by branch and PHP version, so without a new key the pre-existing tarball checkout would be restored and the install step skipped, leaving the change with no effect.

The PHPStan job stays pinned to a single MediaWiki version, matching the same org-wide convention: PHPUnit fans out across versions, static analysis does not. `phpcs` is unchanged.

### DIFF
--- a/.github/workflows/ci-php.yml
+++ b/.github/workflows/ci-php.yml
@@ -21,12 +21,18 @@ jobs:
   test:
     name: "PHPUnit: MW ${{ matrix.mw }}, PHP ${{ matrix.php }}"
 
+    continue-on-error: ${{ matrix.experimental }}
+
     strategy:
       fail-fast: false
       matrix:
         include:
           - mw: 'REL1_43'
             php: '8.3'
+            experimental: false
+          - mw: 'master'
+            php: '8.5'
+            experimental: true
 
     runs-on: ubuntu-latest
 
--- a/.github/workflows/ci-php.yml
+++ b/.github/workflows/ci-php.yml
@@ -152,7 +152,8 @@ jobs:
           echo "QLever did not become ready in time; dumping logs:"; docker logs test_qlever; exit 1
 
       - name: Run PHPUnit
-        run: php tests/phpunit/phpunit.php -c extensions/NeoWiki/phpunit.xml.dist
+        run: composer phpunit
+        working-directory: mediawiki/extensions/NeoWiki
 
   phpcs:
     name: "PHPCS with PHP 8.3"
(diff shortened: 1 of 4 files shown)


malberts:
Superseded by #1213, which carries this work plus the compatibility fixes it needs. The two were not independently mergeable: the `master` row added here fails without those fixes, so merging this first would have put a failing row on master until the second PR landed.

Commit `Add PHPUnit config generation` is superseded rather than dropped: generating the config is unnecessary once MediaWiki is cloned instead of unpacked from an archive, since the template it reads is `export-ignore`d. The authorship is carried on the CI commit in #1213.

> Result of back and forth with @malberts.
> <sub>Written by Claude Code, `Opus 5`</sub>

PR #807, opened 2026-05-04
Refactor frontend validation: declarative constraints + interpreter

alistair3149:
> Result of extensive back-and-forth with @alistair3149.
> Context: the NeoWiki codebase, `docs/planning/Validation.md`, `docs/planning/Validation-Severity.md`, and PR #631.
> <sub>Written by Claude Code, `Opus 4.7 (1M context)`</sub>

Follows-up to #631.

## Summary

- Splits per-Property-Type `validate()` into declarative constraint metadata + a generic interpreter, with a small per-type `validateValue` hook for intrinsic format checks (URL syntax, ISO-8601 parse, SubjectId validity, DateTime min/max).
- `SubjectValidator.validate` returns a structured `SubjectValidationError[]` instead of `boolean`; adds i18n message `neowiki-field-label-required`.
- Forward-compat severity: optional `severity` field on `Constraint` and `ValueValidationError`. No Property Type emits severity yet; activation is a one-line change per type when the severity-levels feature is decided.

## Details

- New `Constraint` discriminated union with six kinds: `required`, `minLength`, `maxLength`, `uniqueItems`, `cardinality` (with `maxItems: number`), `enum` (with `allowedValues: string[]`). Each variant carries optional `severity?: 'error' | 'warning'`.
- New `interpretConstraints(constraints, value)` pure function in `domain/ConstraintInterpreter.ts`. Switches on `constraint.kind`; type-mismatched constraints (e.g. `minLength` on a `NumberValue`) are skipped silently.
- New `isValueEmpty(value)` helper in `domain/Value.ts` centralizing the per-Value-Type "empty" notion.
- `BasePropertyType.validate` becomes concrete and delegates `[...validateValue(value, property), ...interpretConstraints(getConstraints(property), value)]`. The `validateValue`-first ordering preserves the existing per-type test ordering. `getConstraints` is abstract; `validateValue` defaults to `[]`.
- All six Property Types migrated. Text and Select are pure declarative (`getConstraints` only). Number, Url, Relation, DateTime each contribute a small `validateValue` hook (3-8 lines) for the type-intrinsic format check.
- `Constraint`, `ConstraintInterpreter`, `Severity` exported from `public-api.ts` so external TS extensions subclassing `BasePropertyType` can satisfy the abstract contract.
- Backfilled regression tests for `RelationType.validate` (which had none) before migrating it.
[...]

PR #879, opened 2026-06-08
Surface backend 422 validation violations inline in the Subject editor

alistair3149:
Follows-up to https://github.com/ProfessionalWiki/NeoWiki/pull/856.

## Summary

When `$wgNeoWikiEnforceValidation` is on and a save is blocked, the backend returns a 422 with a structured `violations[]` array. Today the editor discards it and shows the opaque toast _"Request failed with status code 422"_. After this PR, the editor parses the body, marks the offending fields inline, keeps the dialog open, and shows a toast that names the subject and points at the highlighted fields. Editing a flagged field clears its red border immediately. The Subject create dialog gets the same treatment. With enforcement off, behaviour is unchanged — writes always persist and violations come back as advisory data in the 200 response.

## What changed

**Persistence.** `ProductionHttpClient` widens `validateStatus` to accept 422 alongside 2xx so the response reaches `RestSubjectRepository` instead of axios throwing first. Other non-2xx still reject, preserving `CsrfSendingHttpClient`'s 403 refresh-and-retry. The three write methods (`createMainSubject`, `createChildSubject`, `updateSubject`) parse the 422 body via a structural guard and throw a typed `ValidationFailedError`; malformed bodies log to `console.error` and fall through to the existing generic-error path.

**Domain.** New `SubjectViolation` interface mirrors the backend wire shape one-to-one. New `ValidationFailedError extends Error` carries `readonly violations`. Independent `isValueEmpty(value: Value | undefined)` helper centralises the per-Value-Type _empty_ notion (one of the wins from #807 worth folding in early).

**Editor inputs.** `ValueInputContract` gains an optional `serverViolations` prop and a `clear-server-violation` emit. `useStringValueInput` merges server-sourced violations into the Codex error slot keyed by `valuePartIndex` (live errors take precedence on the same input), and surfaces field-level (`valuePartIndex: null`) server violations through the field summary regardless of multi/single so they can't silently vanish. Every per-PropertyType input — Text, Url, Number, Boolean, Date, DateTime, Select, Relation — passes the new prop/emit through. The composable also stops filtering invalid values out of `update:modelValue`: the backend is now the authoritative validator, so a bad URL the user types is sent on save and rejected with a real per-index violation rather than silently dropped.

**Dialogs.** Both `SubjectEditorDialog::handleSave` and `SubjectCreatorDialog::handleSave` catch `ValidationFailedError`, set reactive `serverViolations` state, fire the new `neowiki-subject-editor-validation-failed` toast, and leave the dialog open. Violations whose `propertyName` doesn't match a rendered statement (or that have `propertyName: null`, e.g. `schema-not-found`) render in a form-level `CdxMessage` banner above the editor. Editing a field that had a server violation emits up to drop the matching entry so the red border clears before the next save.

**Demo data.** New `Schema:Validation Demo`, `Validation Clean` subject, and `Validation Demo` hub page — a permanent live testbed for the 422-surfacing recipes. Useful for QA, future feature work, and onboarding.

## Out of scope

- **UX polish.** This PR wires up the surfacing; refinements to look-and-feel are deliberately deferred to a separate follow-up to keep the scope tight.
- **Approach B from the spec** — unifying `ValueValidationError` (live per-input errors) with `SubjectViolation`. They share the Codex slot but stay as distinct types here. Worth revisiting once enforcement has been in use for a while.
- **Severity field on `ValueValidationError`.** Backend doesn't emit severity; nothing to surface yet.
- **Aggregate `SubjectValidator` frontend class.** Registered as a service but never called — the spec deliberately leaves it alone. Delete vs. repurpose decided separately.
- **Other #807 wins** — RelationType regression tests and the DateTime empty-StringValue fix — are independent small follow-ups.

## Known surface gaps

- **`SelectInput` and `RelationInput` multi-mode** collapse `valuePartIndex`-keyed server violations into a single field-level message rather than per-chip decoration. Codex's multiselect and lookup widgets don't expose a per-chip error slot, and chip positions aren't stable. Showing only the first relevant violation is the pragmatic landing point.

## Manual Browser Check

Set `$wgNeoWikiEnforceValidation = true` in `LocalSettings.php`. Then, on the imported demo:

1. Open the `Validation Clean` subject page → action=subjects → **Edit** the main subject. Backspace the **Title** value and click **Save changes**. Confirm: dialog stays open, Title gets a red border with _"Please provide a value."_, toast reads _"Validation failed for 'Validation Clean'. Please fix the highlighted fields."_ Network panel shows the PUT returning 422.
2. With the red border still showing, type any character into Title. Confirm the red border clears immediately, before re-clicking Save.
3. Click **Save changes** again. Confirm: success toast, dialog closes.
4. Reopen the editor. Change the second **Tags** entry from `https://example.com/b` to `ftp://example.com/b` and click **Save changes**. Confirm: only the second sub-input gets the red border (the first stays clean), the 422 response body carries `valuePartIndex: 1`, the bad URL stays in the input so the user can fix it.
5. Click **Add subject** under _Other Subjects on this Page_, pick **Validation Demo**, leave Title empty, click **Create subject**. Confirm: 422, dialog stays open, Title decorated, validation-failed toast.
6. Set `$wgNeoWikiEnforceValidation = false`. Repeat step 1's edit. Confirm: PUT returns 200, dialog closes, success toast — proves the step-1 contract from #855 is preserved.

The console should be clean throughout, except for the browser's own _"Failed to load resource: ... 422"_ line which fires for any 422 response.

The `Validation Demo` hub page documents these recipes in-wiki.

## Test plan

- [x] `make tsci` green (build + full vitest suite at 953/953 + lint).
- [x] Plan-required narrow filters green: Value 227, RestSubjectRepository 20, SubjectEditorDialog 19, SubjectCreatorDialog 53.
- [x] `make phpunit` green — backend untouched; the new demo subject round-trips through `SubjectContentDataSerializerTest`.
- [ ] Reviewer: walk the Manual Browser Check above.
- [ ] Reviewer: confirm no console errors during the Create-side 422 step.


alistair3149:
https://github.com/user-attachments/assets/f6596bc2-bc05-48de-920d-7582d7e512ef

Issue #831, opened 2026-05-16
Add type-aware persistence for statement values in Neo4j


malberts:
> Written by Claude Code, `Opus 4.7`. Context: the NeoWiki persistence layer
> (`SubjectUpdater`, `Neo4jValueBuilderRegistry`, `Neo4jQueryStore`), #684, and #830.

## Goal

Add a general, type-aware persistence mechanism so **any property type — core or
extension-defined — can declare, via its value builder, how its value is represented
in Neo4j** for statements (native type / constructor expression), instead of every
value collapsing to an untyped scalar. Extension-provided property types are a
first-class requirement, not an afterthought (cf. the extensibility intent of #684
and the RedHerb example extension).

## Background

Subject statement values are currently persisted to Neo4j as untyped scalars, with no
mechanism for a property type to control its native Neo4j representation. datetime is
the clearest current symptom, but this is a **general** limitation that also blocks
future and extension-defined types (temporal, spatial/point, duration, or any custom
typed value an extension wants to store and query natively).

### Where it happens

- `SubjectUpdater::statementsToNodeProperties()` runs values through
  `Neo4jValueBuilderRegistry::buildNeo4jValue()` and writes them via
  `MERGE (n {id: $id}) SET n = $props` — every value is a plain Cypher **parameter**.
- `Neo4jValueBuilderRegistry` builders can only return scalars (all core builders,
  including `dateTime`, are `$value->toScalars()`). There is no way for a builder to
  emit a Cypher **constructor expression** (`datetime(...)`, `point(...)`, …) or
  otherwise influence the stored type.
- Contrast the page side (#684): `Neo4jQueryStore::extractTypedValues()` does have a
  typed path — but it is **hardcoded to datetime** and lives only in page persistence.

Net effect: no property type (core or extension-defined) can store Subject values as
native Neo4j types, so Cypher type-aware operations (datetime comparisons, spatial
queries, etc.) don't work against Subject statements.

## Design question — shared vs. separate mechanism (depends on #830)

The page side has a hardcoded, datetime-only typed path in `Neo4jQueryStore`; the
Subject side has none and uses the extensible `Neo4jValueBuilderRegistry`. Options:

1. **Generalise the value-builder mechanism** so builders can emit constructor
   expressions / typed representations, and reuse it for both page and subject
   persistence (single extensible path; supersedes the hardcoded page path).
2. **Add a separate Subject-side typed path** mirroring `extractTypedValues()` (two
[...]

Issue #830, opened 2026-05-16
RfC: Page-level computed properties


malberts:
> RfC distilled by Claude from an internal team design discussion and the 2026-05 dev syncs.
> Context: the NeoWiki codebase, GraphModel/Views docs, and the #782 / page-property plugin work.
> <sub>Written by Claude Code, `Opus 4.7`</sub>

## Problem

Some consumers need **page-level computed/generic properties** — e.g. a document-control workflow
that records "approved", "page owner", "last reviewed" — and wants to (a) view them alongside
Subject data and (b) update them outside of page edits.

The current mechanism (computed properties on the Neo4j **Page node** via the `PagePropertyProvider`
plugin) has run into several limitations that, taken together, warrant a decision on the overall
model rather than incremental patches:

1. **Not viewable via Views.** The `{{#view}}` + Layout system renders a single Subject and has no
   primitive for mixing Page-node properties into that rendering. Current workaround is custom Cypher
   via Lua/parser function — acknowledged as unsatisfactory.
2. **Updates outside page edits.** Tracked in #782 (expose a rebuild service so an extension can
   refresh page properties on demand, e.g. from an approval hook).
3. **Graph-as-generic-storage concern.** If meaningful userland data lives only on the page node,
   the graph database stops being purely disposable/rebuildable secondary persistence.
   (Counter-point: with an exposed ad-hoc rebuild (#782), the extension only *triggers* a rebuild
   from registered `getProperties()` — it does not write one-off data directly to the graph, so the
   Page node stays fully derivable from registered providers + revision state.)
4. **Influences subject-side typed-value implementation.** Subject-side typed values (e.g. a datetime
   that calls Neo4j's native datetime constructor in Cypher, as the page-property side already does
   for datetime-typed values such as `creationTime`/`modificationTime`) are *not strictly blocked* by
   this decision. But the outcome determines whether they reuse the same value-conversion/wrapper
   classes as the page-property mechanism or warrant separate ones — worth deciding before investing
   in either. Tracked in #831; the page-property equivalent was #684 (closed).

NeoWiki's userland data is meant to live in **Subjects**; page-node properties were not designed as
a generic storage mechanism. This RfC decides the model.

### Option A — Page as a Subject

Promote what is stored on the Page node to a proper **Page Subject** with a single, generic `Page`
schema (one schema, not per-page-type — per-type concerns stay in the **main subject**). The
schema/values may be read-only via UI/API. A Page then behaves like any other Subject.

[...]

Issue #782, opened 2026-04-27
Expose SubjectPageRebuilder as a public service for on-demand graph refresh


malberts:
## Problem

Page Property updates currently only run on `RevisionFromEditComplete` and `RevisionUndeleted`. Extensions that contribute a `PagePropertyProvider` may need to refresh a page's properties when something happens *outside* of an edit — for example, an approval extension that wants to reflect a state change ("draft" → "approved") on the Page node without requiring the page to be re-saved.

Today, the only way to trigger this is to reach into NeoWiki internals:

```php
$handler = NeoWikiExtension::getInstance()->getStoreContentUC();
$rebuilder = new SubjectPageRebuilder( $handler, $wikiPageFactory );
$rebuilder->rebuild( $title );
```

Both `OnRevisionCreatedHandler` and `SubjectPageRebuilder` live in non-public namespaces and are not registered with `MediaWikiServices`. There is no `ServiceWiring.php` for the extension.

## Proposal

Register `SubjectPageRebuilder` (or a smaller, purpose-built `PagePropertyRefresher`) as a `MediaWikiServices` service so external extensions can do something like:

```php
MediaWikiServices::getInstance()
    ->getService( 'NeoWiki.SubjectPageRebuilder' )
    ->rebuild( $title );
```

This is a small change (add `ServiceWiringFiles` to `extension.json`, write a `ServiceWiring.php`) and gives extensions a stable extension point that doesn't depend on `NeoWikiExtension`'s singleton.

## Notes

- Existing callers: `maintenance/RebuildGraphDatabases.php` (would migrate to the service for consistency).
- `SubjectPageRebuilder::rebuild()` already silently returns `false` for pages without a revision, without a revision user, or without a `SubjectContent` slot — that behaviour should remain.
- A REST endpoint is not requested here. Extensions can wrap the service themselves if they need HTTP access.

> Result of an investigation requested by @malberts.
> Context: the NeoWiki codebase, with focus on the \`PagePropertyProvider\` plugin system and the page-save → graph-store pipeline.
> <sub>Written by Claude Code, \`Opus 4.7\`</sub>

JeroenDeDauw:
I do not like the MediaWiki service container in this proposal. See comments on https://github.com/ProfessionalWiki/NeoWiki/issues/789

Quoting the driving use case for this task:

[...]

PR #882, opened 2026-06-09
SubjectSources: correct load seam, rebuild scope, ADR 8 cite, #831 framing

alistair3149:
> Result of a Subject Sources doc review session with @alistair3149 — multi-agent review with adversarial
> revalidation of each finding against the codebase, then scoping this PR to the verified corrections.
> Context: the NeoWiki codebase, the linked ADRs and planning docs, and issues #782/#789/#830/#831.

Corrects four statements in the Subject Sources exploration doc, verified against the code:

* **Rendering**: the server-side by-id load path is `SubjectLookup` (behind the subject fetch endpoint and `SubjectResolver`), not `SubjectContentRepository`, which is page/revision-keyed and slot-only. Both current `SubjectLookup` implementations resolve ids through the graph index, so Source dispatch sits in front of that port. Also names the frontend schema-load path (raw `Schema:` page source via core REST, bypassing the NeoWiki schema endpoint) as part of the same seam, since source-resolved schemas have no `Schema:` page.
* **Refresh without an edit**: "re-injection on rebuild exists" only holds for pages whose latest revision carries the subject slot; rebuild does not visit other pages.
* **Schemas from Sources**: drop the ADR 8 citation on the "one schema type" claim — ADR 8 covers one-schema-per-Subject, not schema kinds.
* **Related**: replace "#831 — related but separate" with honest scoping: #831's shared-vs-separate design question is gated on #830 (which this doc re-frames); it stays independent for the first slice but becomes a prerequisite for type-aware querying of Subjects materialised from Sources.

### DIFF
--- a/docs/planning/SubjectSources.md
+++ b/docs/planning/SubjectSources.md
@@ -102,8 +102,7 @@ Example sources:
 
 Schemas are resolved through Sources too and can be read-only: the page-metadata schema is code-defined (read-only by
 construction); a remote schema is remote-owned (read-only); a local schema is an ordinary writeable schema. There is
-one schema type throughout — read-only-ness comes from the Source, never from a schema "kind"
-([ADR 8](../adr/008-one-schema-per-subject.md)).
+one schema type throughout — read-only-ness comes from the Source, never from a schema "kind".
 
 ### Schema identity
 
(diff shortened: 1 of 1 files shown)


docs/planning/SubjectSources.md
```diff
@@ -102,8 +102,7 @@ Example sources:
 
 Schemas are resolved through Sources too and can be read-only: the page-metadata schema is code-defined (read-only by
 construction); a remote schema is remote-owned (read-only); a local schema is an ordinary writeable schema. There is
-one schema type throughout — read-only-ness comes from the Source, never from a schema "kind"
-([ADR 8](../adr/008-one-schema-per-subject.md)).
+one schema type throughout — read-only-ness comes from the Source, never from a schema "kind".
```
JeroenDeDauw, line 105: Ah I think this should be dropped as its a detail/correction from my session, not something sensible without that context

PR #884, opened 2026-06-10
Port TS-only validation rules into PHP

alistair3149:
> Result of an extensive design collaboration with @alistair3149, who chose the server-assisted validation direction and steered the scope; this is the first, ships-regardless increment.
> Context: the NeoWiki codebase, the validation-sync design doc, and web research on client/server validation UX.

Aligns three points where the frontend and backend validators had drifted, moving the authoritative rule into PHP (or aligning a default) so a schema validates identically on both sides.

- **minLength/maxLength on the text Property Type.** `TextProperty` gains the two nullable fields (parsed and serialized like `NumberProperty`'s bounds); `TextType` emits per-part `min-length`/`max-length` violations mirroring the TypeScript validator. Empty parts are skipped on both sides, matching the frontend's empties-filtered behaviour.
- **uniqueItems default.** The frontend deserialization default is aligned to `false`, matching the backend and `docs/reference/schema-format.md`. Null length bounds emitted by the backend serializer are coerced to `undefined` on deserialization so they are never mistaken for a real bound (otherwise a `null` `maxLength` would spuriously flag every value in the editor).
- **Property name trimming.** The backend `PropertyName` now trims and rejects whitespace-only names, matching the frontend.

Existing schema-serialization fixtures are updated for the two new text keys.

This is step 1 of a staged validation-sync effort; it is independently valuable as a set of drift bug-fixes and does not depend on the later steps.

## Manual Browser Check

The Validation Demo schema is extended (in this branch) to exercise the new rules. After `make load-test-data` and `make ts-build`:

1. Open the **Validation Demo** subject and edit it.
2. In a text field with a min/max length, type a too-short then a too-long value — confirm `Minimum length is N characters` / `Maximum length is N characters` appear inline as you type, and clear when the value is valid.
3. Confirm a text/url field *without* `uniqueItems` no longer flags duplicate entries live (default is now off, matching the backend).
4. Confirm no spurious "Maximum length is null characters" error ever appears on a property that has no max length.

### DIFF
--- a/DemoData/Page/Validation_Demo.wikitext
+++ b/DemoData/Page/Validation_Demo.wikitext
@@ -14,18 +14,34 @@ With this flag on, writes that introduce new violations are rejected with HTTP 4
 
 * [[Validation Clean]] — a Subject with every property filled in correctly. Use it as the starting point for the recipes below.
 
-The [[Schema:Validation Demo|Validation Demo]] schema has four properties spanning the validation surface:
+The [[Schema:Validation Demo|Validation Demo]] schema has one or more properties for every validatable Property Type, so you can trigger each rule by typing a bad value in the editor:
 
 {| class="wikitable"
-! Property !! Type !! Notes
+! Property !! Type !! What to try
 |-
-| Title || text, required || A simple required text field.
+| Title || text, required || Clear it and save to see a "required" error.
+|-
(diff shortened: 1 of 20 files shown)


alistair3149:
https://github.com/user-attachments/assets/f3b2db6b-89c0-4ad4-9b30-42169545d7f7

PR #885, opened 2026-06-10
Validate Schema and Layout content at save time, covering all write paths

alistair3149:
> Result of a design collaboration with @alistair3149; second increment of a staged validation-sync effort.
> Context: the NeoWiki codebase and the validation-sync design doc.

Move the JSON schema validation and the title-name check from the UI-only `EditFilter` hook into `SchemaContentHandler::validateSave()` and `LayoutContentHandler::validateSave()`. Every `PageUpdater` write — the Vue editors' REST `/v1/page` PUT, the action API, EditPage, and the NeoWiki demo importer — is now gated by one authority, closing the hole where the Vue editors' REST save path bypassed validation entirely. The redundant `EditFilter` content check and the now-unused `JsonSchemaErrorFormatter` are removed. Incidentally fixes a pre-existing latent bug: the old hook referenced a `neowiki-schema-invalid` message that never existed.

Core MediaWiki XML import (`ImportableOldRevisionImporter` → `RevisionStore::insertRevisionOn`) bypasses `validateSave`; this remains a known gap, documented by a characterization test. A real import-time gate is a separate, deferred follow-up.

Independent of #884 (step 1) — different files, mergeable in either order.

## Manual Browser Check

After `make load-test-data` and `make ts-build`:

1. Create or edit a **Schema** via the UI (`Special:CreateSchema` / the schema editor) with valid content and save — confirm it still saves normally.
2. Same for a **Layout**.
3. (Optional, raw editor) Open a `Schema:` page in the plain MediaWiki editor, replace the JSON with something structurally invalid (e.g. remove `propertyDefinitions`), and save — confirm the save is now **rejected** with an inline error, on the REST/API path as well as EditPage.
4. Confirm a normal new Layout created through the Vue creator (schema + view type selected) saves fine (the empty `{"schema":"","type":""}` default is only reachable via a blank raw-EditPage save, which is now correctly rejected).


JeroenDeDauw:
Merge conflict

malberts:
Two corrections to the PR description for the record; the diff itself is sound and I'm fine with the change.

**1. The unguarded path was the demo importer, not the REST PUT.** The description says this closes "the hole where the Vue editors' REST `/v1/page` PUT save path bypassed validation entirely." That path was already gated: REST PUT routes through `ApiEditPage` → `EditPage::internalAttemptSave`, which fires the old `onEditFilter` and aborts invalid content using the same `SchemaContentValidator`/`SchemaName`. The genuinely unvalidated path was the **demo importer** (`PageContentSaver::saveContent()` → `newPageUpdater()->saveRevision()`), which skips `EditPage`/`EditFilter` and so only got the parent `JsonContentHandler`'s JSON well-formedness check. Consolidating into `validateSave()` (which `PageUpdater` runs on every write path) is the correct fix; the rationale just names the wrong path.

**2. The characterization test no longer exists.** The description says the remaining core XML import gap is "documented by a characterization test." `tests/phpunit/EntryPoints/ImportValidationGapTest.php` was added in `5bddd1d6` and removed in the head commit `e0a2977d` ("Slim validateSave integration tests..."). On the current branch the gap is documented only in a commit message and design doc, not in code. Either restore a slim version or drop the sentence.

Minor: the manual-check steps reference `make load-test-data`, which is not a target (it is `make import-demo-data`).

PR #886, opened 2026-06-11
Add server-driven subject validation alongside the editor

alistair3149:
> Result of a design collaboration with @alistair3149; an increment of a staged validation-sync effort (plus its create/edit display-model refinement and a drive-by 404 fix), and the prototype for the demo-first ADR gate.
> Context: the NeoWiki codebase, the validation-sync design doc, and web research on client/server validation UX.

Introduce a debounced, sequence-guarded **dry-run validation path** that runs **alongside** the existing live validators in the Subject editor and creator. This is the *alongside-mode prototype*: server-driven validation runs and surfaces inline, but nothing is removed — so the team can evaluate it live before deciding on any larger change (deleting the duplicated TS validators is a separate, gated step, not in this PR).

### Core

- `RestSubjectRepository` gains `validateSubject` / `validateSubjectUpdate`, calling the existing read-only dry-run REST endpoints and reading violations from the 200 body (shared parser extracted to `violationParsing.ts`). A failed dry-run degrades to no violations and never blocks editing or saving.
- A `useSubjectValidation` composable owns the violations ref: `revalidate()` debounced (configurable via `$wgNeoWikiValidationDebounceMs`, default 300, `0` = blur-only), `flush()` on blur and before save, a request-sequence guard so stale responses never overwrite newer ones, and timer cleanup on scope dispose.
- Both subject dialogs feed the composable's violations into the **same ref** the inline renderer and the existing 422-on-save handling already use; the live per-field validators are unchanged.

### Display model: eager for existing subjects, lazy for new ones

Validation visibility differs by context so we never flag a field the user has not reached yet:

- **Editor (existing subject):** validates **on open** and on every edit; surfaces everything, including `required` — an empty required field on an existing subject is a real gap.
- **Creator (new subject):** validates on edit but **withholds "not filled in yet" violations** (`required`, `label-required`) until save — every field starts empty and the user is on their way to filling them in (`withoutMissingValueViolations`). `SelectInput` likewise no longer flags `required` client-side, deferring to the server like every other input.
- The dry-run is **skipped while a new schema is still an unsaved draft** (it does not exist server-side yet).

Open follow-up: under the default `$wgNeoWikiEnforceValidation = false`, the creator surfaces `required` nowhere (withheld mid-edit, and no 422 at save). Giving non-blocking save-time feedback there is the "warning vs error" model, tracked separately.

### Drive-by (pre-existing on master, unrelated to subject validation)

The schema creator's name-availability check fetched `/v1/page/Schema:{name}`, which **404s** for a not-yet-created name on every debounced keystroke. Switched to the existing `schema-names` search (a 200 response), with MediaWiki-style name normalization so the duplicate check still catches `person` / `Foo_Bar` collisions the way the page-fetch did.

Independent of #884 and #885 (the dry-run endpoints it consumes already exist on master).

## Manual Browser Check

> The point of this PR is to *watch server-driven validation work live*. Automated coverage is complete (1019 frontend tests, lint, build, PHPStan green; the dry-run endpoints have their own backend tests), but the live behavior should be observed before merging.

After `make load-test-data` and `make ts-build`:

**Editing an existing subject** (e.g. **Validation Clean** → its **Validation Demo** subject):

1. **On open**, a pre-existing violation (e.g. an empty required field) surfaces immediately — no edit needed.
2. Server violations appear inline ~300ms after you stop typing an invalid value, *in addition to* the instant client-side validation; **DevTools → Network** shows `POST .../neowiki/v0/subject/.../validate` after the pause and on blur, but **not** when merely tabbing between fields without editing.
3. Editing a flagged field clears its error immediately; the next dry-run re-asserts it if still invalid.
4. Saving still works; an enforced 422 (`$wgNeoWikiEnforceValidation = true`) surfaces inline as before.
5. With `$wgNeoWikiValidationDebounceMs = 0`, validation fires only on blur / pre-save.

**Creating a subject** (Add subject):

6. Required fields and the label stay **clean** while you fill them in — no premature "please provide a value".
7. Creating a **new schema**: typing the name produces **no `Schema:` 404s** (only `schema-names` 200s in the console / Network).

### DIFF
--- a/extension.json
+++ b/extension.json
@@ -147,6 +147,10 @@
 			"description": "Whether write endpoints reject edits that introduce new constraint violations. When false, violations are surfaced in responses but writes always persist.",
 			"value": false
 		},
+		"NeoWikiValidationDebounceMs": {
+			"description": "Debounce delay in milliseconds before the subject editor sends a server-side dry-run validation request. 0 disables the timer, so validation only runs on blur or pre-save.",
+			"value": 300
+		},
 		"NeoWikiQueryLimits": {
 			"description": "Per-tier per-query resource caps. Keys: 'default' (users without apihighlimits) and 'expensive' (users with apihighlimits). Each value: { timeoutSeconds, maxRows }.",
 			"value": {
--- a/extension.json
+++ b/extension.json
@@ -362,6 +362,7 @@
 				"neowiki-field-single-value-only",
 				"neowiki-field-type-mismatch",
 				"neowiki-field-schema-not-found",
+				"neowiki-field-label-required",
 				"neowiki-select-placeholder",
 				"neowiki-select-no-results",
 				"neowiki-select-unknown-option",
--- a/i18n/en.json
+++ b/i18n/en.json
@@ -72,6 +72,7 @@
 	"neowiki-field-single-value-only": "Only one value can be selected.",
 	"neowiki-field-type-mismatch": "This value was saved as type \"$1\" but the property now expects \"$2\".",
 	"neowiki-field-schema-not-found": "The schema \"$1\" is missing. Restore it before saving.",
+	"neowiki-field-label-required": "Please provide a label for the subject.",
 	"neowiki-select-placeholder": "Select an option",
 	"neowiki-select-no-results": "No matching options.",
 	"neowiki-select-unknown-option": "(unknown option)",
--- a/i18n/qqq.json
+++ b/i18n/qqq.json
@@ -155,6 +155,7 @@
 	"neowiki-field-schema-not-found": "Form-level error shown when the backend reports the Subject's Schema page is missing. $1 is the schema name.",
 	"neowiki-field-min-length": "Validation error shown when a text value is shorter than the property's minimum length. $1 is the minimum number of characters.",
 	"neowiki-field-max-length": "Validation error shown when a text value is longer than the property's maximum length. $1 is the maximum number of characters.",
+	"neowiki-field-label-required": "Form-level error shown when the backend reports the Subject is missing its required display label.",
 	"neowiki-select-placeholder": "Placeholder text shown in the select dropdown before any option is chosen.",
 	"neowiki-select-no-results": "Message shown in the multi-select dropdown when the typed text does not match any available option.",
 	"neowiki-field-invalid-datetime": "Validation error shown when the entered date and time value cannot be parsed. DateTime values must be strict ISO 8601 / xsd:dateTime strings with an explicit timezone offset or 'Z' (for example, '2025-06-15T12:00:00Z' or '2025-06-15T12:00:00+02:00'). Partial values such as year-only or date-only are rejected, and min/max bounds are inclusive.",
--- a/resources/ext.neowiki/src/NeoWikiExtension.ts
+++ b/resources/ext.neowiki/src/NeoWikiExtension.ts
@@ -255,6 +255,11 @@ export class NeoWikiExtension {
 		return Neo.getInstance();
 	}
 
+	public getValidationDebounceMs(): number {
+		const value = mw.config.get( 'wgNeoWikiValidationDebounceMs' );
+		return typeof value === 'number' ? value : 300;
+	}
+
 	public newSubjectValidator(): SubjectValidator {
 		return new SubjectValidator(
 			this.getPropertyTypeRegistry(),
--- a/resources/ext.neowiki/src/components/SchemaCreator/SchemaCreator.vue
+++ b/resources/ext.neowiki/src/components/SchemaCreator/SchemaCreator.vue
@@ -91,17 +91,15 @@ function onNameInput(): void {
 }
 
 async function checkDuplicateName( name: string, expectedSequence: number ): Promise<void> {
-	try {
-		await schemaStore.getOrFetchSchema( name );
+	const exists = await schemaStore.schemaNameExists( name );
 
-		if ( expectedSequence !== requestSequence ) {
-			return;
-		}
+	if ( expectedSequence !== requestSequence ) {
+		return;
+	}
 
+	if ( exists ) {
 		nameError.value = mw.msg( 'neowiki-schema-creator-name-taken' );
 		nameStatus.value = 'error';
-	} catch {
-		// Schema not found — name is available
 	}
 }
 
@@ -124,15 +122,13 @@ async function validate(): Promise<boolean> {
 		return false;
 	}
 
-	try {
-		await schemaStore.getOrFetchSchema( name );
+	if ( await schemaStore.schemaNameExists( name ) ) {
 		nameError.value = mw.msg( 'neowiki-schema-creator-name-taken' );
 		nameStatus.value = 'error';
 		return false;
-	} catch {
-		// Schema not found — name is available
-		return true;
 	}
+
+	return true;
 }
 
 function getSchema(): Schema | null {
--- a/resources/ext.neowiki/src/components/SubjectCreator/SubjectCreatorDialog.vue
+++ b/resources/ext.neowiki/src/components/SubjectCreator/SubjectCreatorDialog.vue
@@ -83,7 +83,8 @@
 				<CdxTextInput
 					v-model="subjectLabel"
 					:placeholder="$i18n( 'neowiki-subject-creator-label-placeholder' ).text()"
-					@input="markChanged"
+					@input="handleEditorChange"
+					@blur="handleEditorBlur"
 				/>
 				<template #label>
 					{{ $i18n( 'neowiki-subject-creator-label-field' ).text() }}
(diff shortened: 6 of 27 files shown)


alistair3149:
@JeroenDeDauw @malberts This PR is a demo for the fully backend-driven validation on the frontend. Please test it out and see if there are any UX degradation in the frontend. If there isn't, then I will proceed to remove the now duplicated. TypeScript validators.

For some context, I spent some time researching whether we need to duplicate frontend validation to provide an optimal experience, but it turns out that it is common to use server-driven validation even on the frontend (Laravel Precognition, Livewire, Symfony UX, etc.), and the performance and UX trade-offs aren't as bad as I thought.

extension.json
```diff
@@ -144,6 +144,10 @@
 			"description": "Whether write endpoints reject edits that introduce new constraint violations. When false, violations are surfaced in responses but writes always persist.",
 			"value": false
 		},
+		"NeoWikiValidationDebounceMs": {
+			"description": "Debounce delay in milliseconds before the subject editor sends a server-side dry-run validation request. 0 disables the timer, so validation only runs on blur or pre-save.",
+			"value": 300
+		},
```
malberts, line 153: This description doesn't match the behaviour. It says `0` means "validation only runs on blur or pre-save", but in `useSubjectValidation.ts` the `debounceMs <= 0` branch of `revalidate()` runs the validation immediately on every change, so at `0` a dry-run fires on every keystroke (the opposite of blur-only).

https://github.com/user-attachments/assets/51f3b516-9d23-43b8-84cd-b314bbff7e2b

Setting `0` to cut validation traffic actually produces the most. 

It seems like the description is what was supposed to happen, at least to me that makes more sense than firing a server side validation on every keypress, so I assume this is a bug. If the opposite was intended, then the description should be fixed.

malberts:
When creating a new subject with a new schema, these backend validation attempts return 404 errors ("schema not found"). While this is not user-visible, it seems like avoidable backend traffic.

[Screencast_20260624_222804.webm](https://github.com/user-attachments/assets/af88803e-72fc-4286-a169-bacbc2ae3e9e)

malberts:
When creating a new subject with an existing schema, the validation runs on all fields, marking fields I have not yet touched as invalid:

[Screencast_20260624_223502.webm](https://github.com/user-attachments/assets/4487402f-1d8e-4d64-beb2-d2ca60920d76)

Ignore the Status field's validation, that is a pre-existing issue. In this case, I changed Title, and Homepage got flagged as invalid. On a more complicated form this is going to scream at the user when they have not made any other mistakes.

extension.json
alistair3149: Yes it is a bug. It is patched now to match the description:

https://github.com/user-attachments/assets/a06565cc-ed8a-49fa-afa2-2aff0fcc197f

malberts:
> use server-driven validation even on the frontend (Laravel Precognition, Livewire, Symfony UX, etc.), and the performance and UX trade-offs aren't as bad as I thought.

This is one of the alternatives considered in ADR 21, and "rejected", but not necessarily as something that could not be a follow-up, as the point of the ADR was just about adding backing validation:
https://github.com/ProfessionalWiki/NeoWiki/blob/master/docs/adr/021-add-backend-validation.md#backend-only-validation-frontend-calls-api

However, we need to do some testing around the network round-trip point. While those mentioned frameworks are presumably examples of this working well, they have also likely been optimised for this use case. In our case this involves a MediaWiki round trip which might not necessarily be optimal for field validation. I have no evidence either way, so this is something we should test on the demo wiki first.

alistair3149:
> When creating a new subject with a new schema, these backend validation attempts return 404 errors ("schema not found"). While this is not user-visible, it seems like avoidable backend traffic.

This is a pre-existing issue that is not caused by this PR, but I fixed it as a drive-by.

https://github.com/user-attachments/assets/eb22c254-7514-4ad8-bfc5-7edcf8d79613

> When creating a new subject with an existing schema, the validation runs on all fields, marking fields I have not yet touched as invalid:

Agreed. I have adjusted the validation behavior:
- **Subject creator/New subject:** Do not validate on open - since users did nothing wrong and the field are going to be edited
- **Subject editor/Existing subject:** Validate on open - so that existing violations are visible in the editor.

https://github.com/user-attachments/assets/b1d3ba04-cc3c-40a2-badc-752573074388

> However, we need to do some testing around the network round-trip point.

I agree. That's also why we have `$wgNeoWikiValidationDebounceMs` as a knob to tune the behavior.

JeroenDeDauw:
> In our case this involves a MediaWiki round trip which might not necessarily be optimal for field validation.

Ah yes, instantiating MediaWiki to the point where we can run our validation code isn't cheap

alistair3149:
We can stress test the network RTT when it is deployed to the demo wiki.
The validation logic itself only adds **~2–3 ms** over a bare REST call, and the whole call is around 11ms.

---

To isolate how much of the round-trip is *our validation* vs MediaWiki request overhead, here are 30 warm samples of the dry-run endpoint taken in-container (so this excludes the browser↔server network)

| Endpoint | median | p90 | max |
|---|---|---|---|
| `POST /subject/validate` (dry-run) | 11 ms | 12 ms | 14 ms |
| `GET /schema-names` (baseline trivial REST) | 8 ms | 9 ms | 11 ms |

The empty-statements payload still ran a full pass (it returned `required` for every required property), and the dry-run loads the schema from cache and doesn't touch Neo4j, so this is representative rather than a no-op.

> <sub>Benchmark + write-up by Claude Code (`Opus 4.8`), at @alistair3149's direction.</sub>

malberts:
Confirmed my issues are fixed.

> When creating a new subject with an existing schema, the validation runs on all fields, marking fields I have not yet touched as invalid

This still happens with required date fields, though. For example, when you create a subject using the Everything schema:

Screenshot_20260625_234143

However, since this was not the original purpose of this PR, let's move that to a follow-up and merge.

alistair3149:
Ran the concurrency stress test on the NeoWiki Demo wiki (the network round-trip + load test we said we needed).

**Single request** (server-side, network RTT removed): ~28 ms of MediaWiki processing on the real deploy — the validation logic is ~0 of that, it's MW bootstrap (consistent with the dev-wiki number).

**Under concurrency** — at 20 concurrent requests the validate endpoint's tail blows up while a trivial endpoint at the same concurrency stays flat. Warm round-trip, ms; the ~140 ms floor is network RTT from the test location, so the signal is validate vs baseline, not the absolute numbers:

| Endpoint @ 20 concurrent | p50 | p95 | p99 |
|---|---:|---:|---:|
| `POST /subject/validate` (run 1) | 143 | 175 | **998** |
| `POST /subject/validate` (run 2) | 141 | 158 | **2596** |
| `GET /schema-names` (baseline) | 149 | 184 | 193 |

Zero errors throughout. So it's validate-specific — not the network, not my client.

**Root cause:** the dry-run loaded and re-parsed the Schema wiki page on every call, uncached, so a concurrent burst piles up on that load+parse (a cache stampede).

**Fix:** filed a separate PR — #943 — a `CachingSchemaLookup` (WANObjectCache, keyed by the schema page's latest revision so edits invalidate it transparently). Local before/after, ms:

| | C=1 (p50 / max) | C=20 (p50 / p99 / max) |
|---|---:|---:|
| before (no cache) | 14 / 29 | 48 / 63 / 65 |
| after (#943) | 12 / **12** | **34 / 50 / 50** |

Single-request tail flattens (29 → 12 ms) and the burst drops ~25–30 % at every percentile.

**Next:** we'll need to **re-run this C=20 burst on the Demo wiki after #943 deploys** to confirm the spike is gone on the real environment — the local before/after is a smaller-scale demo (the dev wiki didn't reproduce the full spike). #943 also has one open item (read-permission gating under caching) to settle before merge.

For context: ~100 req/s (C=20) is well beyond realistic load — the 300 ms debounce keeps real editor concurrency at low-tens of req/s, where everything stayed flat. So not a Demo-wiki blocker, but a genuine scalability finding worth fixing.

> <sub>Benchmark + write-up by Claude Code (`Opus 4.8`), at @alistair3149's direction.</sub>

alistair3149:
@JeroenDeDauw @malberts I have rerun the benchmark and the performance seems acceptable. What do you think?
If it is okay for both of you, then we can move to the next step of dropping the frontend validation code.

---

| C=20, warm round-trip p99 (ms) | validate | baseline | validate − baseline |
|---|---:|---:|---:|
| before (no cache) | 998 / 2596 | 193 | +805 … +2400 |
| after (#943) | 514 / 518 / 457 | 417 | +40 … +100 |

At p50/p95 validate now equals the baseline; only a ~40–100 ms residual remains at p99, down from the ~1–2.6 s spike. Zero errors across 800+ requests.

One observation: single-request server time was p50 ~24 ms but p95/p99 ~193 ms — most requests are warm cache hits, while a request landing on a not-yet-warmed worker still loads once. That's the signature of a per-worker cache (APCu) rather than a shared one. The stampede is gone either way; a shared cache (memcached/redis) on the Demo wiki would flatten that last cold-worker tail too, if we ever want it.

So the round-trip concern is resolved on the real deploy.

> <sub>Re-run + write-up by Claude Code (`Opus 4.8`), at @alistair3149's direction.</sub>

PR #888, opened 2026-06-11
Incorporate HW feedback (2026-06-11) into Subject Sources doc

JeroenDeDauw:
The 2026-06-11 BlueSpice call was HW's feedback on this proposal. It answered three of the four "needs HW input" questions and corrected one core assumption.

**Resolved**
- **In-View vs query-enough → query-enough.** HW renders dashboards over Cypher in their own UI and does not need page/approval metadata in the View system for the MVP. So View-rendering of sourced Subjects is off HW's MVP path — Sequencing now puts *materialise + query* first and demotes the in-View slice to "architecture-proving."
- **Page owner / approval → split, accepted.** Page owner is user-set (writeable Subject); approval is extension-managed (separate Subject); HW accepts the two-Subject split.
- **Farm topology → shared graph; per-wiki schemas.** Shared graph confirmed. Schemas are **per-wiki with a delivered common baseline** — *not* a single global set, replacing the earlier "schemas farm-global" assumption. Local-schema Subjects aren't cross-wiki renderable/editable via the View system; cross-wiki access must **degrade gracefully** (new requirement).
- **Materialisation → mandatory** (not read-time-only) — firmed up from inference to HW-confirmed.

**Free-form tables**
Dropped from the open questions and trimmed: free-form/Confluence tables need no core support beyond the **engineered-source plugin point** — an extension registers its own source (no core `WikitextSource`), so the wikitext-vs-editor choice is HW's product decision, not a Source-architecture one.

The schema reversal (global → per-wiki-with-baseline) and the graceful-degradation requirement are the two things most worth a second look against the call.

> Result of design back-and-forth with @JeroenDeDauw, cross-checked against a second Opus review of the same transcript.
> Context: the merged Subject Sources doc and the 2026-06-11 BlueSpice meeting transcript.
> <sub>Written by Claude Code, \`Opus 4.8 (1M context)\`</sub>

### DIFF
--- a/docs/planning/SubjectSources.md
+++ b/docs/planning/SubjectSources.md
@@ -23,8 +23,8 @@ Primary drivers (committed work):
 1. **Wiki-farm metadata.** A farm of separate wikis needs unified metadata: each page's metadata and approval state
    queryable, sortable, and filterable **across all wikis** (e.g. "all pages approved before date X, farm-wide"), plus
    cross-wiki references in query results. This is the main reason a graph database was chosen, and the primary
-   near-term driver. Cross-wiki queryability implies the metadata must be **materialised in the (shared) graph**, not
-   only synthesised on read.
+   near-term driver. HW confirmed (2026-06-11) that this metadata **must be materialised in the (shared) graph**, not
+   only synthesised on read — materialisation is mandatory for it to be queryable.
 2. **External-owned state (document control).** Approval state, last reviewer, etc. are owned by another extension.
    They must be queryable for dashboards, viewable alongside Subject data, updatable **without creating a page
    revision**, and survive a graph rebuild.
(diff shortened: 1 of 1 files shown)


semantisch:
On the Subject Sources direction:

The semantics of `editable` / `queryable` / `versioned` etc., combined with normal editing rights, seem too complex for users.

For users, structured data in the wiki/on the page should be viewable and editable, given the respective access rights. Everything else, like approval state owned by another extension, remote data, page metadata, cached external data etc., can of course be materialised/queryable for dashboards and integrations. But I would not include it in the regular Subject/View/editor model as if it were the same kind of page-owned structured data.

JeroenDeDauw:
@semantisch https://github.com/ProfessionalWiki/NeoWiki/pull/906

PR #896, opened 2026-06-18
Polish NeoWiki user rights

alistair3149:
Tested the browser check steps manually.

> Result of an iterative design discussion with @alistair3149.
> Context: the NeoWiki codebase and MediaWiki core's grant/permission system (`AvailableRightsTest`, `GrantsInfo`, `PermissionManager`) plus the grant usage of the bundled extensions.

Polishes the registration of NeoWiki's user rights (`neowiki-schema-edit`, `neowiki-layout-edit`, `neowiki-query`).

Three commits:

1. **Register the rights as grant permissions.** Without a grant, a right can never be exercised through an OAuth consumer or bot password — a client's rights are intersected with the grants it was authorised for. `neowiki-query` goes under `basic`; the two namespace-gate rights go under `editpage`, following how core gates the MediaWiki namespace via the `editinterface` right. Creating or deleting a Schema/Layout page then composes this gate grant with the standard `createeditmovepage`/`delete` grant, mirroring core.
2. **Add `right-*` messages** so the rights are described on Special:ListGroupRights, Special:UserGroupRights and the grant consent screens instead of shown as raw keys.
3. **Allow all logged-in users to edit schemas and layouts.** Move the two edit rights from `sysop`/`interface-admin` to `user`, matching how the MediaWiki ecosystem gates structural editing: extensions assign rights to existing core groups rather than reserving them for admins, and structured-data systems such as Wikifunctions let any logged-in user edit their typed objects. `interface-admin` was a poor fit (it exists for sitewide CSS/JS). Reads stay open to everyone via `neowiki-query`, and admins can still restrict via `$wgGroupPermissions`.

## Manual Browser Check

(Requires a wiki with the extension loaded.)

1. **Special:ListGroupRights** — the three `neowiki-*` rights appear with descriptions ("Create and edit NeoWiki schemas", "Create and edit NeoWiki layouts", "Run NeoWiki queries"), not raw keys.
2. **Edit as a normal user** — logged in as a non-admin account, edit a `Schema:` page; it should succeed (previously required sysop/interface-admin). Logged out, the same edit is denied, while the Cypher query API / `{{#cypher_raw}}` still works.
3. **Special:BotPasswords** — the "Edit existing pages" grant lists "Create and edit NeoWiki schemas/layouts".


malberts:
> Allow all logged-in users to edit schemas and layouts.

The current state came from the original 2023 implementation, and I don't recall any specific discussions, but I guess the idea at the time was that Schemas are not normal content, and therefore should not be editable by everyone.

I am not sure if we still want to treat these more strictly as _config_, or now more loosely as _content_. At least by default.

Otherwise, the change looks fine.

@JeroenDeDauw thoughts on this?

JeroenDeDauw:
Allowing schema edits by all users by default seems reasonable. It's what SMW has. The main thing is to have the ability for admins to restrict who can edit schemas, which we retain in this PR.

PR #897, opened 2026-06-18
Remove no-op page-move hook; moves already update the graph

### LINKED ISSUE #875: Update neo4j on page move
The `PageMoveComplete` hook is registered but its handler is empty, so moving/renaming a page does not update the graph projection. `Page.name` (and anything else derived from the title) stays stale until a full `RebuildGraphDatabases.php` run.

**Evidence**
- Registered: `extension.json` Hooks → `PageMoveComplete` → `NeoWikiHooks::onPageMoveComplete`
- Empty body: `src/EntryPoints/NeoWikiHooks.php:191`
- By contrast, edit/delete/undelete are handled: `src/EntryPoints/NeoWikiHooks.php:173,184,188`

**Impact**
After a page move, queries and views relying on page-node properties see stale data, with no indication, until someone runs a rebuild.

**Suggested direction**
Implement the handler to update the page node's title-derived properties (mirroring what `onRevisionCreated` does for the moved page).

---
> Surfaced during a Claude Code audit of NeoWiki's install/maintenance story (while planning sysadmin docs), at @alistair3149's direction. Part of a set of install-audit findings.\n> Context: the NeoWiki codebase.\n> <sub>Written by Claude Code, `Opus 4.8`</sub>

alistair3149:
Fixes https://github.com/ProfessionalWiki/NeoWiki/issues/875

The reported staleness does not occur. Every page move creates a null
revision on the new title (inheriting all slots, including the subject
slot), and core fires RevisionFromEditComplete for it synchronously.
NeoWiki already handles that hook and rewrites the page node, keyed by
the move-stable page ID, so Page.name and the other title-derived
properties update immediately, with no rebuild needed.

The empty onPageMoveComplete handler was therefore dead code, and what
made the move path look unhandled. Remove it and its registration.
Implementing it to re-save the page would only add a redundant second
write per move.

Add a regression test that moves a page with subjects and asserts the
graph node's name follows the new title, and drop the now-incorrect
"page moves leave Page.name stale" note from the maintenance docs.

> Result of investigating #875 at @alistair3149's direction; the
> false-positive finding and the remove-dead-code resolution were his call.
> Context: the NeoWiki codebase and MediaWiki core's MovePage internals.

### DIFF
--- a/docs/operations/maintenance.md
+++ b/docs/operations/maintenance.md
@@ -20,7 +20,7 @@ php maintenance/run.php NeoWiki:RebuildGraphDatabases
 It re-saves every Subject from each page's latest revision. Run it to:
 
 - Recover after a Neo4j wipe or restore.
-- Fix drift, such as the stale `Page.name` values left by page moves.
+- Fix any drift between the Neo4j copy and the canonical revision slots.
 
 Two things to plan around:
 
@@ -46,9 +46,7 @@ data may not survive an upgrade, so be ready to rebuild it from scratch.
 
 ## Current limitations
 
-Two known limitations while NeoWiki is pre-release:
+One known limitation while NeoWiki is pre-release:
 
-- **Page moves do not update the graph.** Moving or renaming a page leaves its `Page.name` in Neo4j stale until you
-  rebuild. Tracked in #875.
 - **A Neo4j outage blocks writes.** Edits, deletes, and undeletes fail while Neo4j is unreachable; only reads and
   queries keep working. Tracked in #877.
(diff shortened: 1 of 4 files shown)


malberts:
Was the original staleness observation in #875 based on only code analysis? Or was there actual staleness, perhaps caused by something else?

alistair3149:
> Was the original staleness observation in #875 based on only code analysis? Or was there actual staleness, perhaps caused by something else?

It was a stale code analysis based off the empty PageMoveComplete hook.
The reproduction in this issue debunked it, so I am dropping the hook and added an integration test to lock the behavior.

malberts:
Ready to merge then?

Issue #874, opened 2026-06-08
Neo4j uniqueness constraints are never created in production


alistair3149:
`Neo4jConstraintUpdater::createDefaultConstraints()` creates the `Page.id` and `Subject.id` uniqueness constraints, but it is only ever called from tests — no `src/` or `maintenance/` code invokes it. A real install therefore runs Neo4j **without** these constraints.

**Evidence**
- Definition: `src/GraphDatabasePlugins/Neo4j/Persistence/Neo4jConstraintUpdater.php:14`
- Only callers are tests: `tests/phpunit/GraphDatabasePlugins/Neo4j/Persistence/Neo4jConstraintUpdaterTest.php:34,74,91`
- No production caller (`grep -rn createDefaultConstraints src/ maintenance/` → none)

**Impact**
Duplicate-id nodes are possible and id lookups are unindexed, on every deployment.

The docs claim the opposite: `docs/reference/graph-model.md:93` states "Two uniqueness constraints are created on initialization", which is currently false.

**Suggested direction**
Call `createDefaultConstraints()` from an install/setup path (e.g. wire it into `maintenance/RebuildGraphDatabases.php` and/or a dedicated setup step), then correct the graph-model.md claim once it holds.

---
> Surfaced during a Claude Code audit of NeoWiki's install/maintenance story (while planning sysadmin docs), at @alistair3149's direction. Part of a set of install-audit findings.\n> Context: the NeoWiki codebase.\n> <sub>Written by Claude Code, `Opus 4.8`</sub>

malberts:
Part of this, or separately, the page ID constraints need to be changed as per #908.

PR #908, opened 2026-06-21
Scope Neo4j page-node identity per wiki

### LINKED ISSUE #905: Scope graph node identity per wiki (fix cross-wiki page-node collisions)
## Problem

In a shared-graph wiki farm (BlueSpice Galaxy), multiple wikis write to one Neo4j graph. MediaWiki page ids are per-wiki sequential, so `Page` nodes from different wikis collide on `id` and currently **overwrite each other** (found by @malberts). The `Page.id`-unique invariant (ADR 19 / `graph-model.md`) does not hold across wikis. HW also needs per-wiki query filtering for a near-term demo.

## What to do

- Stamp a **`wiki_id`** property — the [MediaWiki Wiki ID](https://www.mediawiki.org/wiki/Manual:Wiki_ID) (wiki name + table prefix) — on the page and subject nodes each install writes, enabling `WHERE n.wiki_id = '…'` filtering.
- Make **page-node identity wiki-scoped**: the `MERGE` key and uniqueness constraint become `(wiki_id, id)` rather than `id`, fixing the overwrite bug.
- Update **page-node lookups** in existing queries to be wiki-scoped (this ripples into other queries — per @malberts' note).
- **Subject ids stay bare nanoids** (collision is extremely unlikely); they just gain the `wiki_id` property. Subject-id namespacing is deferred.
- **Single-wiki setups**: behaviour unchanged.

## Scope / deferred

- HW writes the per-wiki filters themselves; **server-side ACL query filtering is out of scope** here.
- The full namespacing model — **Source-plugin-provided id prefixing**, applied to subject ids too (the `(source, localId)` model) — is deferred to the broader Subject Sources work. This is the forward-compatible first increment, using the same wiki identifier (MW Wiki ID) the full model will use; HW confirmed they accept the keying evolving later.

## Decision record

Direction recorded in the multi-wiki node-identity ADR (forthcoming). Revises the `Page.id`-unique statement in ADR 19 / `graph-model.md`.

Relates to #831 and the Subject Sources planning doc.

> Result of the Subject Sources design work with @JeroenDeDauw, and @malberts' finding that cross-wiki page ids collide in the shared graph.
> Context: the NeoWiki graph model, the 2026-06-11 HW farm decision, and the MediaWiki Wiki ID manual.

JeroenDeDauw:
Fixes https://github.com/ProfessionalWiki/NeoWiki/issues/905

In a shared-graph wiki farm, MediaWiki page ids are per-wiki sequential, so `Page` nodes from different wikis collided on `id` and overwrote each other.

Stamp a `wiki_id` property (the MediaWiki Wiki ID) on page and subject nodes, and scope page-node identity to the `(wiki_id, id)` pair: the `savePage` MERGE, the page-node lookups (delete, subjects-of-page, HasSubject linking) and the uniqueness constraint now key on both properties. Subject ids stay bare, globally unique nanoids and only gain the `wiki_id` property for filtering; per-subject-id namespacing is deferred to the Subject Sources work.

`wiki_id` is resolved once at wiring time (`WikiMap::getCurrentWikiId()`) and injected into the Neo4j layer, keeping that layer free of MediaWiki globals. The domain `Page` and `Subject` are unchanged. Single-wiki installs are unaffected: with one `wiki_id`, `(wiki_id, id)` behaves exactly like `id`.

Implements the design in ADR 22 (#907) and updates `graph-model.md` accordingly. **Depends on #907** for the ADR file (it should merge first).

> Result of implementing @JeroenDeDauw's design (issue #905 + ADR #907, from the Subject Sources work) and @malberts' cross-wiki collision finding.
> Context: the NeoWiki Neo4j persistence layer, ADRs (esp. 1, 14, 19), ADR #907, graph-model.md, and the MediaWiki Wiki ID manual.

### DIFF
--- a/docs/reference/graph-model.md
+++ b/docs/reference/graph-model.md
@@ -28,9 +28,15 @@ via Relations.
 Every page that contains structured data has a corresponding `:Page` node. These nodes make page metadata available
 for graph queries.
 
+Page identity is scoped per wiki. A shared graph can hold pages from multiple wikis (e.g. a wiki farm), and MediaWiki
+page ids are only unique within a single wiki. Each page node therefore carries a `wiki_id` and is identified by the
+`(wiki_id, id)` pair ([ADR 22](../adr/022-multi-wiki-node-identity.md)). In a single-wiki install there is one
+`wiki_id` value and behaviour is unchanged.
+
 | Property | Neo4j Type | Description |
 |----------|------------|-------------|
-| `id` | integer | MediaWiki page ID (unique) |
+| `wiki_id` | string | [MediaWiki Wiki ID](https://www.mediawiki.org/wiki/Manual:Wiki_ID) (database name + table prefix) of the owning wiki |
+| `id` | integer | MediaWiki page ID (unique per wiki) |
 | `name` | string | Page title |
 | `creationTime` | datetime | When the page was created |
 | `lastUpdated` | datetime | When the page was last modified |
@@ -51,6 +57,11 @@ their Schema (e.g., `:Subject:Person`, `:Subject:Company`). The Schema label cha
 |----------|------------|-------------|
 | `id` | string | Subject ID, 15 characters starting with `s` (unique) |
 | `name` | string | Subject label |
+| `wiki_id` | string | [MediaWiki Wiki ID](https://www.mediawiki.org/wiki/Manual:Wiki_ID) of the wiki that owns the Subject |
+
+Unlike page ids, Subject ids are globally unique nanoids ([ADR 14](../adr/014-improved-id-format.md)), so a Subject's
+identity is its `id` alone. The `wiki_id` is carried only for per-wiki query filtering in a shared graph; Subject-id
+namespacing is deferred ([ADR 22](../adr/022-multi-wiki-node-identity.md)).
 
 ### Dynamic properties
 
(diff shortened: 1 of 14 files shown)


malberts:
Reviewed this and exercised it on a two-wiki shared-graph setup (two wikis projecting into one Neo4j). The core change is correct: every page id shared between the two wikis resolved to two distinct `(wiki_id, id)` nodes, and a full `RebuildGraphDatabases` of one wiki left the other wiki's nodes untouched. **Nothing below blocks merge.**

### No migration for pre-existing graphs

Existing `:Page`/`:Subject` nodes have no `wiki_id`, and `MERGE (page:Page {id, wiki_id})` can't match a node whose `wiki_id` is null, so re-saving a page (or running `RebuildGraphDatabases`, which doesn't clear first; its description already assumes an out-of-band wipe) creates a duplicate node and orphans the original. Reproduced on a live graph. There's no in-app path today (no backfill command), so a pre-#908 install needs a manual step: either backfill (`MATCH (n) WHERE n.wiki_id IS NULL SET n.wiki_id = '<wikiId>'` for a single-wiki graph), or wipe the graph and rebuild.

There is no requirement to have migration, but the drop and rebuild should be communicated to current users.

### Wiki ID uniqueness is a downstream config concern

`wiki_id` is whatever `WikiMap::getCurrentWikiId()` returns, i.e. `$wgDBname` (plus prefix). NeoWiki has no way to tell that two requests arriving with the same DB name are meant to be separate wikis. So a farm built from separate DB services that reuse the same internal database name (a common container-per-wiki setup; this dev stack itself defaults every stack to the same name) gives both wikis the same `wiki_id`, and their page nodes collide again in the shared graph. There's nothing NeoWiki can detect here, so flagging it as a downstream configuration requirement to document for the shared-graph case: each wiki in a shared graph must have a distinct Wiki ID (distinct dbname or `$wgDBprefix`).

Multiple databases in one database container already avoids this issue, because they necessarily have unique names.

### Explicitly deferred in ADR 22 / the PR, confirmed in testing

ADR 22 and the description defer per-subject-id namespacing and push read filtering onto consumers. Both surfaced in the shared-graph test; recording to confirm they behave as documented:

- **Subjects overwrite across wikis when subject ids coincide.** Subjects are `MERGE`d by bare `id`, so two wikis carrying the same subject id collapse to one node (last writer wins). Matches the deferred design. One thing to carry into the Subject Sources work: the "collision extremely unlikely by construction" premise holds for randomly minted nanoids, but not for delivered/shared content with fixed ids (the demo data's `s1demo…` ids, reused across both wikis, triggered it) nor for ids supplied through the page JSON, so the overwrite is reproducible in practice, not only theoretical.
- **`getPageIdOfSubject` (`Neo4jPageIdentifiersLookup`) is unscoped** and leans on the same one-subject-one-page invariant: no `wiki_id` filter, no `LIMIT`/`ORDER BY`, returns the first row. It would return an arbitrary page if a subject id is ever shared across wikis.
- **Read paths (subject-label lookup, the Cypher REST endpoint) aren't wiki-filtered**, per the consumer-applies-`wiki_id` / ACL-out-of-scope decision. Pre-existing and out of scope here; noting only.

### Constraint type

`(wiki_id, id)` is created as a property-uniqueness constraint; consider `IS NODE KEY` instead, so it also requires `wiki_id` to be present. Property-uniqueness ignores null-`wiki_id` nodes (exactly the migration orphans above), whereas a node key rejects them. Partially tracked in #874, where the constraint is actually applied.

PR #814, opened 2026-05-07
Redesign DemoData content

alistair3149:
Fixes https://github.com/ProfessionalWiki/NeoWiki/issues/793

> Result of extensive back-and-forth iteration with @alistair3149.
> Context: the NeoWiki extension codebase, the existing demo wiki, the audiences listed in #793, and the BlueSpice / ECHOLOT context documents.
> <sub>Written by Claude Code, `Opus 4.7 (1M context)`</sub>

## Summary

Replaces the previous feature-index demo wiki with three use-case hubs and supporting infrastructure. Visitors land on a showcase `Main_Page` with cards linking to:

- **Museum collection**: refined European Museums dataset (ECHOLOT/CH/GLAM audience)
- **ACME Inc**: fictional corporate knowledge base (KM and IT-lead audience)
- **Research catalog**: new research-output dataset (academic and ECHOLOT case-study audience)

Plus a **Developers hub** replacing four standalone capability pages with a single source-then-rendered reference for the programmable surface (Cypher, parser functions, Lua, REST API, Reactive UI, special pages).

## Notable changes

- 6 new schemas: Department, Office, Publication, Author, Institution, Project
- 24+ new subject pages including the expanded ACME corpus and full research catalog
- 3 new Lua modules: `Module:Card` for the Main_Page, `Module:FeaturedRow` for hub featured-subject rows, `Module:LuaExample` for the Developers hub
- `Module:NeoWikiDemo.query` extended with `columns=` and `linkColumns=` for ordered and linked Cypher result tables
- 51 subjects gain prose wikitext (1-2 paragraphs each, Wikipedia conventions: bolded title on first mention, wikilinks to other demo subjects)
- New `DemoData/AGENTS.md` documenting authoring conventions, wikitext and Cypher gotchas, reusable modules, the hub skeleton, and Codex CSS guidance
- Retirements: Berlin / NeoWiki / ProWiki / Professional_Wiki subjects, the four standalone capability pages, three orphan schemas

## Test plan

- [ ] `make reinstall-db && make load-test-data` succeeds with zero `FAILED` lines
- [ ] `Main_Page` renders the hero panel and three cards; each card link resolves
- [ ] All four hub pages (`Museum_collection`, `ACME_Inc`, `Research_catalog`, `Developers`) render their five-section structure
- [ ] Embedded Subject views render in Featured rows; the row scrolls horizontally on narrow viewports
- [ ] Cypher result tables on hubs render with linked subject names
[...]

PR #910, opened 2026-06-22
Direct first-time visitors on the demo Main Page

JeroenDeDauw:
## What

Redesigns the demo wiki `Main_Page` to **direct** a first-time visitor on what to do, instead of only describing NeoWiki.

- Adds a role-based **Start here** section (Just exploring / Trying it hands-on), restoring the on-ramp the pre-#814 page had.
- Restores the "this demo exists for feedback" framing, with issue-tracker and contact links.
- Reframes the sandbox note from a warning ("edits may not persist") into an invitation to edit.
- Leads with the **Museum collection** live query (clearest proof of "queryable structured data", and the GLAM/ECHOLOT-relevant dataset) and the multiple-subjects-per-page **Rijksmuseum** Data tab.
- Uses **Subject views** (a Sandbox subject) as the safe hands-on editing demo.
- **Validation Demo** is linked but deliberately not a headline action: its inline rejection only fires with `$wgNeoWikiEnforceValidation = true`, which its own page documents as a `LocalSettings.php` change, so featuring it could mislead a hosted-demo visitor who sees nothing happen.
- Keeps the existing hero and use-case cards. Removes one em-dash per house style.

## Why

Prep for the NeoWiki soft-launch outreach (ECHOLOT rapid user-testing task force first). The outreach email should point people at the demo and let the wiki explain itself; today the homepage describes but does not direct.

## Notes

- Content-only change under `DemoData/`. Needs a demo re-import (`make load-test-data`) to appear on neowiki.dev.
- Rendering not verified locally; relies on review and the import step.

### DIFF
--- a/DemoData/Page/Main_Page.wikitext
+++ b/DemoData/Page/Main_Page.wikitext
@@ -3,6 +3,22 @@
 <div style="color:var(--color-subtle, #54595d);max-width:560px;margin-top:4px">Schemas, infoboxes, relations, and Cypher, native to MediaWiki.</div>
 </div>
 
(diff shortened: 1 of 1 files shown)


JeroenDeDauw:
image

alistair3149 (approved):

PR #911, opened 2026-06-22
Demo & dev environment: fix `make down`, add `make demo`/`make pull`, tidy install docs

JeroenDeDauw:
> Result of iterative work with @JeroenDeDauw: he hit the `make up` / `make down` inconsistency, then drove the broader demo/dev-environment cleanup.
> Context: the NeoWiki dev-environment `Makefile`, `Docker/docker-compose*.yml`, and the install docs. Tested on both `podman-compose` and the docker compose plugin, plus a real isolated `make demo` run.

Tidies the demo and developer environments, which had drifted apart.

**`make down` fix (the original bug).** `make up` brings up the profile-less try-it-out stack, but `make down` used `--profile dev` + the dev overlay, so after the documented `make up` flow it tried to remove the `mailcatcher`/`node`/`test_neo` containers it never created — noisy "no such container" errors (loud on podman-compose, which the install docs target). The two compose backends have opposite requirements (podman-compose errors when `--profile dev` is set but the containers are absent; the docker compose plugin needs the profile to remove containers gated in a loaded file), so the fix moves the dev-only sidecars from the base file into `docker-compose.dev.yml` and lets a base-file `down --remove-orphans` reap them as orphans. Clean on both backends for the `make up` and `make dev` teardown paths.

**Install docs.** README routes to the demo (`installation.md`) vs dev (Development section). `Docker/README.md` drops the try-it-out/server instructions it duplicated from `installation.md` (links to them instead) and keeps only its unique reference content — reserved host ports (an anchor target of inbound links) and the file layout — with the Dockerfile target labels corrected (`final-mw`, not `production-mw`, is the published `:latest` demo).

**`make demo` / `make pull`.** `make demo` is a one-command evaluator setup (pull, start, install, seed) mirroring `make dev`, idempotent on re-run. `make pull` refreshes the published image, which `make up` alone never does once it's cached. `reset` gains a comment explaining its intentional `$(DC)`-down / `$(DC_DEV)`-up asymmetry and a corrected help string.

### DIFF
--- a/Docker/README.md
+++ b/Docker/README.md
@@ -72,10 +72,11 @@ inside the stack via `make bash` or `docker compose exec`.
 
 - `Dockerfile` — multi-stage build with `production-mw` (prebuilt demo), `final-mw`
   (release tag), and `dev-mw` (the dev image with mounted NeoWiki source).
-- `docker-compose.yml` — base services (`mediawiki`, `db`, `neo`, plus profile-gated
-  `test_neo`, `node`, `mailcatcher`, `caddy`).
+- `docker-compose.yml` — base "try-it-out" services (`mediawiki`, `db`, `neo`) plus
+  the profile-gated `caddy` (the `server` profile, for HTTPS hosting).
 - `docker-compose.dev.yml` — dev overlay; switches `mediawiki` to the dev image,
-  bind-mounts the NeoWiki source, and sets `MW_MODE=dev`.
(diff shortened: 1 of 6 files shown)


JeroenDeDauw:
@alistair3149 @malberts already merged to make testing the dev env demo easier, but you might still want to do a sanity check. This is fully approved by me and mostly driven by me. Follow to our chat earlier today and the recent changes alistair made to the repo setup and install docs.

PR #912, opened 2026-06-22
Add Extending NeoWiki documentation

alistair3149:
> Result of a brainstorming session with @alistair3149 (chose a lean, reference-first scope; requested this review + UX copy pass).
> Context: the NeoWiki codebase, the RedHerb example extension, and the existing `docs/` + glossary.
> <sub>Written by Claude Code, `Opus 4.8` (1M context)</sub>

## What

Adds a published reference page, `docs/reference/extending.md`, documenting NeoWiki's extension points:

- **Backend (PHP):** Property Types, Neo4j value builders, Page Property Providers, the `NeoWikiGetFrontendModules` hook, and read-side services via `NeoWikiExtension::getInstance()`.
- **Frontend (JS/Vue, no TypeScript build):** the `neowiki.registration` hook + `registerPropertyType`, the `ext.neowiki` public API, and mounting standalone Vue features.
- **Conventions** (i18n/validation codes, icons) and a **"Not yet extensible"** section (View Types, graph backends, `.d.ts`).

Each extension point links to the RedHerb source that demonstrates it. The RedHerb README is slimmed to an example index pointing at the new page, and the page is linked from the docs index and the glossary. RedHerb remains the live, test-backed example.

A second commit reconciles the glossary's *View Type* entry (which previously implied extensions could register new View Types) with the new page's "Not yet extensible" status.

## Notes for reviewers

- The published page is the canonical reference; RedHerb's README is now just an index. Snippets and claims were verified against the current source.
- Links into `tests/RedHerb` use absolute GitHub URLs because `tests/` isn't part of the published docs tree; the only over-120-char lines are those unbreakable URLs (consistent with existing reference pages).
- **Sidebar placement** (frontmatter `order: 7` in the `reference` group) renders on neowiki.ai post-merge — worth a glance once deployed.


JeroenDeDauw:
@alistair3149 did you already fully review this yourself, vouching for every word?

alistair3149:
Yes I did a read-through manually and compare it to the existing RedHerb codebase and documentation. Though it is best that @malberts take a second look because he is more knowledgeable on RedHerb and the extensibility topic, and if there are any additions that should be noted.

JeroenDeDauw:
I undid the changes to the glossary and actually went a bit the other direction there. The glossary is not an extensibility guide. Only thing that fits there is a differentiation between a set being fixed and open.

JeroenDeDauw (approved):

Issue #899, opened 2026-06-19
ext-neowiki-view only rendered in  Vector 2022 skin


LibrErli:
In our neowiki-dev installation, the `#ext-neowiki-view` is only renderd in Vector 2022 skin (in neowiki.dev it shows also up in the legacy skin).  Is it something we have to configure on our site?

krabina:
We also tested chameleon skin, which is an important skin for MediaWiki. Also does not work there.

malberts:
Do you have more information about:
* installation steps
* environment/MW version/LocalSettings
* are there any errors in the browser developer console?

It renders on Chameleon/Timeless/Medik/Monobook on our demo wiki: https://neowiki.dev/wiki/ACME_Inc (only one skin is available at a time on here, but these screenshots are from there):

<img width="1131" height="1110" alt="Image" src="https://github.com/user-attachments/assets/fd9ccd15-700d-4b7f-ad36-55569768335a" />

<img width="1218" height="1147" alt="Image" src="https://github.com/user-attachments/assets/9b044b20-377c-4a0d-a05e-a4679d3cba5d" />

<img width="1218" height="1147" alt="Image" src="https://github.com/user-attachments/assets/2ef3d911-cfe7-4304-becb-cdae756d7474" />

<img width="1234" height="939" alt="Image" src="https://github.com/user-attachments/assets/ad9de150-58c4-4441-9737-d64dfd8e2e35" />

malberts:
@LibrErli I found the error message on https://neowiki.knowledge.wiki/index.php/ACME_Inc. Investigating.

malberts:
Latest master branch has a fix for this.

PR #914, opened 2026-06-22
Clone only the bundled MediaWiki submodules NeoWiki loads

JeroenDeDauw:
> Result of back-and-forth with @JeroenDeDauw, who set the direction and the keep-set.
> Context: the NeoWiki dev env — the `bootstrap` Makefile target, `Docker/SettingsTemplate.php` (what actually loads), the cloned MW core's `.gitmodules`, and a footprint measurement of the existing clone.

`make dev`'s `bootstrap` cloned MediaWiki core with `--recurse-submodules`, pulling all 37 bundled extension/skin submodules. NeoWiki's `Docker/SettingsTemplate.php` only loads 9 of them (6 extensions + 3 skins); the other 28 were fetched and never enabled.

This clones core without `--recurse-submodules` and inits only the loaded set, cutting the bundled-submodule fetch from 37 to 9 repositories (~520 MB less to download and check out on a fresh bootstrap).

Safe because those 9 are exactly what `SettingsTemplate.php` loads: CodeEditor's only dependency (WikiEditor) is included, `--recursive` pulls VisualEditor's `lib/ve`, and the dropped submodules are never referenced. The list must stay in sync with `SettingsTemplate.php`; an added comment notes this.

Forward-only: existing checkouts keep their full bundle until re-bootstrapped from scratch.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -101,7 +101,7 @@ bootstrap: ## Clone MW core into Docker/mediawiki/ and prep gitignored files (id
 			"$${MW_GIT_URL:-https://github.com/wikimedia/mediawiki}" \
 			Docker/mediawiki; \
 		echo "Fetching the bundled extensions/skins NeoWiki loads..."; \
-		git -C Docker/mediawiki submodule update --init --recursive \
+		git -C Docker/mediawiki submodule update --init --recursive --depth 1 \
 			extensions/CodeEditor \
 			extensions/ParserFunctions \
 			extensions/Scribunto \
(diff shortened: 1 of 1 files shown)


JeroenDeDauw:
Verified locally

Makefile
```diff
@@ -87,14 +87,30 @@ _dev-tools-impl:
...
+		git -C Docker/mediawiki submodule update --init --recursive \
+			extensions/CodeEditor \
+			extensions/ParserFunctions \
+			extensions/Scribunto \
+			extensions/SyntaxHighlight_GeSHi \
+			extensions/VisualEditor \
+			extensions/WikiEditor \
+			skins/MonoBook \
+			skins/Timeless \
+			skins/Vector; \
```
alistair3149, line 113: We can probably skip timeless and monobook

malberts: It might be useful to have non-Vector skins around for issues like https://github.com/ProfessionalWiki/NeoWiki/issues/899, although the rest of our config masked that specific issue anyway.

Issue #796, opened 2026-04-29
Add scriptable View Type


JeroenDeDauw:
> Drafted by @JeroenDeDauw with Claude Code, after a back-and-forth to capture design discussions that aren't yet recorded in an issue, ADR, or planning doc.
> Context: ADR 018 (Views), the View Type plugin system, the existing `mw.neowiki` Lua API, and prior design discussions.
> <sub>Written by Claude Code, `Opus 4.7`</sub>

## Goal

Provide a built-in View Type that delegates rendering to a Lua module, so template authors can produce arbitrary output for a Subject (or set of Subjects) without waiting for a new built-in View Type to be added to NeoWiki core.

This is a plug-in on the View Type plugin system (#659) — i.e. it's how we eat our own dog food on the plug-in API, and it's the escape hatch that makes adding new built-in View Types less urgent.

## Why

- **Unblocks dashboards and custom layouts.** End-user use cases (HW's QM dashboard, ad-hoc Subject renderings) currently require either waiting for a dedicated View Type or working around the system entirely. A scriptable View Type lets these be authored in-wiki.
- **Composes with the rest of the Lua API.** A scripted view can internally use `nw.getMainSubject()`, `nw.query()` (#736), and `nw.getSchema()` (#737) to render dashboards over query results, not just single Subjects.
- **Stress-tests the View Type plug-in API.** If the plug-in surface is rich enough to express \"hand the data to Lua and let it render,\" it's likely rich enough for most third-party View Types.

## In scope

- A View Type that, when used by a Layout, dispatches rendering to a configured Lua module.
- Wiring through the existing `{{#view}}` invocation path so referencing a Layout that uses this View Type Just Works.
- Sensible behaviour when the Lua module is missing, errors out, or returns something unexpected.
- Tests covering the happy path and the failure modes.

## Out of scope

- Built-in table / list / factbox View Types — separate issues.
- A visual builder for scripted views.
- Sandboxing concerns beyond what MediaWiki's Scribunto already provides.

## Things to figure out (not prescriptions)

- **What the Lua entry point receives.** Options include: the Subject data as a table, the Layout's settings, the Display Rules (so Lua can honour them), or some combination. There's a tradeoff between giving Lua everything and keeping the API small.
- **How the Lua module is referenced from the Layout.** A module name is the obvious starting point; whether to also pin a function name or rely on a convention is open.
[...]

Issue #785, opened 2026-04-27
RfC: Server-side rendering of Views


alistair3149:
> Concern raised and framed by @alistair3149, who independently identified the parser-cache, edge-cache, SEO, and blank-flash consequences. I verified the rendering path and wrote up this problem statement; no solution is proposed — that's for discussion.
> Context: the NeoWiki codebase, https://neowiki.dev/wiki/Professional_Wiki, and verification of the rendering path in browser DevTools.
> <sub>Written by Claude Code, `Opus 4.7`</sub>

## Problem

NeoWiki Views (the `{{#view}}` parser function and the auto-rendered Main Subject View) emit no HTML during wikitext parsing. The infobox is rendered entirely client-side by Vue, mounted onto a `<div class="ext-neowiki-view" data-mw-neowiki-subject-id="...">` element that is injected outside `mw-parser-output`. The wikitext parser output for these regions is empty.

## Verification

On https://neowiki.dev/wiki/Professional_Wiki , anonymous and signed out, `document.querySelector('.mw-parser-output').innerHTML` returns only parser-cache metadata comments — no Subject content, no infobox markup. The visible "Professional Wiki GmbH" infobox appears only after Vue boots and the data fetches resolve.

## Consequences

- **No parser cache value for Views.** MediaWiki's parser cache stores wikitext output. Since Views produce empty output, the parser cache holds nothing reusable for the View region; the per-view saving is zero.
- **No edge / file cache value for Views.** A reverse proxy or CDN caching the rendered page HTML caches a page with no View content; the viewer still has to wait for JavaScript and the API.
- **No content for SEO / crawlers.** Crawlers that don't execute JavaScript (and any other tool reading server-rendered HTML) see a page with no infobox text, no property labels, and no relation targets. Subject data is invisible to such consumers.
- **Blank flash on every anonymous view.** The infobox appears only after page HTML → JS boot → Vue mount → fetch main Subject → fetch each related Subject → fetch Schema. Even on a fast network, anonymous viewers see a layout shift / blank region until the data resolves.
- **No graceful degradation.** A user with JavaScript disabled, or a mid-load failure of `ext.neowiki`, sees no Subject content at all.

Together these mean an anonymous viewer pays the full data-fetching cost on every visit; none of MediaWiki's caching layers help the View.

## Scope

This RfC concerns the rendering architecture for Views. It does not propose a solution; the design space is large enough that several stakeholders should weigh in before a direction is chosen.

[...]

PR #921, opened 2026-06-24
Expose View Type registration to extensions

### LINKED ISSUE #919: Expose View Type registration to extensions
> Result of a review discussion with @JeroenDeDauw: while reconciling the glossary with the new *Extending NeoWiki* docs, we found View Types are half-exposed, and decided to track closing that gap as sequenced-for-later rather than exposing it now.
> Context: the NeoWiki codebase (`ViewTypeRegistry`, `public-api.ts`, the `neowiki.registration` frontend registrar, RedHerb), the `docs/reference/extending.md` reference, and issues #659, #796, #785.
> <sub>Written by Claude Code, `Opus 4.8` (1M context)</sub>

## Goal

Third-party extensions can register custom View Types, at parity with how they can already register Property Types.

## Current state

The View Type plug-in system (#659) exists — `ViewTypeRegistry` is part of the public API and `infobox` is registered through it — but there is no wired-up path for third-party extensions:

- the `neowiki.registration` hook exposes no View Type registration (there is no equivalent of `registerPropertyType`);
- `NeoWikiExtension.getViewTypeRegistry()` returns a fresh registry per call, so a third-party registration would not reach NeoWiki's own rendering;
- RedHerb has no View Type example.

There is also no published component contract for View Types, unlike Property Types (`ValueDisplayContract`, `ValueInputContract`, `AttributesEditorContract`). A View Type is currently just a raw Vue component with an implicit prop shape, validated only by `infobox`.

## Sequence before exposing

1. Define and publish the View Type component contract (analogous to the `Value*Contract` files), so extensions have a stable interface to implement.
2. Validate that contract with a second consumer — the scriptable View Type (#796) and/or a built-in table/card — before freezing it.
3. Resolve the server-side-rendering direction (#785); a client-only Vue-component registration contract may not survive it.

## Scope when unblocked

- Add a registration entry point: `registerViewType` on the frontend registrar (plus a backend registrar method if needed).
- Make the View Type registry durable so third-party registrations reach rendering.
- Add a RedHerb example.
- Document it in `docs/reference/extending.md` (move View Types out of "Not yet extensible").

## Related

- #659 — View Type plug-in system (closed; the registry)
- #796 — Add scriptable View Type (first internal consumer)
- #785 — RfC: Server-side rendering of Views

JeroenDeDauw:
> Result of pair work with @JeroenDeDauw: investigated #919's prerequisites, implemented the client-only path per direction to proceed, then browser-verified — which surfaced and fixed a registration/render race.
> Context: the NeoWiki codebase (`ViewTypeRegistry`, `FrontendRegistrar`, `NeoWikiApp`, `public-api.ts`, `Infobox`, RedHerb), issues #919/#659/#796/#785, and open PR #912.
> <sub>Written by Claude Code, `Opus 4.8` (1M context)</sub>

Fixes https://github.com/ProfessionalWiki/NeoWiki/issues/919

Extensions can now register custom View Types via the `neowiki.registration` hook, at parity with Property Types.

- `FrontendRegistrar.registerViewType( { typeName, component } )`, alongside `registerPropertyType`.
- `NeoWikiExtension.getViewTypeRegistry()` is memoized so a hook registration reaches the registry used for
  rendering (it previously returned a fresh registry per call).
- The registry is **reactive**. An extension's frontend module can finish loading after `NeoWikiApp` has resolved the
  page's views, so a registration arriving late (common on a warm cache, where the Subject/Schema/Layout load is fast)
  now re-triggers resolution and swaps the fallback infobox for the registered component, instead of leaving the wrong
  view rendered.
- Published the `ViewTypeProps` component contract (`subjectId`, `canEditSubject`, `layoutName`) and the
  `ViewTypeRegistration` DTO from `ext.neowiki`. `Infobox` consumes `ViewTypeProps` via `defineProps<ViewTypeProps>()`,
  matching how value components consume `ValueDisplayProps`.
- RedHerb gains a `redherb-card` example: a two-column term/value card (loosely modelled on a document-control header)
  that renders values through `nw.resolveDisplayProperties` + the value-display component registry and offers editing
  via the shared `nw.SubjectEditorDialog` when `canEditSubject` — demonstrating reuse of NeoWiki's components and
  stores from a separate, plain-JS extension. Documented in RedHerb's README.

Verified in-browser: across cold and (previously failing) warm reloads, a Layout using `redherb-card` consistently
renders the card; each value renders through its property type's component; the edit button opens the reused
`SubjectEditorDialog`; and existing infoboxes still render after the `Infobox` prop change.

**Sequencing:** #919 sequences this behind #785 (SSR of Views) and #796 (a second View Type consumer to validate the
contract); both remain open, and this lands the client-only registration path per direction to proceed. If #785
settles on server-side rendering, the client-only registration contract may need revisiting.

**Docs follow-up:** the issue's "move View Types out of *Not yet extensible* in `docs/reference/extending.md`" is
deferred — that file is introduced by open PR #912 and is not on master, so editing it here would conflict. It should
be updated as part of #912 (or a follow-up once #912 merges). The new extension point is documented in
`tests/RedHerb/README.md` in the meantime.


JeroenDeDauw:
image

JeroenDeDauw:
> Result of a review pass requested and scoped by @JeroenDeDauw, who directed surfacing only the points not already covered in the PR description.
> Context: the NeoWiki codebase (`ViewTypeRegistry`, the Property Type registries, `NeoWikiApp`, `FrontendRegistrar`, RedHerb), ADR 018 (Views), and issues #919/#785/#796/#912.
> <sub>Written by Claude Code, `Opus 4.8` (1M context)</sub>

Two design notes from review that aren't covered above — both judgment calls, not blockers.

**1. The reactivity fix is view-type-only; the sibling registries that use the same hook stay non-reactive.**

This PR makes `ViewTypeRegistry`'s backing Map `reactive()` so a registration arriving after `NeoWikiApp` resolved its views re-triggers resolution. But `PropertyTypeRegistry` and `TypeSpecificComponentRegistry` are populated through the *same* synchronous `neowiki.registration` hook and remain plain `Map`s. The warm-cache race this fixes for views looks reachable for a **custom Property Type rendered in a built-in Infobox**: if the extension module arrives after the Infobox first paints, `getValueDisplayComponent( '<custom>' )` throws `Unknown property type` and, with a non-reactive registry, doesn't recover on a later render.

Plausible rather than confirmed — it depends on ResourceLoader batch/cache timing I didn't pin down — but it's the same mechanism. Worth either confirming custom Property Types survive that warm-cache path, or treating "registration-backed registries are reactive" as one consistent decision rather than a view-only special case.

**2. The published contract is narrower than ADR 018's View Type plugin contract.**

ADR 018 §"View Types" says a View Type plugin defines three things: *what `settings` it supports*, *how to render*, and *editing UI for its type-specific options*. `ViewTypeProps` / `ViewTypeRegistration` cover only rendering. So `redherb-card` reads `layout.getSettings().fullWidthProperties` untyped/ad-hoc, and that setting can only be authored by hand-editing the Layout JSON (as `DemoData/Layout/SandboxCard.json` does) — `LayoutCreator` lets you select the View Type but not configure its settings.

Reasonable MVP slice given the 0.x/alpha note on `public-api.ts`, but it likely means `ViewTypeRegistration` grows a settings/editing surface later. Flagging so the gap against ADR 018 is explicit rather than implied-complete.

JeroenDeDauw:
Needs squash-merge

PR #927, opened 2026-06-25
Add an event-centric Person/Birth demo corpus

malberts:
## What this adds

A new `DemoData/` corpus modelling people and their births in the event-centric style of CIDOC CRM (the cultural-heritage reference model Arches is built on), where a Birth is its own Subject rather than flat fields on a Person. It covers three generations of the Bach family (8 persons, 5 births, 4 places).

New schemas: `Person` (CIDOC E21), `Birth` (E67), `Place` (E53). A hub page, `People and life events`, presents the data with two Cypher tables and a NeoWiki-to-CIDOC mapping table (with cidoc-crm.org links), plus a card on the Main Page.

## Why these choices

- **An event, not fields.** Modelling Birth as its own Subject lets one event tie together a child, mother, father, place, and date, and lets a person appear in many events in different roles. J. S. Bach is the child in his own Birth and the father in three of his sons' Births; Johann Ambrosius is the child in his and the father in J. S. Bach's. Flat `Born`-style fields cannot express that.
- **Its own page, not a child Subject.** Each Birth is a standalone Subject on its own page, not a child Subject nested under a person's page. A child Subject has to live on a single participant's page, which would imply that participant has primacy (the person born over their parents, say); an event with several participants shouldn't have to nominate a primary owner. Keeping the Birth separate sidesteps that. (It could technically be a child Subject; this is a deliberate modelling choice.)
- **A musician, not a painter.** The demo already models painters under the `Artist` schema. A musician overlaps no existing schema, so this corpus does not create two competing representations of the same kind of entity.
- **CIDOC directionality preserved.** The only stored edges run Event to Person (`P98 brought into life`, `P96 by mother`, `P97 from father`, `P7 took place at`). The Person-side "was born" (inverse `P98i`) is not stored as a second edge; it is derived by a new `personEvents` function in `Module:NeoWikiDemo` that reverse-queries the graph. This mirrors how a triplestore (and Arches) asserts a CIDOC property once and reads its inverse off the same edge, and it keeps `Person` deliberately lean.
- **Honest, partial mapping.** Where the demo simplifies, the mapping table says so. `Place.Country` is labelled a denormalised country-name label with no direct CIDOC property (a faithful model would use `P89 falls within` to a containing `E53 Place`); the `Date` row notes the value sits inside the `E52 Time-Span` (via `P82`), collapsed here to one date.
- **Open-world scope, not a data gap.** The root Birth (Johann Ambrosius) records a child, place, and date but no parents. That is a deliberate three-generation scope boundary, not a claim that his parents are unknown (they are documented), framed so it does not misrepresent the record.

## Date storage

NeoWiki currently stores `date` values as a string list, so the Cypher reads `b.Date[0]`. When typed `date` storage lands (the mechanism from #812, which currently covers only `dateTime`), this becomes a one-token change to `toString(b.Date[0])` in the `personEvents` function and the two hub queries; nothing else changes.

## Verification

Imported via `make dev` into a local stack and checked in the browser:
- Import completes with no failures.
- `Schema:Person` / `Birth` / `Place` exist; Birth subjects show participants, place, and date.
- Person pages render the derived life-events table (J. S. Bach born in Eisenach, father in three births; Maria Barbara mother in two; Anna Magdalena mother in one), confirming the reverse-query and the multi-role / two-mothers behaviour.
- The hub's two Cypher tables and the CIDOC mapping table populate; the Main Page shows the new card.

## Screenshots

|  |  |
|---|---|
| **An event as its own Subject:** the Birth links child, mother, father, place, and date. | birth-of-johann-sebastian-bach |
| **A lean Person beside derived life events:** the Person infobox carries no birth fields; "was born" and "father in" are computed by reverse-querying the events. | person-johann-sebastian-bach |
| **The hub:** Cypher tables over the Bach family plus the NeoWiki-to-CIDOC mapping table. | hub-people-and-life-events |
| **A Place subject:** Country and Coordinates. | place-eisenach |
| **The Birth schema** (CIDOC E67) and its properties. | schema-birth |
| **The hub linked from the demo Main Page.** | main-page-card |

### DIFF
--- a/DemoData/Module/NeoWikiDemo.lua
+++ b/DemoData/Module/NeoWikiDemo.lua
@@ -278,4 +278,82 @@ function p.schema( frame )
 	return renderRowsAsTable( rows, { 'Name', 'Type', 'Required', 'Details' } )
 end
 
+local PERSON_EVENTS_MONTHS = {
+	'January', 'February', 'March', 'April', 'May', 'June',
+	'July', 'August', 'September', 'October', 'November', 'December'
+}
+
+-- Formats an ISO yyyy-mm-dd string as "21 March 1685" without relying on
+-- MediaWiki's date parsing, which is unreliable for pre-modern years.
+local function personEventsFormatDate( iso )
+	local y, m, d = string.match( iso or '', '^(%d+)-(%d+)-(%d+)$' )
+	if not y then
+		return iso or ''
+	end
+	return tonumber( d ) .. ' ' .. ( PERSON_EVENTS_MONTHS[tonumber( m )] or m ) .. ' ' .. y
+end
+
+local function personEventsBirthLinks( rows )
+	local links = {}
+	for _, row in ipairs( rows or {} ) do
+		links[#links + 1] = '[[' .. row.birth .. ']]'
+	end
+	return links
+end
+
+-- Renders the current Person page's life events by reverse-querying the graph:
+-- the canonical edges run Birth -> Person, so "was born" (CIDOC P98i) is derived
+-- here rather than stored as an inverse edge. Date is read as b.Date[0] because
+-- the `date` property is stored as a string list; switch to toString(b.Date[0])
+-- if/when date is stored as a native Neo4j date.
+function p.personEvents( frame )
+	local name = mw.title.getCurrentTitle().text
+
+	local born = nw.query(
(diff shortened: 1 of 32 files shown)


malberts:
I did not manually verify the CIDOC IDs and rules yet, only via Claude Code. However, the final NeoWiki demo data structure does reflect the kind of modelling approach where the statements are complex, and they cannot be stored as relations on the person subject.

PR #928, opened 2026-06-25
Add NeoWikiAutoRenderMainSubject config to disable automatic Main Subject rendering

alistair3149:
musee-dorsay-autorender-off

## Summary

Adds a global boolean config `NeoWikiAutoRenderMainSubject` (default `true`). When set to `false`, a page's Main Subject is no longer rendered automatically as an infobox on content pages. This lets setups that render subjects through the web API or place views explicitly with `{{#view}}` opt out of the built-in automatic rendering.

Default behaviour is unchanged.

## Changes

- **`extension.json`** — new `NeoWikiAutoRenderMainSubject` config key (default `true`).
- **`NeoWikiExtension::shouldAutoRenderMainSubject()`** — reads the value live from `MainConfig` per request, following the existing `NeoWikiEnforceValidation` behavioural-config pattern (overridable in tests via `overrideConfigValue()`, no singleton rebuild).
- **`NeoWikiHooks::handleContentPage()`** — a guard clause skips only the automatic `mainSubjectHtml()` injection when disabled. The ResourceLoader module load and the `#ext-neowiki-app` container (subject-creator UI) still run, and explicit `{{#view}}` is unaffected — it resolves through `ViewParserFunction` and the static `ViewHtmlBuilder::viewPlaceholderHtml()`, never `mainSubjectHtml()`.
- **`docs/operations/installation.md`** — documents the setting.

No frontend change: when no placeholder is emitted, there is nothing for the frontend to mount.

## Testing

New integration test `AutoRenderMainSubjectTest` (`@group Database`) drives `onBeforePageDisplay` and asserts on the rendered HTML:

- default config → the `ext-neowiki-view` placeholder is present with the subject id;
- `NeoWikiAutoRenderMainSubject = false` → no `ext-neowiki-view` placeholder;
- `NeoWikiAutoRenderMainSubject = false` → the `#ext-neowiki-app` container is still present.

`make phpunit filter=AutoRenderMainSubjectTest` → `OK (4 tests, 4 assertions)`. `make cs` clean.

Manual smoke test against demo data (set `$wgNeoWikiAutoRenderMainSubject = false;`, observe rendered HTML, then revert):

- main-subject page (`Johannes_Vermeer`): the auto-infobox placeholder disappears while `#ext-neowiki-app` remains; reverting restores it;
- mixed page (`Musée_d'Orsay`): the auto main-subject placeholder is removed, while the three explicit `{{#view}}` placeholders still render — confirming explicit views are unaffected.

> Result of a brainstorming session with @alistair3149, who set the scope (suppress only the automatic Main Subject infobox, leaving explicit `{{#view}}` and the subject-creator UI intact), the single-global-boolean granularity, and the config name.
> Context: the NeoWiki codebase and project context docs.

### DIFF
--- a/tests/phpunit/EntryPoints/AutoRenderMainSubjectTest.php
+++ b/tests/phpunit/EntryPoints/AutoRenderMainSubjectTest.php
@@ -36,7 +36,7 @@ class AutoRenderMainSubjectTest extends NeoWikiIntegrationTestCase {
 
 		$html = $this->renderContentPage( 'Auto render disabled', $revision->getId() );
 
-		$this->assertStringNotContainsString( 'ext-neowiki-view', $html );
+		$this->assertStringNotContainsString( self::SUBJECT_ID, $html );
 	}
 
 	public function testAutoRendersMainSubjectByDefault(): void {
@@ -47,7 +47,7 @@ class AutoRenderMainSubjectTest extends NeoWikiIntegrationTestCase {
 
 		$html = $this->renderContentPage( 'Auto render default', $revision->getId() );
 
-		$this->assertStringContainsString( 'data-mw-neowiki-subject-id="' . self::SUBJECT_ID . '"', $html );
+		$this->assertStringContainsString( self::SUBJECT_ID, $html );
 	}
 
 	public function testRendersAppContainerWhenAutoRenderDisabled(): void {
--- a/tests/phpunit/EntryPoints/AutoRenderMainSubjectTest.php
+++ b/tests/phpunit/EntryPoints/AutoRenderMainSubjectTest.php
@@ -1,0 +1,79 @@
+<?php
+
+declare( strict_types = 1 );
+
+namespace ProfessionalWiki\NeoWiki\Tests\EntryPoints;
+
+use MediaWiki\Context\RequestContext;
+use MediaWiki\Title\Title;
+use ProfessionalWiki\NeoWiki\EntryPoints\NeoWikiHooks;
+use ProfessionalWiki\NeoWiki\Tests\Data\TestSubject;
+use ProfessionalWiki\NeoWiki\Tests\NeoWikiIntegrationTestCase;
+
+/**
+ * @covers \ProfessionalWiki\NeoWiki\EntryPoints\NeoWikiHooks::onBeforePageDisplay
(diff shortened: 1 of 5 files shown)


tests/phpunit/EntryPoints/AutoRenderMainSubjectTest.php
```diff
@@ -0,0 +1,79 @@
...
+
+	public function testAutoRendersMainSubjectByDefault(): void {
+		$revision = $this->createPageWithSubjects(
+			'Auto render default',
+			TestSubject::build( id: self::SUBJECT_ID )
+		);
+
+		$html = $this->renderContentPage( 'Auto render default', $revision->getId() );
+
+		$this->assertStringContainsString( 'data-mw-neowiki-subject-id="' . self::SUBJECT_ID . '"', $html );
```
JeroenDeDauw: I wonder if we can just check for `self::SUBJECT_ID` to reduce binding to implementation details. Would that check fail if we disable  `NeoWikiAutoRenderMainSubject`?

tests/phpunit/EntryPoints/AutoRenderMainSubjectTest.php
```diff
@@ -0,0 +1,79 @@
...
+		$revision = $this->createPageWithSubjects(
+			'Auto render disabled',
+			TestSubject::build( id: self::SUBJECT_ID )
+		);
+
+		$this->overrideConfigValue( 'NeoWikiAutoRenderMainSubject', false );
+
+		$html = $this->renderContentPage( 'Auto render disabled', $revision->getId() );
+
+		$this->assertStringNotContainsString( 'ext-neowiki-view', $html );
```
JeroenDeDauw: Was this assertion mutation tested? How come the next test doesn't have the same string check but then with `assertStringContainsString`?

JeroenDeDauw (commented):
I am looking at the review myself

JeroenDeDauw (approved):

tests/phpunit/EntryPoints/AutoRenderMainSubjectTest.php
alistair3149: No it should still work, checking the data attribute is more strict but probably not necessary

PR #932, opened 2026-06-25
Add dev-environment preflight checks and make doctor

alistair3149:
image

> Directed by @alistair3149 across an extended back-and-forth: questioned PR #930's compose-only check, set the scope (lean gate + warnings, Docker-runtime only, no base-tooling audit), asked for the standalone `make doctor` plus the Docker-presence and `make pull` gating, and gave the Compose-messaging steer (drop the "v2" qualifier and distro package name, keep the canonical docs URL). Three adversarial review rounds and smoke tests were run at their direction.
> Context: the NeoWiki dev-env Makefile and `Docker/scripts` tooling, the merged `_check-compose` (#930) it evolves, and the dev-env setup failure-mode history.

Follows-up to https://github.com/ProfessionalWiki/NeoWiki/pull/930

`_check-compose` (#930) fails fast when `docker compose` is missing, but its probe is client-side: it passes when the Docker daemon is down or the socket is permission-denied — the two most common first-timer states — which then fail late, after the multi-hundred-MB MediaWiki-core clone, with the same cryptic error #930 set out to prevent.

This evolves `_check-compose` into a tested `Docker/scripts/preflight.sh`, exposed as the `_preflight` lifecycle gate and a standalone `make doctor`. Scope is the Docker runtime only — the one prerequisite that genuinely varies between hosts; base tooling (git, curl, coreutils) is treated as a given.

**Hard checks** (abort before any expensive work):
- Docker (or a compatible runtime) present at all — short-circuits with an install-Docker message instead of redundant downstream errors.
- Docker Compose available (`docker compose version`).
- Daemon reachable and permitted (`docker info`) — closes the gap #930 leaves open.

**Warnings** (printed, exit 0):
- Container engine — surfaces a Podman misdetection (a stray `podman` binary makes the tooling run as the container user, writing root-owned files into bind mounts).
- Dev port range 8484-8499 fully saturated.

Checks are collect-all (compose + daemon report together) with inline fix guidance. `_preflight` gates `up`/`demo`/`dev`/`dev-tools`/`pull`. Tested via `Docker/tests/test-preflight.sh` (a `docker` stub seam, 25 checks) wired into `make test-scripts`.

## Manual check
- `make doctor` on a healthy host → prints the ✓ list (compose, daemon, engine, free port) and exits 0; appears in `make help`.
- `make test-scripts` → runs the new 25-check suite alongside the existing set-port suite.
- Failure paths (point `DOCKER_BIN` at a stub, or remove it): missing compose, daemon down, both, and an absent docker binary each abort with the right message and exit non-zero; `make pull`/`make up` run the preflight before any container action.

### DIFF
--- a/Docker/scripts/preflight.sh
+++ b/Docker/scripts/preflight.sh
@@ -12,11 +12,19 @@ set -u
 
 DOCKER_BIN="${DOCKER_BIN:-docker}"
 PREFLIGHT_VERBOSE="${PREFLIGHT_VERBOSE:-}"
+PODMAN_BIN="${PODMAN_BIN:-podman}"
+PORT_RANGE_START="${PORT_RANGE_START:-8484}"
+PORT_RANGE_END="${PORT_RANGE_END:-8499}"
 
 failed=0
 
 pass() { [ -n "$PREFLIGHT_VERBOSE" ] && printf '  \033[32m✓\033[0m %s\n' "$1"; return 0; }
 fail() { printf '  \033[31m✗\033[0m %s\n' "$1" >&2; failed=1; }
+warn() { printf '  \033[33m!\033[0m %s\n' "$1" >&2; }
+
+is_port_free() {
+    ! (echo > "/dev/tcp/127.0.0.1/$1") 2>/dev/null
+}
 
 # Compose v2 present. `docker compose version` is a client-side probe and does not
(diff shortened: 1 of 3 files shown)


JeroenDeDauw (approved):

JeroenDeDauw:
I added a commit on top

PR #943, opened 2026-06-26
Cache schema lookups to avoid re-loading the Schema page per request

alistair3149:
> Result of the performance investigation prompted by the validation round-trip concern on #886, at @alistair3149's direction.
> Context: the NeoWiki codebase + a load test of the dry-run validation endpoint on the neowiki.dev demo wiki.

Follows-up to #886, addressing the validation round-trip performance concern raised there.

## Problem

A load test of the dry-run validation endpoint on the neowiki.dev demo wiki showed a **validate-specific tail-latency spike under concurrency**: at 20 concurrent requests, `POST /subject/validate` p99 jumped to ~1–2.6 s (across two runs), while a trivial REST endpoint (`/schema-names`) at the same concurrency, vantage and client driver stayed flat (p99 ~193 ms). The median was unaffected; only the tail spiked, and only for the validate endpoint.

Root cause: every `getSchema()` call loaded the Schema wiki page (revision + content) and re-deserialized it, with **no caching** (`WikiPageSchemaLookup` → `PageContentFetcher` → `RevisionLookup` + parse, per call). Under a concurrent burst, requests pile up on that load+parse — a cache stampede.

## Fix

A `CachingSchemaLookup` decorator over the existing `SchemaLookup`, backed by `WANObjectCache`:

- Keyed by the Schema page's **article id + latest revision id**, so editing a schema produces a new key → transparent invalidation (no stale schemas, no hooks).
- `getWithSetCallback` provides cross-request caching plus stampede protection.
- Behavior-preserving: every consumer of `getSchemaLookup()` (validation, rendering, save) gets the same Schema, just cached.
- Read permission (`Authority::probablyCan('read', …)`) is checked **before** the cache, so a cache hit can't serve a Schema to a user who can't read its page — the cache only holds user-independent content. (A cheap in-memory check on the already-resolved `Title`.)
- Replica-lag aware: the populate read passes the replica's lag to WANObjectCache (`getCacheSetOptions`), so a value generated from a lagged replica gets a reduced TTL and self-heals instead of being pinned under the new revision's key.

## Per-path impact

`getSchemaLookup()` feeds every schema-dependent operation, so this helps more than the dry-run — but by different amounts. Loading a schema reads raw revision content and deserializes it; it never runs the parser, so it isn't covered by the parser cache. The exception is rendering: a `{{#view}}` is part of its host page's parser-cached `ParserOutput`, so there the schema is only loaded on a parser-cache miss.

| Path | Parser-cached? | Schema loaded | Benefit |
|---|---|---|---|
| Validation (dry-run REST) | No | every request | **direct** — the stampede path |
| Save (create / replace + graph projection) | No (write) | every save | **direct** |
| Schema fetch (`GET …/schema`) | No | every call | **direct** |
| Rendering (`{{#view}}` on a page view) | Yes | only on a parser-cache miss (edit / expiry) | indirect — most views are already absorbed by the parser cache |

## Verification

- **Unit test:** load-once (cache hit), reload-on-revision-change (invalidation), null-result caching, missing-page and read-denied short-circuits.
- **Local before/after load test** (dev wiki, temporarily `CACHE_DB`): single-request tail **29 ms → 12 ms** (the load+parse cached away); C=20 burst **~25–30 % lower at every percentile** (p50 48→34, p99 63→50), zero errors.
- **Smoke:** validate returns correct violations with the decorator active on both `CACHE_NONE` (no-op) and `CACHE_DB`; a non-existent schema cleanly 404s.
- cs (phpcs + phpstan) clean; `ValidateSubjectApi` integration test green.

The local wiki did **not** reproduce neowiki.dev's dramatic spike (different environment — no network, local DB, fewer workers), so the before/after is a modest-scale demonstration; the real win is larger on the loaded wiki where the spike appeared.

## Notes

- The title is resolved in both the decorator (for the cache key) and the inner (for the content). They share `LinkCache`, so the second resolution is an in-process hit; resolving once would mean leaking a `Title` through the Application-layer `SchemaLookup` interface, which isn't worth it.

## Deployment note

The fix needs an object cache to do anything: no-op on `CACHE_NONE`, per-worker on `CACHE_ACCEL`/APCu, full cross-process stampede protection on a shared cache (memcached/redis/DB). The **definitive confirmation** is to re-run the C=20 burst on neowiki.dev after this deploys (where the spike was), which also confirms neowiki.dev has an object cache configured.

### DIFF
--- a/src/NeoWikiExtension.php
+++ b/src/NeoWikiExtension.php
@@ -87,6 +87,7 @@ use ProfessionalWiki\NeoWiki\Persistence\MediaWiki\Subject\MediaWikiSubjectRepos
 use ProfessionalWiki\NeoWiki\Persistence\MediaWiki\Subject\PointInTimeSubjectLookup;
 use ProfessionalWiki\NeoWiki\Persistence\MediaWiki\Subject\StatementDeserializer;
 use ProfessionalWiki\NeoWiki\Persistence\MediaWiki\Subject\SubjectContentDataDeserializer;
+use ProfessionalWiki\NeoWiki\Persistence\MediaWiki\CachingSchemaLookup;
 use ProfessionalWiki\NeoWiki\Persistence\MediaWiki\LayoutPersistenceDeserializer;
 use ProfessionalWiki\NeoWiki\Persistence\MediaWiki\WikiPageSchemaLookup;
 use ProfessionalWiki\NeoWiki\Persistence\MediaWiki\WikiPageLayoutLookup;
@@ -440,10 +441,14 @@ class NeoWikiExtension {
 	}
 
 	public function getSchemaLookup(): SchemaLookup {
-		return new WikiPageSchemaLookup(
-			pageContentFetcher: $this->getPageContentFetcher(),
-			authority: $this->getRequestAuthority(),
-			schemaDeserializer: $this->getPersistenceSchemaDeserializer()
+		return new CachingSchemaLookup(
+			schemaLookup: new WikiPageSchemaLookup(
+				pageContentFetcher: $this->getPageContentFetcher(),
+				authority: $this->getRequestAuthority(),
+				schemaDeserializer: $this->getPersistenceSchemaDeserializer()
+			),
+			cache: MediaWikiServices::getInstance()->getMainWANObjectCache(),
+			titleFactory: MediaWikiServices::getInstance()->getTitleFactory()
 		);
 	}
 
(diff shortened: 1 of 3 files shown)


JeroenDeDauw:
This is _fine_, but let's hold off with more caching work. Now is not the right time. Let's first build and then optimize when it becomes important.

alistair3149:
> This is _fine_, but let's hold off with more caching work. Now is not the right time. Let's first build and then optimize when it becomes important.

This is only required for scaling the backend validation per the benchmark in https://github.com/ProfessionalWiki/NeoWiki/pull/886#issuecomment-4810906975, so that we can make the decision of whether backend validation is performant enough for wider deployment.

PR #950, opened 2026-06-26
Implement RelationAttributesEditor

### LINKED ISSUE #383: Create RelationAttributesEditor
For use in SchemaEditor

See `NumberAttributesEditor` and `TextAttributesEditor` as examples

AC: 
* [x] Attributes Editor template exists and is registered with the `registry` in `NeoWikiExtension.ts`

Further tasks, can all be done in separate PRs:
* [ ] Template contains `relation` input. Mandatory. This is the "name" of the relation. In the backend we are calling it the "relation type". The property name might be "Main product", while this attribute is something like "Has main product". The later is what we store in Neo4j and thus what gets used in Cypher queries. While the former, the Property name, is used for display of Statments in the Views and Subject Editor. Might be worth revisiting this topic to make sure we think this is the best approach before implementing.
* [ ] Template contains `targetSchema` input. Mandatory. Ability to search/select or similar
* [ ] Template contains `multiple` input (same as in `TextAttributesEditor`)
* [ ] The component emits event on change for all of the attributes

Related:
* https://github.com/ProfessionalWiki/NeoWiki/issues/387

alistair3149:
Fixes https://github.com/ProfessionalWiki/NeoWiki/issues/383

> Result of extensive back-and-forth with @alistair3149: the design discussion — keeping the relation edge label as a field distinct from the property name (grounded in ADR 017/019, the graph model, and the demo data), the verbatim-default-from-name decision, and the finding that there is deliberately no naming convention — is theirs, as is the manual catch of the target-schema display bug.
> Context: the NeoWiki codebase, the ADRs (017 names-as-identifiers, 019 graph DB architecture), the graph-model / schema-format / subject-format reference docs, and the demo schemas.

## Summary

Fills in the `RelationAttributesEditor` so the Schema editor can configure a relation property's three type-specific attributes. The editor receives the property and emits `update:property` with a partial, matching the existing `AttributesEditor` contract.

- **Relation type** (`relation`) — the Neo4j edge label, stored separately from the property **name** (the display label used in Views and the Subject editor). It defaults verbatim to the property name when empty (derive-on-create only — never re-synced to later name edits, which would silently rewrite the query surface), and is overridable. No transform is applied: a survey of the demo data shows no naming convention to honour (`Has product`, `Located in`, `Created by`, and several where the relation equals the name).
- **Target schema** (`targetSchema`) — chosen via the store-backed `SchemaLookup`, generalized here with a `selected` prop and relocated to `components/common/` (alongside `SubjectLookup`) for reuse.
- **Allow multiple** (`multiple`) — a checkbox, mirroring `UrlAttributesEditor`.

### Why the edge label stays a separate field

The relation edge label is the Cypher/RDF query surface; the property name is an editable display label. Decoupling protects queries from cosmetic renames and keeps the demo's predicate-shaped labels (e.g. `City` → `Located in`). The broader "display name vs canonical internal name" split that property-name i18n will eventually require — at which point the edge label folds into a universal internal name — is left for a future ADR.

### Also in this PR

- **Fixes `SchemaLookup` not showing a pre-selected value** — it set `CdxLookup`'s `selected` but not `input-value`, so a stored target schema rendered blank. Now binds `v-model:input-value` (as `SubjectLookup` does).
- Removes a vestigial `uniqueItems` parse line on the relation property type (no consumer; relations carry no `uniqueItems` on either the TS or PHP side).

### Known follow-up (not in this PR)

Required-field errors are inline-only and do not block Save. This is a pre-existing, Schema-editor-wide limitation shared by the Number/Text/URL editors and the property-name field; gating Save on attribute-editor validity is cross-cutting work that belongs in its own change.

## Manual Browser Check

1. Log in and open a schema with relation properties — e.g. `Schema:Company`.
2. Click **Edit schema** and select the `Main product` relation property.
3. Confirm **Relation type** shows `Has main product`, **Target schema** shows `Product`, and **Allow multiple values** is unchecked.
4. Select `Products` → confirm **Allow multiple values** is checked and **Target schema** is `Product`.
5. **Add property definition**, switch its **Type** to relation → confirm **Relation type** pre-fills with the property name; search and pick a target schema; toggle **Allow multiple values**.
6. Clear the relation type field → confirm the inline `Relation type is required.` error appears.
7. **Save schema**, reopen the schema, and confirm the relation / target schema / multiple values round-trip.


alistair3149:
https://github.com/user-attachments/assets/e34cb022-1298-43d1-9f2b-5b854006108c

PR #951, opened 2026-06-27
Fix relation-type and target-schema editing in the schema editor

JeroenDeDauw:
Follows-up to https://github.com/ProfessionalWiki/NeoWiki/pull/950

> Result of @JeroenDeDauw's review of #950 and a follow-on design discussion: he requested the review, diagnosed the target-schema field as a free-text input that should be a constrained picker, and chose the searchable-dropdown direction.
> Context: the NeoWiki codebase, PR #950, the sibling lookups, and a live browser check on Schema:Company.

Two editing fixes in the Schema editor's relation property, both stemming from fields that let the displayed value drift away from what gets stored.

## Relation type (free-text edge label)

`updateRelation` now emits the trimmed value even when empty, and the `relation` watcher no longer falls back to the property name. Clearing the field shows the required error and updates the model instead of silently keeping the previous value. The name fallback remains the on-create default (initial value + `onMounted`), preserving the "derive-on-create only" behaviour from #950.

## Target schema (reference to an existing schema)

The field is a `CdxLookup`, so typed text could drift from the actual selection (e.g. `Produc` while `Product` stayed selected) with no feedback — and that stale value could be saved by changing another field. It is now a **constrained picker**: it only ever holds an existing schema or nothing.

- **Pick** from the menu commits the schema.
- **Empty + blur** clears it (and the relation editor shows "Target schema is required.").
- **Unmatched text + blur** reverts to the last committed value.

Filtering still uses the existing server-side prefix search (`schema-names` API). `SubjectCreatorDialog`, which shares `SchemaLookup`, already ignores empty selections, so its behaviour is unchanged.

Required-field errors remain inline-only and still do not block Save — that schema-editor-wide gating was deferred in #950 and is unchanged here.

## Known follow-up (not in this PR)

"Browse the full schema list before typing" is not included: the `schema-names` API is a prefix search capped at 10 results that rejects an empty query, so there is no cheap way to preload all names. Enabling browse needs a small new list-all-schema-names endpoint.

## Tests

- `RelationAttributesEditor`: clearing emits an empty relation; a cleared field stays empty (no name snap-back); whitespace-only is trimmed/required; `null` is passed to the lookup when no target schema is set.
- `SchemaLookup`: server search + result rendering; pick commits; unmatched text reverts on blur; empty + blur clears; an unset field is not cleared.
- Full vitest suite green (1055 tests); build + lint clean.

## Manual browser check (done; recorded here for re-verification)

On `Schema:Company` → **Edit schema** → select **Main product**:

1. **Relation type:** clear it → stays empty, shows "Relation type is required."; type `  Owns  ` → stored trimmed.
2. **Target schema:** backspace `Product` → `Produc`, click away → reverts to `Product`.
3. **Target schema:** type `Off`, pick **Office** from the dropdown → commits, no mid-type blanking.
4. **Target schema:** clear it, click away → empty + "Target schema is required."

### DIFF
--- a/resources/ext.neowiki/src/components/SchemaEditor/Property/RelationAttributesEditor.vue
+++ b/resources/ext.neowiki/src/components/SchemaEditor/Property/RelationAttributesEditor.vue
@@ -25,7 +25,7 @@
 				{{ $i18n( 'neowiki-property-editor-target-schema' ).text() }}
 			</template>
 			<SchemaLookup
-				:selected="property.targetSchema ?? null"
+				:selected="property.targetSchema || null"
 				@select="updateTargetSchema"
 			/>
 		</CdxField>
@@ -54,7 +54,7 @@ const emit = defineEmits<AttributesEditorEmits<RelationProperty>>();
 const relationInput = ref( props.property.relation || props.property.name.toString() );
 
 watch( () => props.property.relation, ( newValue ) => {
-	relationInput.value = newValue || props.property.name.toString();
+	relationInput.value = newValue;
 } );
 
 onMounted( () => {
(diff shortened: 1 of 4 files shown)


JeroenDeDauw:
Folded into #953 for a single holistic review. #953 now targets `master` and contains all of this work — the relation-type clearing carries over unchanged, and the target-schema picker here (CdxLookup-based) was fully replaced there by a CdxCombobox one, so reviewing them separately would have meant reviewing code that no longer exists.

PR #953, opened 2026-06-27
Relation-property editing: clearable relation type and browsable target-schema picker

JeroenDeDauw:
Follows-up to https://github.com/ProfessionalWiki/NeoWiki/pull/950 (which implemented the RelationAttributesEditor for #383). **Supersedes #951** — that work is folded in here for a single holistic review, because #951's `CdxLookup`-based target-schema picker was fully replaced by the `CdxCombobox` one below, so reviewing them separately meant reviewing code that no longer exists.

> Result of extensive back-and-forth with @JeroenDeDauw: he directed the fixes from the #950 review, chose the browsable-picker direction and the reuse-the-summaries-endpoint approach, and caught the empty-box-browse, eager-validation, and field-width issues in live testing.
> Context: the NeoWiki codebase, #950/#951, the Codex Lookup/Combobox behaviour, and live browser checks on Schema:Company.

Two parts to relation-property editing in the Schema editor.

## Relation type (free-text edge label)

Clearing the field now updates the model instead of silently keeping the previous value; the property name is used only as the on-create default (initial value + `onMounted`), never re-synced to later name edits. Whitespace is trimmed.

## Target schema (reference to an existing schema)

A browsable, constrained picker built on `CdxCombobox`:

- **Browse:** clicking the field opens the full schema list (empty or filled), with descriptions.
- **Filter:** typing narrows it (client-side, instant).
- **Constrained:** only an exact existing schema commits (pick from the list or type it exactly); unmatched text reverts to the committed value on blur, so the field never holds an invalid value.
- **Validation:** the required error appears only after the field has been touched, not immediately on a newly added relation property.
- Full-width, with a "Select a schema" placeholder.

Backed by a cached, paginated all-schemas summary data source (reusing `GET /neowiki/v0/schemas` — no new backend). The shared `SchemaPicker` component also upgrades the subject creator's schema picker.

## Commits

1. Add an all-schemas summary data source
2. Add a browsable, constrained target-schema picker
3. Reflect cleared attributes in the relation property editor

## Tests & checks

Full vitest suite green (1057); build + lint clean. Browser-verified on `Schema:Company` → Main product: browse-on-empty-click, filter, pick, type-exact commit, revert unmatched on blur, clear → required-after-touch, full-width.

### DIFF
--- a/extension.json
+++ b/extension.json
@@ -413,7 +413,7 @@
 				"neowiki-subject-creator-schema-title",
 				"neowiki-subject-creator-existing-schema",
 				"neowiki-subject-creator-new-schema",
-				"neowiki-schema-lookup-placeholder",
+				"neowiki-schema-picker-placeholder",
 				"neowiki-schema-display-property-name",
 				"neowiki-schema-display-property-type",
 				"neowiki-schema-display-property-required",
--- a/extension.json
+++ b/extension.json
@@ -302,6 +302,7 @@
 				"CdxToggleSwitch",
 				"CdxToggleButtonGroup",
 				"CdxLookup",
+				"CdxCombobox",
 				"CdxTable",
 				"CdxInfoChip"
 			],
@@ -412,7 +413,7 @@
 				"neowiki-subject-creator-schema-title",
 				"neowiki-subject-creator-existing-schema",
 				"neowiki-subject-creator-new-schema",
-				"neowiki-subject-creator-schema-search-placeholder",
+				"neowiki-schema-lookup-placeholder",
 				"neowiki-schema-display-property-name",
 				"neowiki-schema-display-property-type",
 				"neowiki-schema-display-property-required",
--- a/i18n/en.json
+++ b/i18n/en.json
@@ -103,7 +103,7 @@
 	"neowiki-subject-creator-schema-title": "Have an existing schema?",
 	"neowiki-subject-creator-existing-schema": "Use existing",
 	"neowiki-subject-creator-new-schema": "Create new",
-	"neowiki-schema-lookup-placeholder": "Select a schema",
+	"neowiki-schema-picker-placeholder": "Select a schema",
 	"neowiki-subject-creator-label-field": "Subject label",
 	"neowiki-subject-creator-label-placeholder": "Enter a label for the subject",
 	"neowiki-subject-creator-save": "Create subject",
--- a/i18n/en.json
+++ b/i18n/en.json
@@ -103,7 +103,7 @@
 	"neowiki-subject-creator-schema-title": "Have an existing schema?",
 	"neowiki-subject-creator-existing-schema": "Use existing",
 	"neowiki-subject-creator-new-schema": "Create new",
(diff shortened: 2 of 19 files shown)


JeroenDeDauw:
[Screencast_20260627_203933.webm](https://github.com/user-attachments/assets/ba2c3189-b41f-4ced-8d3c-b0f70c19ba45)

JeroenDeDauw:
Pushed two review-driven follow-up commits:

**Harden and simplify the schema picker** (`610089c0`)
- `getAllSchemaSummaries` now pages by request offset, not by loaded count — a read-restricted or malformed Schema page (counted in `totalRows` but skipped by the loader) previously produced duplicate entries in the cached picker list.
- `SchemaPicker`'s `onMounted` now handles a failed schema load instead of leaving an unhandled rejection and a silently empty picker.
- The picker commits the canonical schema name rather than the raw combobox text, so whitespace-padded input can't store a target schema that doesn't resolve.
- Menu state simplified to a `computed` over a query ref; the now-dead `searchAndFetchMissingSchemas` store action removed; the unreachable "clear target schema" test dropped and coverage added for the above.

**Reject empty relation type / target schema on save** (`32b8e2e0`)
- A relation property could be saved with an empty `relation` (edge label), which later hard-crashed Neo4j projection when a Subject populated the relation, via `Cypher::escape('')` throwing. Now rejected at save time by the existing `SchemaContentValidator` (an `if/then` on relation-type definitions requiring non-whitespace `relation`/`targetSchema`), rather than corrupting on read.

Not changed here: the editor's Save is still gated only on `hasChanged`, so the per-field "required" errors remain display-only. The backend guard above removes the data-integrity danger; adding a frontend save-gating signal (to block an invalid save before the round-trip) would be a separate improvement.

> Fixes directed by @JeroenDeDauw, who picked which review findings to act on.
> Context: the NeoWiki codebase, the #953 diff, mutation-tested verification of the changed tests, and an end-to-end trace of the relation-projection crash.

malberts:
This changed the existing schema picker when creating a new subject. The new behaviour was unexpected to me, at least in part because I was used to the previous UI. When I start typing, it does not show me a list, and instead I have to click the dropdown. But the more jarring side-effect is the moment I type a matching name, it selects it automatically and jumps to the subject form:

[Screencast_20260702_003228.webm](https://github.com/user-attachments/assets/edb7e80b-04ec-4732-8daa-63e7fddb8c23)

The initial typed letters are also dropped when I type and then click the dropdown icon the first time:

[Screencast_20260702_005311.webm](https://github.com/user-attachments/assets/1004200e-9380-467c-ad6e-c114c81eb659)

It is not clear to me that these changes are intentional here, beyond the mention of the common component changing.

JeroenDeDauw:
The automatic selection for that flow is not intentional. 

Also the lack of results that show if you just type isn't desired. I wonder why none showed for you. Did you move to the field with keyboard navigation? Maybe the dropdown just opens when you click in the input?

malberts:
The field is already auto-focused when the dialog opens. And clicking in the field, not the dropdown button, I guess does not re-trigger the necessary event because it is already focused.

Issue #695, opened 2026-04-02
Layout: Handle orphaned display rules


JeroenDeDauw:
Follows-up to https://github.com/ProfessionalWiki/NeoWiki/pull/692

When a Schema property is renamed or deleted, Layout display rules referencing it become orphaned. Currently these are silently filtered out in the editor and dropped on save.

Options:
- Current behavior
- Show orphaned rules with a visual warning in the editor
- Auto-remove them on editor open with a notification
- Preserve them as-is and only warn

> Context: the NeoWiki codebase and architecture review of PR #692.
> <sub>Written by Claude Code, `Opus 4.6`</sub>

PR #954, opened 2026-06-27
Rework the Layout display-rule editor into a single show/hide list

JeroenDeDauw:
A Layout with no display rules previously showed every schema property under "Hidden · N" with eye-off icons while also stating "All properties shown in schema order". An empty allowlist means "show all", but the editor rendered the empty case as "nothing shown", so the editor contradicted the actual rendering and the first "Show" click left only a single property visible.

Replace the two-section Shown/Hidden editor with a single reorderable list of all schema properties:

- Each row has a visibility toggle. The eye / eye-off icon reflects the current state (matching Coda, Notion, Airtable); a tooltip names the action it performs.
- Hidden rows dim and sink to the bottom and are not draggable (only shown rows carry a drag handle).
- The header shows "N of M shown".
- "Show all properties" appears only when a property is hidden, and reveals the hidden properties while preserving the existing order.
- The last visible property's hide toggle is disabled (tooltip: "At least one property must be shown"), so the list can't collapse to zero shown and flip back to showing everything.

Domain semantics are unchanged: an empty display-rule list still means "show all properties in schema order". Display-rule editing logic is extracted into a pure, unit-tested module.

## Notes for reviewers

- **Show-none deferred:** the editor keeps at least one property shown (the last visible property's hide toggle is disabled). A true "show none" state is intentionally deferred — it's a non-breaking add later, since a "show all" layout is persisted as the *absence* of the displayRules key, never as `[]`.
- **Orphaned rules:** display rules referencing a property no longer in the schema are now pruned on the first edit (previously preserved). They already render nothing, so this only affects stored data. Fuller handling is tracked in #695.

> Result of extensive back-and-forth with @JeroenDeDauw — three prototype approaches were built on branches and reviewed in the UI; he chose this single-list approach and drove the refinements (non-draggable hidden rows, action tooltips, "Show all" only when something is hidden, and disabling the last property's hide toggle).
> Context: the NeoWiki codebase (LayoutEditor Vue components, display-rule resolution, Codex design tokens) and a Coda column-manager screenshot provided as a reference.


JeroenDeDauw:
Main driver is that originally the UI was very confusing when all properties where shown, as it looked like there where hidden. New:

[Screencast_20260627_234649.webm](https://github.com/user-attachments/assets/88510caa-18ac-4ddf-8ded-313262b69ee3)

PR #957, opened 2026-06-27
Render Select property attributes when adding a new property

### LINKED ISSUE #924: Select property attributes do not render when adding a new property
### Summary

In the Schema Editor, adding a property and setting its type to **Select** does not show the Select-specific attributes (the **Options** field and the **Initial value** field). Only the basic fields and the "Allow multiple values" checkbox appear. The Options field only shows up later, when reopening the schema for editing after it has been saved.

### Steps to reproduce

1. Go to `Special:Schemas` and click **Create** (or **Edit schema** on an existing schema).
2. Click **Add property definition** (the new property defaults to type Text).
3. Change **Type** to **Select**.
4. Observe the property editor.

### Expected

The Select attributes editor renders, including the **Options** chip input and the **Initial value** selector, the same as when editing an already-saved Select property.

### Actual

The **Options** and **Initial value** fields are missing; only Name, Type, Description, Require a value, and Allow multiple values are shown. The browser console logs two errors:

```
TypeError: Cannot read properties of undefined (reading 'map')
    at SelectAttributesEditor optionChips computed (renderSlot)
TypeError: Cannot read properties of undefined (reading 'map')
    at SelectType.validate (triggered by the Initial value SelectInput)
```

Saving the schema and reopening it for editing shows the Options field correctly.

### Root cause

A newly added property is created as `type: 'text'` with no `options`/`multiple` fields. Switching the type to `select` only changes `type`; it never initializes `options`, so it stays `undefined`. `SelectAttributesEditor` (the `optionChips` computed) and `SelectType.validate()` both call `property.options.map(...)` without a guard and throw, so those parts of the editor fail to render. On reload, the deserializer defaults `options` to `[]`, which is why editing a saved schema works. Every other property type's attribute editor already tolerates the missing-field state (optional chaining or undefined-tolerant helpers); Select is the only one that does not.

### Screenshots

| | |
|---|---|
| **New Select property: Options field missing (the bug).** A freshly added property switched to type Select; the Options chip input and Initial value selector never render. | Image|
| **Editing a saved Select property: Options field present.** The Company schema's `Status` property loaded from saved data; Options (Active/Inactive/Acquired/Dissolved) and Initial value render correctly. | Image|
| **New Text property: attributes render immediately.** For contrast, a newly added Text property shows Character length and Initial value right away. |Image|

JeroenDeDauw:
> Produced by the `/fix-issue` workflow, run by @JeroenDeDauw against #924 (which already contained a detailed root-cause analysis).
> Context: the NeoWiki codebase (Select property type and Schema Editor components) plus that issue. Verified with the vitest suite and a browser repro on the dev wiki.

Fixes https://github.com/ProfessionalWiki/NeoWiki/issues/924

A newly added property definition is a bare PropertyDefinition (type `text`,
no type-specific fields). Switching its type to Select via the dropdown only
changes `type`, leaving `options` undefined until the schema is saved and
reloaded (the deserializer defaults `options` to `[]`).

The Select code called `.map()`/`.find()`/`[]` on `options` without a guard,
so the `optionChips` computed in SelectAttributesEditor and `SelectType.validate()`
(triggered by the Initial value SelectInput) threw, and the Options and Initial
value fields failed to render. Every other property type already tolerates a
bare property by reading its optional fields defensively; Select now does the
same by treating a missing `options` as an empty list.

### DIFF
--- a/resources/ext.neowiki/src/components/SchemaEditor/Property/SelectAttributesEditor.vue
+++ b/resources/ext.neowiki/src/components/SchemaEditor/Property/SelectAttributesEditor.vue
@@ -42,7 +42,7 @@ const emit = defineEmits<AttributesEditorEmits<SelectProperty>>();
 const optionsError = ref<string | null>( null );
 
 const optionChips = computed( (): ChipInputItem[] =>
-	props.property.options.map( ( option ) => ( { value: option.label } ) )
+	( props.property.options ?? [] ).map( ( option ) => ( { value: option.label } ) )
 );
 
 const updateOptions = ( chips: ChipInputItem[] ): void => {
--- a/resources/ext.neowiki/src/components/SchemaEditor/Property/SelectAttributesEditor.vue
+++ b/resources/ext.neowiki/src/components/SchemaEditor/Property/SelectAttributesEditor.vue
@@ -42,7 +42,7 @@ const emit = defineEmits<AttributesEditorEmits<SelectProperty>>();
 const optionsError = ref<string | null>( null );
 
 const optionChips = computed( (): ChipInputItem[] =>
-	( props.property.options ?? [] ).map( ( option ) => ( { value: option.label } ) )
+	props.property.options.map( ( option ) => ( { value: option.label } ) )
 );
 
 const updateOptions = ( chips: ChipInputItem[] ): void => {
--- a/resources/ext.neowiki/src/components/SchemaEditor/PropertyDefinitionEditor.vue
+++ b/resources/ext.neowiki/src/components/SchemaEditor/PropertyDefinitionEditor.vue
@@ -17,8 +17,9 @@
 				{{ $i18n( 'neowiki-property-editor-type' ).text() }}
 			</template>
 			<CdxSelect
-				v-model:selected="localProperty.type"
+				:selected="localProperty.type"
 				:menu-items="typeOptions"
+				@update:selected="changePropertyType"
 			/>
 		</CdxField>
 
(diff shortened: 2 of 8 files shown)


JeroenDeDauw:
image

alistair3149:
Switched this from guarding every `property.options` read to fixing the root cause:

`PropertyDefinitionEditor` now reinitializes the type-specific fields when the type changes (a Select gets `options: []`), so the editor never emits a bare property. The read-site guards and their tolerance tests are dropped (Select source back to baseline) and replaced by an editor-level regression test. Mirrors how the backend and the factory/deserializer already guarantee the invariant at construction. See 8515ed0c.

What do you think? I verified in browser and either approach fixed the issue.

JeroenDeDauw:
AI suggested that when I looked at it and I wasn't convinced. Not objecting to it either. I think there is more `??` happening for other types that could be removed now.

I wonder if this is changing behavior when you change the property type. Did we retain fields from the other type earlier? What happens if you go from a number type with a bunch of attributes like min and max set, then switch to string, then switch back to number? We never defined the desired behavior for that and might be changing it here.

alistair3149:
> I think there is more ?? happening for other types that could be removed now.

No there isn't other types that can be removed. Select is special because `options` is required and is defaulted to `[]` in every construction path. Relation might have a case but it is due to a change anyways.

> I wonder if this is changing behavior when you change the property type. Did we retain fields from the other type earlier? What happens if you go from a number type with a bunch of attributes like min and max set, then switch to string, then switch back to number?

Yes. The old code kept the previous type's fields (and would even serialize them); new code resets type-specific fields + clears the default on each switch, so number→text→number no longer restores min/max. The old retain was what caused the bug. 

If the "restore" behavior is desirable on a product level, storing that separately would be cleaner.

JeroenDeDauw:
> The old retain was what caused the bug.

Don't think that's right. AFAIK, the issue was that attributes of the new type are missing, not that those of the old one were present.

alistair3149:
> > The old retain was what caused the bug.
> 
> Don't think that's right. AFAIK, the issue was that attributes of the new type are missing, not that those of the old one were present.

I should have framed it better. The retain behavior comes with the property type switch question that is not from the original issue ticket.

I went with the reset approach because it is cleaner in terms of data structure, when the data only contains constraints of the current property type (which would also save us the headache from the collision caused by property types that share the same constraint)

PR #972, opened 2026-06-30
Surface server violations on untouched single relations

alistair3149:
> Result of working through a validation parity audit with @alistair3149 — a precursor to removing NeoWiki's duplicated client-side validators in favour of server-driven validation. Context: the NeoWiki frontend validation flow and the create/edit dry-run model.

RelationInput gated its field message and status on the field being touched (focused then blurred). For a single (non-multiple) relation, a server-sourced violation — such as a `required` error from the save-time 422 — stayed hidden until the user focused the field, and was also excluded from the dialog's anchorless banner because its property name matches a rendered field. An empty required relation could pass review invisibly.

Server-sourced violations now surface regardless of touch; only the live client check still waits for blur. The creator is unaffected: its dry-run already strips `required`/`label-required`, so an untouched required relation produces no server violation to show.

## Manual Browser Check

Use a schema with a **required, single-value relation** property (e.g. a "Company" relation).

1. **Editor surfaces it:** open the Subject editor for an existing subject whose required relation is empty, without touching the relation field. Expect a red `required` error on the field immediately (previously silent).
2. **Save path:** attempt to save with the required relation still empty. Expect the inline error on the relation field, not only a generic toast.
3. **Creator stays quiet (regression guard):** create a new subject with that schema and do not touch the relation field. Expect no premature `required` error until you interact and blur, or attempt to save.
4. **Multiple mode unchanged:** a required multi-value relation still shows its message immediately, as before.

### DIFF
--- a/resources/ext.neowiki/src/components/Value/RelationInput.vue
+++ b/resources/ext.neowiki/src/components/Value/RelationInput.vue
@@ -80,11 +80,8 @@ function relevantServerViolations(): readonly SubjectViolation[] {
 	return ( props.serverViolations ?? [] ).filter( ( v ) => v.propertyName === name );
 }
 
-const fieldMessages = computed<ValidationMessages>( () => {
-	if ( Object.keys( liveFieldMessages.value ).length > 0 ) {
-		return liveFieldMessages.value;
-	}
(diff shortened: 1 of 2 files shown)


alistair3149:
https://github.com/user-attachments/assets/9a942be4-118c-4544-ab4c-a4107a5d41a7

PR #973, opened 2026-06-30
Remove client-side Property Type validation and add ADR 25

alistair3149:
Stack on #972 

Validated manually by going through the validation clean page, no regression found.

---

> Result of a validation parity audit and back-and-forth with @alistair3149, then removing the duplicated client-side validators per the agreed "server-only, including the extension API" direction. Context: the NeoWiki frontend validation flow, the property-type plugin system, and a live browser test against the Validation Demo.

Follows-up to https://github.com/ProfessionalWiki/NeoWiki/pull/972.

Validation is now server-driven: a debounced dry-run endpoint returns constraint violations as the user edits, and save-time enforcement returns them as a 422. The duplicated TypeScript validators mirrored the PHP ones and drifted from them, so this removes them entirely.

- The value inputs and the shared composables (`useFieldServerViolation`, `useStringValueInput`) surface server violations only; they no longer call `validate()`.
- Drop `validate()` from the Property Type classes and `BasePropertyType`, and remove the `ValueValidationError` type.
- Drop the `validate` hook from the frontend extension API (`PropertyTypeRegistration`, `PropertyTypeAdapter`). Custom Property Types validate through their registered PHP validator, the same as core types.
- Remove the unused client-side `SubjectValidator` service and the `useValueValidation` helper.

See ADR 25 (`docs/adr/025-backend-driven-frontend-validation.md`).

## Manual Browser Check

Use the `Validation Demo` schema: open `Validation Clean` → **Data** tab → **Edit** main subject. Confirm validation still surfaces, now server-driven (a ~300 ms debounce before errors appear):

1. **Field-level:** set `Summary` to fewer than 10 characters → "Minimum length is 10 characters." appears inline under the field.
2. **Per-index multi-value:** change one `Tags` entry to a non-http URL (e.g. `ftp://example.com/b`) and blur → only that row gets "Please enter a valid URL."; the sibling rows stay clean.
3. **Creator stays quiet:** **Add subject** → `Validation Demo`, leave required fields empty → no premature errors until you press Create/Save.
4. **Network:** the editor issues `POST .../subject/{id}/validate`; the frontend renders the returned violations and computes nothing locally.


JeroenDeDauw (changes requested):
There is some AI slop in here with made up reasoning. I will push some fixes to the branch

docs/adr/025-backend-driven-frontend-validation.md
```diff
@@ -0,0 +1,46 @@
...
+
+Pros:
+
+* A single canonical validator in PHP. No PHP/TypeScript sync burden, and no drift bugs.
+* Core and extension Property Types validate the same way, through the server.
+* Less frontend code to carry.
+
+Cons:
+
+* Validation feedback depends on the server round-trip, making the UIs less snappy. We also have to implement debouncing.
```
alistair3149: Debounce also exists in the TS implementation, I think we can just remove the sentence.

JeroenDeDauw (approved):

PR #976, opened 2026-06-30
Refresh a page's sourced data in the graph without a page edit

### LINKED ISSUE #889: Refresh a page's sourced data in the graph without a page edit
## Problem

A page's data is written to the graph on `RevisionFromEditComplete` / `RevisionUndeleted` and on full rebuild. When data contributed by an extension changes **outside** of an edit, the graph keeps the stale value until the page is next saved or a full rebuild runs.

The driving case is approval: an extension marks a revision approved (or changes approval state) without editing the page, and needs that reflected on the page's graph data so it stays queryable. This generalises to any `PagePropertyProvider`-contributed data.

## What's needed

A first-class, publicly-callable operation — **"refresh this page's sourced data"**. An extension reacting to its own state change calls it for a `Title`; NeoWiki re-derives the page's data from the registered providers (and the page's subject slot) and updates the graph. **No new revision is created.**

The behaviour already exists internally: `SubjectPageRebuilder::rebuild( Title )` re-runs the provider pipeline (`PagePropertiesBuilder::getPagePropertiesFor`, which iterates all registered `PagePropertyProvider`s) and re-saves the page node. The gap is purely access: `SubjectPageRebuilder` lives in `Application/` and must be hand-constructed, so callers would depend on its constructor.

## Approach

- Expose the operation on the **top-level factory** — the public surface established by #789 (which deliberately avoided the MediaWiki service container) — wrapping `SubjectPageRebuilder`. No service container, no hand-constructed internals.
- No REST endpoint needed; an extension can wrap it if it wants HTTP.
- `maintenance/RebuildGraphDatabases.php` can migrate to the same entry point for consistency.

## Acceptance criteria

- A public method on the top-level factory refreshes a given page's graph data on demand (by `Title`), **without** creating a MediaWiki revision.
- It re-derives the page's data from the registered `PagePropertyProvider`s and the page's subject slot, and updates the graph.
- The skip cases (no revision / no revision user / no `SubjectContent` slot) are handled and reported, not silently ignored (see below).
- `maintenance/RebuildGraphDatabases.php` uses the same entry point.
- Covered by tests, including the skip cases and a successful refresh.

## Open decision — result/error semantics

Internally `SubjectPageRebuilder::rebuild()` returns a bare `false` for the skip cases. That is fine for the internal helper, but the **public** method an extension calls should signal outcomes more explicitly than a boolean — e.g. a small result type distinguishing *refreshed* / *skipped (no revision)* / *skipped (no subject slot)* — so callers can react. Decide between a result value and exceptions; default to a clear return value over a bare boolean.

## Scope

Scoped to page-scoped provider data — the concrete, buildable case today. The broader "refresh any sourced data (incl. cached external pulls)" generalisation in the [Subject Sources doc](https://github.com/ProfessionalWiki/NeoWiki/blob/master/docs/planning/SubjectSources.md#refresh-without-an-edit) follows once the Source system exists; this issue is the on-demand trigger for what already works. Once #905 lands (wiki-scoped page-node identity), this refresh inherits that write path — no separate handling, just sequence the two so they don't conflict.

Replaces #782. Related: #831, #905.

> Result of the Subject Sources design work with @JeroenDeDauw; replaces #782.
> Context: the NeoWiki codebase (the `SubjectPageRebuilder` / `PagePropertyProvider` pipeline), the Subject Sources planning doc, and #789's outcome.

alistair3149:
Fixes https://github.com/ProfessionalWiki/NeoWiki/issues/889

Adds a public, no-edit operation to refresh a page's graph data on demand. An extension reacting to its own state change (the driving case is approval — an extension marks a revision approved without editing the page) calls:

```php
NeoWikiExtension::getInstance()->newSubjectPageRebuilder()->rebuild( $title );
```

NeoWiki re-derives the page's data from the registered `PagePropertyProvider`s and the page's subject slot and updates the graph, **without creating a MediaWiki revision**.

The operation returns a `PageRefreshOutcome` enum rather than a bare boolean, so callers can react to each outcome:

- `Refreshed`
- `SkippedMissingRevision`
- `SkippedMissingRevisionAuthor`
- `SkippedMissingSubjectSlot`

Expected skips are return values; genuine infrastructure failures (e.g. the graph store being unreachable) still throw.

## Changes

- New `PageRefreshOutcome` backed enum.
- `OnRevisionCreatedHandler` now reports whether it wrote a page node, replacing the previously silently-swallowed no-subject-slot case (and drops a vestigial argument in the undelete path).
- `SubjectPageRebuilder::rebuild()` maps the four states to the enum.
- `NeoWikiExtension::newSubjectPageRebuilder()` exposes the operation on the top-level factory established by #789 — no MediaWiki service container, no hand-constructed internals.
- `maintenance/RebuildGraphDatabases.php` uses the same entry point and reports skips by reason.

## Testing

- New `OnRevisionCreatedHandlerTest` pins the wrote-vs-skipped contract using a real graph-store spy.
- `SubjectPageRebuilderTest` covers all four outcomes.
- `make cs` clean; smoke-tested by running the rebuild maintenance script end to end (75/75 pages refreshed).

Replaces #782. #905 (wiki-scoped page-node identity) already landed, so this refresh inherits that write path unchanged.

> Result of a brainstorming + implementation session with @alistair3149, who settled the design decisions (an enum result type over a value object or exceptions, the `new*` facade accessor, and keeping the `SubjectPageRebuilder`/`rebuild()` names).
> Context: the NeoWiki codebase (the `SubjectPageRebuilder` / `PagePropertyProvider` pipeline), the Subject Sources planning doc, and issue #789's factory outcome.

### DIFF
--- a/docs/concepts/glossary.md
+++ b/docs/concepts/glossary.md
@@ -143,5 +143,9 @@ itself, as opposed to Subject Statements, which are structured data about the en
 Built-in Page Properties include `name`, `namespaceId`, `creationTime`, `lastUpdated`, `categories`, and `lastEditor`.
 Extensions can contribute additional Page Properties.
 
+Page Properties are written when the page is edited and during a full graph rebuild. When extension-contributed
+data changes outside an edit, an extension can refresh them on demand without creating a revision (see
+[Extending NeoWiki](../reference/extending.md)).
+
 When using Neo4j, Page Properties are available on every Page node and are queryable via Cypher
 (e.g., `MATCH (page:Page) WHERE page.lastUpdated > datetime("2024-01-01")`).
--- a/docs/reference/extending.md
+++ b/docs/reference/extending.md
@@ -107,7 +107,6 @@ the Page node. No new revision is created. `rebuild()` returns a `PageRefreshOut
 
 - `Refreshed` — the Page node was updated.
 - `SkippedMissingRevision` — the page has no current revision.
-- `SkippedMissingRevisionAuthor` — the current revision has no visible author.
 - `SkippedMissingSubjectSlot` — the page carries no NeoWiki subject slot, so there was nothing to write.
 
 Genuine failures (such as the graph store being unreachable) throw rather than returning an outcome.
--- a/docs/reference/extending.md
+++ b/docs/reference/extending.md
@@ -90,6 +90,28 @@ Register with `NeoWikiRegistrar::addPagePropertyProvider()`. The context exposes
 creation and modification times, categories, and last editor. Example:
 [`src/StaticPagePropertyProvider.php`](https://github.com/ProfessionalWiki/NeoWiki/blob/master/tests/RedHerb/src/StaticPagePropertyProvider.php).
 
+### Refreshing a page's data without an edit
+
+A page's graph data is written on edit and on full rebuild. When data your extension contributes through a
+`PagePropertyProvider` changes *outside* an edit — for example an approval extension marking a revision approved —
+the graph keeps the old value until the page is next saved. Trigger a refresh on demand:
+
+```php
+$outcome = NeoWikiExtension::getInstance()
+	->newSubjectPageRebuilder()
+	->rebuild( $title );
+```
+
+NeoWiki re-runs every registered `PagePropertyProvider` for the page (and re-reads its subject slot) and updates
+the Page node. No new revision is created. `rebuild()` returns a `PageRefreshOutcome`:
+
+- `Refreshed` — the Page node was updated.
+- `SkippedMissingRevision` — the page has no current revision.
+- `SkippedMissingRevisionAuthor` — the current revision has no visible author.
+- `SkippedMissingSubjectSlot` — the page carries no NeoWiki subject slot, so there was nothing to write.
+
+Genuine failures (such as the graph store being unreachable) throw rather than returning an outcome.
+
 ### Reading NeoWiki data and authorization
 
 `NeoWikiExtension::getInstance()` exposes read-side services usable from any MediaWiki extension point
--- a/maintenance/RebuildGraphDatabases.php
+++ b/maintenance/RebuildGraphDatabases.php
@@ -4,6 +4,7 @@ declare( strict_types = 1 );
 
 namespace ProfessionalWiki\NeoWiki\Maintenance;
 
+use Exception;
 use LogicException;
 use Maintenance;
 use MediaWiki\MediaWikiServices;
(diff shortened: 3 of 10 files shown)


JeroenDeDauw:
**Refresh can mislabel a genuine slot-read failure as `SkippedMissingSubjectSlot`**

Follow-up on the refresh contract. `OnRevisionCreatedHandler::getNeoContent()` (`src/EntryPoints/OnRevisionCreatedHandler.php:58-73`) catches *every* `RevisionAccessException` and returns `null`, which `rebuild()` maps to `PageRefreshOutcome::SkippedMissingSubjectSlot`. But in MW core that exception covers more than an absent slot — `RevisionSlots::getContent()` → `SlotRecord::getContent()` is documented to throw `BadRevisionException` ("the revision is permanently missing") and `RevisionAccessException` "for other storage access errors" (e.g. a blob-store read failure). Both are subclasses of `RevisionAccessException`, so both land in this catch.

That collides with the contract this PR documents in `docs/reference/extending.md`:

> - `SkippedMissingSubjectSlot` — the page carries no NeoWiki subject slot, so there was nothing to write.
>
> Genuine failures (such as the graph store being unreachable) throw rather than returning an outcome.

So a genuine read failure is returned as the benign "nothing to write" outcome instead of throwing.

Concrete impact:

- In `RebuildGraphDatabases`, pages are selected *because* their latest revision carries the slot (`getSubjectPageIds()`), so a transient slot/blob read failure prints `Skipped X: no subject slot` and drops that subject from the freshly-rebuilt graph — silently, with a message that says the opposite of what happened.
- A public-API caller (the approval case) gets a benign skip where the docs promise an exception.

The swallow itself predates this PR (the `// TODO: log this` catch), but this PR is what turns it into a public contract, so it seems like the right moment to reconsider.

Possible direction: distinguish "slot genuinely absent" from "slot present but unreadable" — e.g. gate on `$revisionRecord->hasSlot( MediaWikiSubjectRepository::SLOT_NAME )` and return `SkippedMissingSubjectSlot` only when the slot is truly absent, letting genuine read failures propagate per the "genuine failures throw" contract. Alternatively, if you'd rather keep the swallow, softening the doc wording would at least make the contract honest. Happy to implement whichever you prefer.

> Surfaced by an AI review Jeroen ran on this PR; he asked me to raise it for discussion rather than change the behavior unilaterally.
> Context: the NeoWiki codebase (the `SubjectPageRebuilder` / `OnRevisionCreatedHandler` pipeline) and MediaWiki core's `RevisionSlots` / `SlotRecord`.
> <sub>Written by Claude Code, `Opus 4.8 (1M context, max)`</sub>

PR #977, opened 2026-06-30
Degrade gracefully when a property type is not registered

### LINKED ISSUE #778: Unknown property type silently breaks the UI instead of degrading gracefully
Observed during verification of the frontend extension mechanism (#686 / #754).

## Problem

When a property-type-owning extension is disabled (or an extension registration fails) while stored Subject / Schema data still references that property type, the NeoWiki frontend aborts loudly in the console but the user sees silent UI failure.

## Reproduction

1. Create a schema with a property of a type owned by an extension (e.g. `Color` from a future RedHerb release).
2. Create a subject with a value for that property, save it.
3. Disable or unload the extension that owns the type (e.g. comment out `wfLoadExtension` in `LocalSettings.php`, restart MediaWiki).
4. Open the Schema page, a Layout page, or any Subject view that references the type.

## What happens

- `PropertyDefinitionDeserializer.propertyDefinitionFromJson` (`resources/ext.neowiki/src/domain/PropertyDefinition.ts:64`) calls `this.registry.getType( json.type )`.
- `PropertyTypeRegistry.getType` (`PropertyType.ts:63`) throws `Unknown property type: <typeName>`.
- On Schema view and Layout view, the throw escapes the `queueMicrotask` body in `neowiki.ts` (no `try/catch` around `SchemaDeserializer.deserialize` / `LayoutDeserializer.deserialize`), aborting deserialization before `app.mount()` runs. The `#ext-neowiki-view-schema` / `#ext-neowiki-view-layout` container is left empty.
- On Subject views and anywhere else, deserialization runs later inside Vue components / stores (via `SubjectDeserializer` → `StatementDeserializer` → `ValueDeserializer.deserialize()` which calls `getType` at `ValueDeserializer.ts:40`). The throw surfaces as an unhandled Vue reactivity error; the affected infobox / subject UI fails to render.
- **User-facing:** the NeoWiki UI container is empty. Other MediaWiki chrome (title, sidebar, navigation) still renders, so the page looks "mostly fine" to a non-technical reader.
- **HTTP:** stays 200; only the JS app fails.
- **Console:** loud — the exception is visible with a full stack trace. No user-facing error message.

## Why this matters

- Any property type that lives in an extension triggers this under disable / failure — core types are safe only because they can't disappear at runtime.
- "Looks fine but shows no data" is a worse failure mode than an inline error because users won't know to check the console.
- A single malformed property in a large schema takes the entire view down.

## Suggested direction

Graceful-degrade individual unknown types rather than aborting the whole view.

- In `PropertyDefinitionDeserializer.propertyDefinitionFromJson` and `ValueDeserializer.deserialize`, catch the `Unknown property type` throw and substitute a placeholder. Two sensible paths:
  - **Placeholder `PropertyType` implementation** registered automatically. The registry returns a stand-in for any unknown name; that stand-in satisfies `BasePropertyType` / `PropertyTypeRegistration` (see existing `presentation/PropertyTypeAdapter.ts` for the seam) and carries the original `typeName` as metadata.
  - **Sentinel objects returned from deserializers** rather than instances, requiring consumers to handle the "unknown" variant explicitly. More invasive, stricter type contract.
- Add an `UnknownValueDisplay.vue` / `UnknownValueInput.vue` pair that renders a placeholder row showing the raw stored value plus a muted "unknown type: `<typeName>`" label. Data remains visible and intact, just not interpreted.
- Emit a `mw.notify`-level warning (not an error banner) on first encounter per page so developers / admins notice without it being alarming for regular readers.
- Optionally disable editing for unknown-type rows to prevent accidental data loss through a generic input.

## Not blocking

Extensions owning property types aren't common yet (#754 only just lands the mechanism). Worth fixing before the first production extension ships a type — but not a merge-blocker for the initial mechanism PR.

## References

- Discovered in #754 "Follow-ups" section.
- Repro path: apply #777 (moves DateTime into RedHerb), then disable RedHerb in `LocalSettings.php` and open any page that has DateTime data.

alistair3149:
Fixes https://github.com/ProfessionalWiki/NeoWiki/issues/778

When stored Subject or Schema data references a property type whose owning
extension is disabled or failed to load, the frontend previously threw
"Unknown property type" and the entire view rendered blank, with no
user-facing indication of what went wrong.

Unregistered types now degrade gracefully instead of aborting the view:

- Deserialization produces an `UnregisteredTypeValue` that preserves the raw
  stored data, so values round-trip unchanged on save. A property definition
  of an unregistered type retains its type-specific keys for the same reason
  (pinned by a schema deserialize→serialize round-trip test, and by a
  statement-level round-trip test).
- The component registry falls back to read-only placeholder components for
  display, editing, and attributes, so a single unregistered type no longer
  takes down the whole view.
- The raw value stays visible alongside a muted "type not available" note,
  and a one-time warning notification surfaces the problem to editors
  without alarming readers.
- In the schema editor, the property's type shows as a disabled "Unavailable
  type" entry in the type selector instead of an empty select, and the
  attributes panel notes that stored settings are preserved.

This is the frontend half. The backend half — the server no longer throwing
when a Subject read/render/write references a type it cannot resolve — landed
separately in #1020 (issue #1017) and is now on master; this branch is rebased
on top of it. With both halves in place the Subject-side degradation is
reachable end to end, and the two halves share one vocabulary: the PHP and
TypeScript value types are both named `UnregisteredTypeValue`, and the
`ValueType::UnregisteredType` / `ValueType.UnregisteredType` enum now carries
the same string value on both sides.

## Manual Browser Check

Reproduces the issue scenario (data referencing a type the frontend does not
know). On a wiki with NeoWiki installed:

1. Create a Schema page (`Schema:UnknownTypeTest`) whose JSON references an
   unregistered property type, e.g.:
   ```json
   {
     "description": "Schema exercising an unregistered property type",
     "propertyDefinitions": {
       "Name": { "type": "text" },
       "Swatch": { "type": "zzz-color", "default": "#ff5733", "palette": "warm" }
     }
   }
   ```
2. Open the Schema page. **Expected:** the schema table renders fully (not
   blank). The `Swatch` row shows an "Unavailable type" chip; its initial
   value `#ff5733` is shown with a muted `The type "zzz-color" is not
   available on this wiki.` note beneath it. No uncaught console error.
3. Open the schema editor and select `Swatch`. **Expected:** the type select
   shows a disabled "Unavailable type" entry (other types remain selectable),
   the attributes panel says stored settings are preserved, and the
   initial-value field is read-only with the "not available" note.

Before this change, step 2 left the view container empty with only a console
exception.

<sub>Directed by `@alistair3149` (scoped the approach, ran the review, set up the worktree). Context: the NeoWiki codebase and issue #778, with an adversarial multi-agent review and an in-browser reproduction. Written by Claude Code, `Opus 4.8 (1M context)`.</sub>

<sub>Follow-up commits (rename to `UnregisteredTypeValue`, string unification, type-selector fallback, note dedup, round-trip test) directed by `@JeroenDeDauw` after an in-session review discussed over several rounds; verified via vitest/build/lint and in-browser on a worktree stack. Written by Claude Code, `Fable 5 (max)`.</sub>

<sub>Rebased onto master after the backend half (#1020) merged; aligned the `ValueType.UnregisteredType` enum string with the backend and added a statement-level round-trip test. Directed by `@alistair3149` following a code review. Written by Claude Code, `Opus 4.8 (1M context)`.</sub>


alistair3149:
image
image

malberts:
This only fixes the frontend. I'll log a separate task for the backend.

malberts:
I am not sure about the "Unknown Value" naming here. I know https://github.com/ProfessionalWiki/NeoWiki/issues/778 used that wording, but it overlaps with the concept of "this statement has a value, but it is not known", as per https://github.com/ProfessionalWiki/NeoWiki/issues/937, and has a similar meaning in Wikibase.

alistair3149:
> I am not sure about the "Unknown..." naming here. I know #778 used that wording, but it overlaps with the concept of "this statement has a value, but it is not known", as per #937, and has a similar meaning in Wikibase.

Perhaps we can use `unregistered`?

malberts:
> Perhaps we can use unregistered?

That, or "unsupported".

"unregistered" is more excplicit that it is not registered, whereas "unsupported" doesn't imply anything about how to support it. "unsupported" seems to come from existing code, though.

A possibly YAGNI question: could there be a case where a type is registered, but somehow not supported? And how much does such a distinction matter? Best example I can come up with right now is if you define some very weird type that relies on another service being available in the wiki (let's say Qdrant), and if you use that on a different wiki the type might be registered successfully, but won't run correctly, and therefore becomes unsupported.

Ideally we should use one word consistently. Currently there are strings referring to "unknown", "unregistered" and "unsupported".

malberts:
It would be nice to also have a generic label, and perhaps a warning icon, in the dropdown here:
image

Although, there are enough other indicators on the form that something is wrong with the type. It just looks slightly broken if the dropdown is empty when something is actually selected.

Can we avoid the duplicate "Unsupported type: xxx"?

JeroenDeDauw:
This is inconsistent: `unknown type` and `unsupported type`
image

JeroenDeDauw:
High-level review — the direction is right (preserve-and-round-trip beats a blank view, and matches the ADR 12/21 intent of not losing stored data). Mergeable once the naming is settled. Key points for continuing:

1. **Naming**: it's the *type* that is unregistered, not the value, so: `UnknownValue` / `ValueType.Unknown` → `UnregisteredTypeValue` / `ValueType.UnregisteredType` (and `newUnregisteredTypeValue()`, the component names, `setUnregisteredTypeFallback()`). This also keeps "unknown value" free for #937's somevalue-style semantics. Worth doing pre-merge: these are exported via `public-api.ts`, so renaming later is a breaking change for extension authors. For the UI strings, keep "unregistered" out of user-facing copy and use one consistent phrasing across all three messages, e.g. "The type "$1" is not available on this wiki" — that also resolves the current "unknown type" / "unsupported type" mix.

2. **Reach, and how #1017 should be scoped**: this PR fully fixes Schema and Layout pages (they deserialize page content client-side) and the "type registered in PHP but missing in JS" case. REST-fed surfaces (subject editor/creator, infobox) don't benefit yet: subject reads 500 (#1017), and `GetSchemaApi` serves the schema through `SchemaPersistenceDeserializer`, which *silently drops* unregistered-type properties — so the placeholder machinery never gets fed there. #1017 should therefore converge on the same preserve-and-surface semantics: replace the schema-side silent drop too, and on the write path accept unregistered-type statements with a warning-severity violation (ADR 26) rather than rejecting, per ADR 12/21.

3. **Test to add**: a schema deserialize→serialize round-trip test asserting the retained type-specific keys survive a re-save. The invariant currently rests on `SchemaSerializer` happening to spread all object keys; a later cleanup to an explicit field list would silently reintroduce data loss without failing the existing deserializer-only test.

4. **PR description accuracy**: manual-check step 3 isn't executable as written — the new input is read-only, and POSTing such a value is rejected server-side (#1017) — so reword it to the "subject already had a value for the type" case. The "Subject validation skips types it cannot interpret" bullet describes pre-existing backend behavior, not something in this diff.

> Review requested by @JeroenDeDauw as a goal/approach check; findings discussed with him over two rounds (he redirected the naming and prompted the frontend-vs-backend scoping question); this summary posted on his request without him reviewing its exact wording. Static review, nothing executed: full diff, #778/#1017 and the PR thread, backend paths traced on master (validator, schema/statement deserializers, `GetSchemaApi`), serializer key-spread verified on the branch.

JeroenDeDauw:
Status: the review points above are now applied in 721dc07b (pushed directly, CI green), so this is ready for a look at that commit and then merge.

- **Naming settled as `UnregisteredTypeValue`** / `ValueType.UnregisteredType` (+ derived names): it is the *type* that is unregistered, not the value, and "unknown value" stays free for #937's has-a-value-but-unknown semantics. Renaming pre-merge matters because these names are exported via `public-api.ts`.
- **User-facing strings unified** on one phrasing: chip "Unavailable type", note `The type "$1" is not available on this wiki.` — no more unknown/unsupported mix, and "unregistered" stays out of user copy.
- **@malberts's two UX points fixed**: the type selector now shows a disabled "Unavailable type" entry (with alert icon) instead of rendering empty, and the attributes panel got its own "settings are preserved" note instead of duplicating the type-name note.
- **Round-trip test added** pinning that an unregistered property's type-specific keys and raw default survive a schema re-save.
- **PR description updated** to match, including an honest note that the Subject disable-extension case is blocked on the backend (#1017) — that half is in progress as draft #1020.

Verified: full vitest/build/lint, plus in-browser on a worktree stack (schema page renders, editor selector + notes behave as described).

> Status summary requested by @JeroenDeDauw; the underlying changes were discussed with him over several rounds in-session, this comment's wording was not reviewed by him.

Issue #895, opened 2026-06-16
Finish the graph database plugin system


JeroenDeDauw:
Finish the graph database plugin system so new backends are purely additive

---

> Result of a back-and-forth with @JeroenDeDauw reviewing ADR 019's plugin-system gaps against the current code, and deciding to split the plugin-system completion out from the SPARQL work.
> Context: the NeoWiki codebase (`src/GraphDatabasePlugins/Neo4j/`, `GraphDatabasePlugin`, `CompositeGraphDatabasePlugin`, `NeoWikiExtension`), [ADR 019](https://github.com/ProfessionalWiki/NeoWiki/blob/master/docs/adr/019-graph-database-architecture.md), and the merged refactors #817 / #840.
> <sub>Written by Claude Code, `Opus 4.8`</sub>

## Goal

Complete the graph-database plugin system so that adding a new backend (starting with the SPARQL/QLever plugin, #586) is **purely additive**: the implementer writes a new plugin and registers it, without first having to refactor core.

Per ADR 019, the shared contract is **at the domain-event level, not the query level**. The Neo4j query engines stay Neo4j-specific, and SPARQL will bring its own `{{#sparql}}` / `/query/sparql` surface. **Designing a cross-backend query/validation abstraction is explicitly NOT part of this work** — that is best left to emerge once a second backend exists.

## Scope

- [x] **Formalize the plugin contract at the event level.** Move `GraphDatabasePlugin` out of `src/Persistence/` to a plugin-contract location. (#897)
  - Landed as `savePage` + `deletePage` in `src/Domain/GraphDatabase/`. **No move event was added**, deliberately: core creates a null revision on every move, which fires `RevisionFromEditComplete` → `savePage`, and the page node is keyed by the move-stable page ID. #897 removed the no-op move hook instead of filling it in; `tests/phpunit/EntryPoints/PageMoveGraphProjectionTest.php` regression-tests real `MovePage` calls against the graph.
- [x] **Introduce a `Neo4jPlugin` composition root** that wires the Neo4j plugin and contributes its own user-facing surfaces (REST route, parser function, Lua library). (#978)
- [x] **Register surfaces per-plugin, only when configured.** A wiki with no Neo4j plugin must not expose `POST /neowiki/v0/query/cypher`, `{{#cypher_raw}}`, or `nw.query`. *(#1016, subsumes #860.)* All three surfaces are tested in both directions.
[...]

PR #840, opened 2026-05-19
Move Neo4j-bound code into src/GraphDatabasePlugins/Neo4j/

JeroenDeDauw:
Follows-up to https://github.com/ProfessionalWiki/NeoWiki/pull/817

## Summary

Group everything Neo4j-specific (sync code, raw query surface, REST endpoint, parser function, Lua adapter, validators, normalizer, exceptions) under one plugin directory. Internal layout mirrors the main repo (`Application/`, `Persistence/`, `EntryPoints/`) so readers don't have to learn a new convention to navigate the plugin.

Targets the #817 branch rather than `master` so the discussion stays scoped to the layout change; the underlying feature work stays in #817.

## Rationale

Per [ADR 019](https://github.com/ProfessionalWiki/NeoWiki/blob/feat/public-query-api/docs/adr/019-graph-database-architecture.md) each backend is a plugin; per [ADR 001](https://github.com/ProfessionalWiki/NeoWiki/blob/feat/public-query-api/docs/adr/001_Domain_Centric_Architecture.md) clean architecture is a means, not an end. The classes moved here have only one implementation and only Neo4j-bound consumers, so the prior split across `src/Application/Query/`, `src/Application/Cypher*`, `src/Persistence/Neo4j/`, and `src/EntryPoints/*Cypher*` added layer boundaries that didn't pay for themselves and obscured that all of it belongs to one plugin.

## Layout

```
src/GraphDatabasePlugins/Neo4j/
├── Application/
│   ├── Neo4jQueryService.php
│   ├── Neo4jQueryRequest.php
│   ├── Neo4jQueryResult.php
│   ├── Neo4jQueryLimits.php
│   ├── CypherQueryValidator.php          (interface)
│   ├── CompositeCypherQueryValidator.php
│   ├── KeywordCypherQueryValidator.php
│   ├── ExplainCypherQueryValidator.php
│   └── Exception/ {8 exception classes, unchanged names}
├── Persistence/
│   ├── Neo4jQueryStore.php               (unchanged)
│   ├── Neo4jQueryEngine.php              (was QueryEngine.php)
│   ├── Neo4jWriteQueryEngine.php         (was WriteQueryEngine.php)
│   ├── Neo4jSubjectUpdater.php           (was SubjectUpdater.php)
│   ├── Neo4jSubjectUpdaterFactory.php    (was SubjectUpdaterFactory.php)
│   ├── Neo4jSubjectRelationUpdater.php   (was SubjectRelationUpdater.php)
[...]

Issue #586, opened 2026-02-13
Support QLever


JeroenDeDauw:

krabina:
Here is the link: https://drive.google.com/file/d/1O1aWVvboPtZx7bCuXykbgDtka2s7-5z-/view?usp=drive_link

krabina:
Instead of Oxigraph, a potential next candidate is Virtuoso because of market share.

JeroenDeDauw:
The main work is to [define the RDF mappings and implement the mapping system](https://github.com/ProfessionalWiki/NeoWiki/issues/723). Beyond that, we'll have to create a triple/RDF store plugin (easy), and that one could support multiple such databases (likely trivial). So I expect us to not have to choose between these databases and to easily achieve support for multiple once the RDF work is out of the way.

lozanaross:
@luguenth from TIB team will also work on this task, not as sole assignee, but should be involved in the decision-making & dev work here.

JeroenDeDauw:
Important context: https://github.com/ProfessionalWiki/NeoWiki/blob/master/docs/adr/019-graph-database-architecture.md

The final form of QLever support will be affected by the RDF mapping, which still has a lot of product work remaining before we can implement. That said, we could already add the connection to QLever and write stub data to it, or some simple mapping, as long as it's clearly marked that this is a disposable, temporary, and incomplete mapping.

JeroenDeDauw:
We've realized the graph database plugin system isn't entirely done yet. That needs to happen before we can create the stub QLever plugin. Tracked via:
* https://github.com/ProfessionalWiki/NeoWiki/issues/895

JeroenDeDauw:
This is an ECHOLOT requirement: [FR1.1 — support Linked Open Data standards (e.g., RDF, SPARQL)](https://github.com/ECHOLOT-ECCCH/echolot-WP2/issues/13).

JeroenDeDauw:
Status update: the blockers this issue sat behind have cleared, and implementation has started.

**What changed since June:**

- #895's write-path contract is done on master: backends are purely additive (plugin registry, per-plugin failure isolation, rebuild fan-out). The query-surface seams were deliberately left for this issue to shape.
[...]

PR #978, opened 2026-06-30
Graph-database plugin registry and Neo4j composition root

alistair3149:
> Result of an extended brainstorm → design → implement → review → smoke-test session driven by @alistair3149, who scoped this to scope 2 and 5 of #895, chose the registry pattern and the `Domain/GraphDatabase` home, and made the key calls (strengthening the additivity test to an end-to-end assertion, then holding before opening this PR).
> Context: the NeoWiki codebase, [ADR 019](https://github.com/ProfessionalWiki/NeoWiki/blob/master/docs/adr/019-graph-database-architecture.md), #895, and the merged refactors #817 / #840.

For https://github.com/ProfessionalWiki/NeoWiki/issues/895 (scope 2 and 5)

Make Neo4j the first **registered** plugin behind a composition root, so adding a backend (starting with SPARQL/QLever, #586) becomes purely additive. Behavior-preserving in production — only Neo4j is seeded.

## What this does

- Relocate the `GraphDatabasePlugin` contract and `CompositeGraphDatabasePlugin` out of `Persistence/` into `Domain/GraphDatabase/`, consistent with the other registries.
- Add a `GraphDatabasePluginRegistry` (mirrors `PagePropertyProviderRegistry`) and `NeoWikiRegistrar::addGraphDatabasePlugin()`, satisfying ADR 019's "extensions can register additional graph database plugins".
- Extract a `Neo4jPlugin` composition root that owns Neo4j construction; the composite is built from the registry, seeded with Neo4j by core, replacing the hardcoded `new CompositeGraphDatabasePlugin( $this->getNeo4jPlugin() )`.
- RedHerb registers an example plugin, and an integration test proves an extension-registered plugin receives `savePage`/`deletePage` end-to-end.

## Verification

- `make cs` green; the relevant unit and integration tests pass (`CompositeGraphDatabasePlugin`, `Neo4jQueryStore`, `NeoWikiRegistrar`, `NeoWikiRegistrationHook`, `PageMoveGraphProjection`, `Neo4jQueryService`, and the new `ExtensionGraphDatabasePluginDispatch`).
- Smoke-tested on a live wiki: the Cypher REST query path works against Neo4j, and `RebuildGraphDatabases` re-projected 75/75 pages through the new composite with the node count unchanged.


JeroenDeDauw:
Ran a multi-agent review over this and opened #987 into this branch to address its main finding.

Seeding core's Neo4j plugin into the extension-writable registry makes `getGraphDatabasePluginRegistry()` reentrant (building Neo4j → `getSchemaLookup` → `getPropertyTypeRegistry` → `ensureExtensionsRegistered`). Consequences:

- On a Cypher/write-first request (`getNeo4jPlugin()` is the first accessor) the `if ( $this->neo4jPlugin === null )` guard is defeated and the plugin is built twice, so the save/delete composite and the query/write path end up holding **different** `Neo4jQueryStore` instances. Benign only because the store is readonly/stateless today.
- `ensureExtensionsRegistered()` (fired on most subject/schema renders + validations) now eagerly builds both Neo4j clients where it built none before; since `withDriver()` validates the URL scheme eagerly, a malformed neo4j URL now breaks ordinary renders, not just graph ops.
- Plugin order is nondeterministic — the "do not rely on order" comment.

#987 composes core straight into the composite and keeps the registry extension-only, which removes all three and restores atomic memoization. Full suite green, plus a live Cypher smoke test.

One finding I deliberately left out of #987: the composite fan-out has no per-plugin error isolation — a throwing extension plugin aborts the loop and can skip core's projection. That felt like an isolate-vs-fail-fast decision better made explicitly than bundled into this fix.

> Result of a multi-agent code review of this PR that I directed. Context: the NeoWiki codebase, ADR 019, #895, and this PR's diff.
> <sub>Written by Claude Code, `Opus 4.8 (1M context, max)`</sub>

PR #982, opened 2026-07-03
Provide revision content to PagePropertyProviderContext

### LINKED ISSUE #979: Provide revision content to `PagePropertyProviderContext`
The current implementation already accesses the revisions content to retrieve the  categories, but it does add it to the context.

* https://github.com/ProfessionalWiki/NeoWiki/blob/d1f7341d85026f5c7a39dc2acb08122179b8e24b/src/PagePropertiesBuilder.php#L41-L53
* https://github.com/ProfessionalWiki/NeoWiki/blob/d1f7341d85026f5c7a39dc2acb08122179b8e24b/src/PagePropertiesBuilder.php#L78-L88

A `PagePropertyProvider` that wants to analyze the content would need to perform additional steps to retrieve it from the context.

I believe it would be useful for future extension developers to provide the actual content of the revision to `PagePropertyProvider` implementations via the context. It would ease generating page properties based on other things in the content ("@-mentions", "task list items", ...).

JeroenDeDauw:
> Result of back and forth with @JeroenDeDauw on the design direction.
> Context: the NeoWiki codebase and the discussion on the issue.

Fixes https://github.com/ProfessionalWiki/NeoWiki/issues/979

`PagePropertiesBuilder` already loaded the revision's main slot content and parsed it to extract categories, but threw everything else away. Providers wanting to derive Page Properties from the content had to re-fetch and re-parse it themselves - and since the context carries no revision ID, they could only look up the latest revision of the page, which is redundant work and subtly racy.

The context now additionally exposes:

* `content` - the serialized main slot content, e.g. the wikitext
* `contentModel` - e.g. "wikitext"
* `parserProperties` - the MediaWiki page properties recorded while parsing the content (e.g. "defaultsort", or values set by parser hooks via `ParserOutput::setPageProperty`)

Design notes:

* The context is a Domain value object holding only scalars and arrays. Exposing `RevisionRecord`, `Content`, or `ParserOutput` directly would put MediaWiki core classes into the Domain layer and make provider tests heavyweight. The relevant data is flattened instead, following the approach already used for `categories`.
* `parserProperties` covers the template-expansion-safe path: extensions whose parser hooks already record values during the parse can read them back in their provider instead of re-analyzing raw markup. The parse NeoWiki already performs for categories is reused, so no additional parse happens.
* When the content is unavailable (e.g. suppressed), `content` and `contentModel` are empty strings and `parserProperties` is empty, matching the existing behavior of `categories`.
* `contentModel` became load-bearing while this PR was open: since https://github.com/ProfessionalWiki/NeoWiki/pull/1226 every page is projected, so main slots that are not wikitext (JSON, Lua, CSS, …) now reach providers too, and the raw `content` string is only interpretable together with its model.

Note: content-derived properties refresh only when a new revision is stored, so values coming from transcluded templates can go stale - the same existing limitation as categories.

### DIFF
--- a/src/Domain/Page/PagePropertyProviderContext.php
+++ b/src/Domain/Page/PagePropertyProviderContext.php
@@ -13,6 +13,14 @@ readonly class PagePropertyProviderContext {
 	 * @param string $modificationTime In the standard MediaWiki format, ie 20230726163439
 	 * @param string[] $categories
 	 * @param string $lastEditor Plain username of the last editor, e.g. "JohnDoe". Empty string if unknown.
+	 * @param string $content Serialized main slot content of the revision, e.g. the wikitext.
+	 *   Empty string if the content is unavailable.
+	 * @param string $contentModel Content model of the main slot, e.g. "wikitext".
+	 *   Empty string if the content is unavailable.
+	 * @param array<string, int|float|string|bool|null> $parserProperties Properties recorded during parsing of the
+	 *   content (MediaWiki page properties, e.g. "defaultsort" or values set by parser hooks via
+	 *   ParserOutput::setPageProperty). These are inputs from the parse, not to be confused with the NeoWiki
+	 *   Page Properties that providers return.
 	 */
 	public function __construct(
 		public PageId $pageId,
@@ -22,6 +30,9 @@ readonly class PagePropertyProviderContext {
 		public string $modificationTime,
 		public array $categories,
 		public string $lastEditor,
+		public string $content,
+		public string $contentModel,
+		public array $parserProperties,
 	) {
 	}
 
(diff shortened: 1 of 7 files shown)


src/Domain/Page/PagePropertyProviderContext.php
```diff
@@ -22,6 +30,9 @@ public function __construct(
 		public string $modificationTime,
 		public array $categories,
 		public string $lastEditor,
+		public string $content,
+		public string $contentModel,
+		public array $parserProperties,
```
JeroenDeDauw, line 35: @osnard unsure if this is sufficient for your use cases and which of those you actually need? Perhaps just parserProperties?

osnard: This seems alright. Thanks!

While I agree that "scraping raw wikitext is fragile", I'd still prefer to at least have the chance to access it. Is there a special reason you pass it as a string rather than as `Content` object reference?

JeroenDeDauw: Yes, we try to keep MediaWiki objects out of the domain layer

Issue #101, opened 2024-09-25
Verify that REST APIs check permissions


JeroenDeDauw:
And fix issues if present

If there are no accurate checks yet and those are non-trivial to implement, then such accurate checks can be skipped and tracked by https://github.com/ProfessionalWiki/NeoExtension/issues/87 by expanding its scope to include backend.

alistair3149:
> Result of an audit directed by @alistair3149 (scope and fix approach decided in review).
> Context: the NeoWiki codebase, MediaWiki REST/permissions internals, and an adversarial multi-agent review of all 17 REST endpoints.
> <sub>Written by Claude Code, `Opus 4.8 (1M context)`</sub>

**Audit summary**

Traced all 17 REST endpoints (handler → action/query → repository/lookup → MediaWiki core) and adversarially verified every gap.

**Baseline.** MediaWiki's REST framework auto-enforces only the wiki-**global** `read` right; it does no per-title checks, and in this version its write gate is a no-op. Per-title `read`/`edit` is always the handler's job — core's own REST handlers do it explicitly. NeoWiki's checks were coarse (global-only) in a few places.

**Correctly protected**
- All 6 Subject writes: CSRF + global `edit` + (now, via #986) per-title `edit`.
- Schema reads (`/schema/{name}`, `/schemas`): per-title `read` via `CachingSchemaLookup`.
- `/query/cypher`: dedicated `neowiki-query` right + rate limit + read-only rejection.
- RevDel/suppressed revision content is honored wherever Subjects/Schemas/Layouts are read.

**Confirmed gaps**

| Gap | What | Severity | Status |
|---|---|---|---|
| **A** — per-title `edit` not checked on writes | 6 write endpoints | High — page-protection bypass (also bypassed user blocks) | **Fixed** in #986 |
| **B** — per-title `read` not enforced on Subject/Layout reads | `GET /subject/{id}` (incl. `?revisionId`), `GET /page/{id}/subjects`, `GET /layout/{name}`, `POST /subject/{id}/validate`, and the `GET /subject-labels` search | High **only on read-restricted wikis** (private / Lockdown / restricted namespaces); no impact on a default open wiki | Deferred |

[...]

PR #986, opened 2026-07-03
Enforce per-page edit permission on Subject write endpoints

alistair3149:
For https://github.com/ProfessionalWiki/NeoWiki/issues/101

The six Subject write REST endpoints (create main/child Subject, replace Subject, delete Subject, set main Subject, set Subject ordering) authorized writes with the wiki-global `edit` right (`Authority::isAllowed('edit')`) only. Subjects live in a page's revision slot, so any user holding the global edit right could create, replace, delete, or reorder Subjects on an edit-protected page or in a restricted namespace via the REST API, bypassing page protection. Blocked users were unaffected by the check too, since neither the REST framework's write gate nor `isAllowed('edit')` consults blocks.

`SubjectAuthorizer` now takes the target page and `AuthorityBasedSubjectAuthorizer` checks `Authority::definitelyCan('edit', $title)`, which enforces page protection, restricted namespaces, and user blocks. Create / set-main / set-ordering pass the request's page id; delete / replace resolve the Subject's owning page via the existing `PageIdentifiersLookup`. When the page cannot be resolved, authorization falls back to the global `edit` right instead of skipping the check, a fail-closed floor that cannot bypass protection because an unresolvable Subject is a no-op write.

The page-tools and RedHerb sidebar "create/edit subject" hints now reflect per-page edit rights.

Per-title read gaps and cross-page enumeration leaks surfaced by the same audit are out of scope for this change.

## Manual Browser Check

1. Create a page and protect it for editing (edit = sysop).
2. Log in as a non-sysop user that holds the `edit` right, and open one of that wiki's pages so `mw` is loaded.
3. From the browser console, POST to `<wgScriptPath>/rest.php/neowiki/v0/page/<protectedPageId>/mainSubject` with an `X-CSRF-TOKEN` header (`await new mw.Api().getToken('csrf')`) and body `{"label":"x","schema":"Employee","statements":{}}`.
   → **403** "You do not have the necessary permissions to create this subject".
4. Repeat against an unprotected page → **201**.

> Directed by @alistair3149; the fix approach and scope were decided in review with them.
> Context: the NeoWiki codebase, MediaWiki REST/permissions internals, and an adversarial multi-agent audit of all 17 Subject/Schema/Layout/query REST endpoints.

### DIFF
--- a/src/Application/Actions/CreateSubject/CreateSubjectAction.php
+++ b/src/Application/Actions/CreateSubject/CreateSubjectAction.php
@@ -39,14 +39,19 @@ readonly class CreateSubjectAction {
 			throw new InvalidArgumentException( 'SubjectLabel cannot be empty' );
 		}
 
-		if ( ( $request->isMainSubject && !$this->subjectAuthorizer->canCreateMainSubject(
-				) ) || !$this->subjectAuthorizer->canCreateChildSubject() ) {
+		$pageId = new PageId( $request->pageId );
+
+		$authorized = $request->isMainSubject
+			? $this->subjectAuthorizer->canCreateMainSubject( $pageId )
+			: $this->subjectAuthorizer->canCreateChildSubject( $pageId );
+
+		if ( !$authorized ) {
 			throw new RuntimeException( 'You do not have the necessary permissions to create this subject' );
 		}
 
 		$subject = $this->buildSubject( $request );
 
-		$pageSubjects = $this->subjectRepository->getSubjectsByPageId( new PageId( $request->pageId ) );
+		$pageSubjects = $this->subjectRepository->getSubjectsByPageId( $pageId );
 
 		try {
 			if ( $request->isMainSubject ) {
@@ -66,7 +71,7 @@ readonly class CreateSubjectAction {
 			return;
 		}
 
-		$this->subjectRepository->savePageSubjects( $pageSubjects, new PageId( $request->pageId ), $request->comment );
+		$this->subjectRepository->savePageSubjects( $pageSubjects, $pageId, $request->comment );
 		$this->presenter->presentCreated( $subject->id->text, $violations );
 	}
 
--- a/src/Application/Actions/DeleteSubject/DeleteSubjectAction.php
+++ b/src/Application/Actions/DeleteSubject/DeleteSubjectAction.php
@@ -4,21 +4,29 @@ declare( strict_types = 1 );
 
 namespace ProfessionalWiki\NeoWiki\Application\Actions\DeleteSubject;
 
+use ProfessionalWiki\NeoWiki\Application\PageIdentifiersLookup;
+use ProfessionalWiki\NeoWiki\Application\SubjectAuthorizer;
 use ProfessionalWiki\NeoWiki\Application\SubjectRepository;
 use ProfessionalWiki\NeoWiki\Domain\Subject\SubjectId;
-use ProfessionalWiki\NeoWiki\Application\SubjectAuthorizer;
+use RuntimeException;
 
 readonly class DeleteSubjectAction {
 
 	public function __construct(
 		private SubjectRepository $subjectRepository,
-		private SubjectAuthorizer $subjectAuthorizer
+		private SubjectAuthorizer $subjectAuthorizer,
+		private PageIdentifiersLookup $pageIdentifiersLookup
 	) {
 	}
 
 	public function deleteSubject( SubjectId $subjectId, ?string $comment ): void {
-		if ( !$this->subjectAuthorizer->canDeleteSubject() ) {
-			throw new \RuntimeException( 'You do not have the necessary permissions to delete this subject' );
+		// A null pageId (unresolvable Subject) makes the authorizer fall back to the global 'edit' right.
+		// This cannot bypass page protection: the repository resolves the page via the same lookup, so an
+		// unresolvable Subject results in a no-op delete rather than a write to a protected page.
+		$pageId = $this->pageIdentifiersLookup->getPageIdOfSubject( $subjectId )?->getId();
+
+		if ( !$this->subjectAuthorizer->canDeleteSubject( $pageId ) ) {
+			throw new RuntimeException( 'You do not have the necessary permissions to delete this subject' );
 		}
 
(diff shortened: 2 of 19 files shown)


JeroenDeDauw:
## Review: looks good — no blocking issues, a few hardening suggestions

I reviewed this across correctness, security, design, tests, and scope (multi-agent pass + manual tracing + a live end-to-end test on a dev wiki). **The fix is sound and I recommend it.** The reported hole is genuinely closed: for any Subject that resolves to a real page, all six write endpoints now gate on `Authority::definitelyCan( 'edit', $title )`, which uses `PermissionManager::RIGOR_FULL` and therefore enforces page protection, restricted namespaces, cascading protection, and user blocks — not just the wiki-global `edit` right. The null-pageId fallback is safe by construction (an unresolvable Subject is a no-op write on every path). Nice catch on the operator-precedence bug in `CreateSubjectAction` too (main-subject creation previously required *both* main and child rights).

### Verification (end-to-end, on a running wiki)

Beyond the green suite (1376 tests), I exercised the real `definitelyCan` path against a real protected page, as a **non-sysop user holding the global `edit` right** (the exact threat model):

| Action | Target page | Result |
|---|---|---|
| Create main Subject | sysop-protected | **403** ✅ |
| Create main Subject | unprotected | **201** ✅ |
| Delete Subject | on protected page | **403** ✅ (Subject survived) |
| Replace Subject | on protected page | **403** ✅ |
| Replace Subject | on unprotected page | **200** ✅ (control) |

I also confirmed the headline test `testDeniedOnProtectedPageDespiteGlobalEditRight` is a genuine regression test: reverting `canEditPage()` to the old `isAllowed( 'edit' )` flips it from 403 to 201 (fails), and it creates a real page so `TitleFactory::newFromID` resolves and the assertion actually reaches `definitelyCan`.

### Non-blocking suggestions (author's call)

1. **Resolve the owning page once for delete/replace.** The page is currently resolved via `getPageIdOfSubject()` for the auth check and then again inside the repository for the write (Replace resolves it up to 3×). Threading a single resolved `PageId` from the action into the repository write would (a) remove the redundant Neo4j round-trips and (b) close a narrow TOCTOU: the two independent lookups could in principle disagree if the graph is re-projected between them (e.g. mid-rebuild), which is the only way the "unresolvable ⇒ no-op" safety argument can slip. Very low exploitability today, but resolving once makes the invariant local instead of relying on two reads agreeing. This would also let `SubjectAuthorizer` take a non-nullable `PageId`, so the "null ⇒ fall back to global edit" branch (and its two explanatory comments) disappears.

2. **Consider `authorizeWrite()` instead of `definitelyCan()` for the actual write.** `definitelyCan` uses `RIGOR_FULL` (replica reads — can miss a block/protection applied moments earlier) and passes rate-limit `0` (checks but doesn't count the hit), so these endpoints don't consume the `edit` throttle. `authorizeWrite()` (`RIGOR_SECURE` + counts the hit) is the method intended for authorizing an operation that creates a revision. Neither gap is a regression (the old `isAllowed('edit')` did neither), so it's pure hardening — but worth a conscious decision, especially for API/agent-driven bulk writes that would otherwise bypass the edit rate limit.

3. **Test coverage.** Only *create* has an end-to-end 403-on-protected-page test; delete and replace (and set-main/ordering) rely on the shared authorizer + action-level spy tests. I've live-verified delete/replace deny correctly, but a REST-level denial test per verb would guard the handler wiring against regression. Also worth a test pinning the null-lookup no-op-safety invariant directly. Minor: `ReplaceSubjectActionTest::testAuthorizesAgainstTheSubjectsResolvedPage` asserts `PageId(1)`, which is the repo's default id, so a "passes a hardcoded default instead of the resolved page" bug wouldn't be caught — `DeleteSubjectActionTest` gets this right with a distinctive `PageId(7)`.

4. **Minor / informational.**
   - `CreateSubjectAction` returns `201 created` even when `savePageSubjects` no-ops on an unresolvable page (the `PageContentSavingStatus` from `PageContentSaver` is discarded — the existing `// TODO: expose failure information`). Not a permission issue, but it masks failures on this now security-relevant path.
   - The null fallback uses `isAllowed('edit')`, which doesn't evaluate blocks; a blocked user deleting an unresolvable Subject gets 200/404 rather than 403. Status-code only, no data impact (the write no-ops), but the safety comments reason about protection, not blocks.
   - Replace's inline comment says the update "is a no-op" for an unresolvable Subject; it's actually a 404 (`SubjectNotFoundException` via `getSubject()`), which is slightly more reassuring than "no-op".
   - `tests/phpunit/TestDoubles/SucceedingSubjectActionAuthorizer.php` implements a non-existent `SubjectActionAuthorizer` interface with the old no-arg signatures and is unreferenced — pre-existing dead code the audit surfaced; could be deleted while here.

### Scope note — what remains after this PR

The six subject-write REST endpoints are the complete mutating REST surface for subjects; Schema/Layout have no write REST endpoints (they're namespace-protected pages handled by core), and the `validate`/`cypher` POSTs don't persist (Cypher writes are rejected). One same-class hole is **out of scope and untouched**: `Special:NeoJson::onSubmit()` writes subject content straight through `PageContentSaver` with only a CSRF check — no `definitelyCan`/`isAllowed` gate — so it bypasses page protection for arbitrary named pages. It's inert on a default wiki (`NeoWikiEnableDevelopmentUI` defaults `false`), but it's a full subject-write bypass wherever the dev UI is enabled. Worth a follow-up (same `definitelyCan('edit')` gate, or keeping it strictly dev-gated) — happy to file an issue. The remaining `#101` items are on the read side (per-title read authorization on the GET/validate/cypher surfaces), which this PR explicitly scoped out.

No changes were pushed — all of the above are suggestions for you to weigh; nothing here blocks merge.

> Reviewed by Claude Code at @JeroenDeDauw's direction, via the review-pr workflow.
> Context: the NeoWiki codebase, MediaWiki REST/permission internals (`UserAuthority` / `PermissionManager` / `TitleFactory` / `WikiPageFactory`), a live end-to-end test on a dev wiki, and an adversarial multi-agent review (security-bypass, correctness/edge-case, design/API, tests, and endpoint-scope lenses).
> <sub>Written by Claude Code, `Opus 4.8 (1M context) (max)`</sub>

PR #1002, opened 2026-07-07
Enforce edit permission on the NeoJson special page

alistair3149:
Follows-up to https://github.com/ProfessionalWiki/NeoWiki/pull/986

`Special:NeoJson` lets an editor overwrite a page's Subject content, but its submit handler wrote through `PageContentSaver` with only a CSRF check and no per-title edit authorization. A user holding the wiki-global `edit` right could therefore overwrite Subjects on an edit-protected page or in a restricted namespace — the same class of page-protection bypass fixed for the REST endpoints in #986, here in the special-page entry point. The page is dev-UI gated (inert unless `NeoWikiEnableDevelopmentUI` is on), so it is not exploitable on a default install, but it is a real bypass where the dev UI is enabled.

`SpecialNeoJson::execute()` now checks per-title edit rights via `SubjectAuthorizer::canEditSubject` (`Authority::definitelyCan('edit', $title)`) and throws `PermissionsError` before the form is built, which also blocks a direct POST since `execute()` runs first.

Also folds in three cleanups from the #986 review:

- Delete the dead `SucceedingSubjectActionAuthorizer` test double, which implemented a non-existent `SubjectActionAuthorizer` interface and was unreferenced.
- `ReplaceSubjectActionTest` asserts a distinctive `PageId` rather than the repository's default, so a "passes a default instead of the resolved page" bug would be caught.
- Fix a `ReplaceSubjectAction` comment: an unresolvable Subject 404s, it is not a no-op.

Part of the permission audit tracked in #101.

## Manual Browser Check

Requires `$wgNeoWikiEnableDevelopmentUI = true`.

1. Create a page and protect it for editing (edit = sysop).
2. As a non-sysop user that holds the `edit` right, visit `Special:NeoJson/<ProtectedPageTitle>` → expect a **permission error** (no editable JSON form).
3. Visit `Special:NeoJson/<an unprotected page you can edit>` → the JSON edit form renders as before.

> Directed by @alistair3149; follow-up to the #986 review.
> Context: the NeoWiki codebase and MediaWiki permission internals.


alistair3149:
https://github.com/user-attachments/assets/68672ae6-97fc-4cd9-b0ff-4bfd20d696c9

Issue #993, opened 2026-07-05
Implement the Subject Sources foundation (ADR 23)


JeroenDeDauw:
Tracking issue for implementing the foundation decided in [ADR 23: Subject Sources](https://github.com/ProfessionalWiki/NeoWiki/blob/master/docs/adr/023-subject-sources.md). The design record is the ADR plus the [SubjectSources planning doc](https://github.com/ProfessionalWiki/NeoWiki/blob/master/docs/planning/SubjectSources.md). The work orders live in the comments below and are written to double as Claude Code session prompts; repo-level dev facts (commands, test conventions) land in `CLAUDE.md` via #994.

Whoever picks this up owns the implementation, including the remaining design calls within the ADR's boundaries — most notably the Source interface contract, still listed as an open question in the ADR, which T2 freezes. Calls within the ADR's decisions and Won'ts are the implementer's to make: decide, then record the decision in the PR description. Anything that would contradict the ADR text should be raised here first.

The spine (serial):
- [ ] T1 — Widen `SubjectId` to `(source, localId)`; the bare nanoid form stays valid and means local
- [ ] T2 — Source interface + registry + `LocalSource` refactor; the T2 PR freezes the Source contract and gets @JeroenDeDauw's review before T3/T4 build on it

On the foundation (parallel, after T2 merges):
- [ ] T3 — Schema references become `(source, name)`, resolved through the schema's own Source, degrading gracefully when unresolvable
- [ ] T4 — Restrict Relation targets to resolvable Sources (the v1 guard)

Working agreements:
* T1 → T2 strictly serial; nothing else touching `SubjectId` serialization should be in flight while they are.
* One branch and PR per task, referencing this issue. Keep mechanical changes (renames, signature threading) in commits separate from behavioral ones — that is what makes the wide T1 diff reviewable.
* Track state by ticking this checklist and the work-order boxes.
* Stuck for more than a focused day on one item: stop and write up options instead of grinding. Never weaken T1's backward-compatibility guarantees to make a later task easier.

Already in flight, independent of the spine (sequencing step 1 of the planning doc): #976 (refresh a page's sourced data, #889), #982 (revision content for page-property providers). Related but independent: #831.

Deferred, demand-gated (planning doc §Sequencing 3–4, §Boundaries) — file separately when demand fires, do not build now: sourced Subjects rendered in Views (read-only), the on-wiki SMW/Wikibase source, free-form tables, remote federation, RDF/IRI export, write-back.

[...]

PR #1005, opened 2026-07-08
Widen SubjectId to a source-qualified pair

malberts:
For https://github.com/ProfessionalWiki/NeoWiki/issues/993 (work order T1).

`SubjectId` (PHP and TS) is now a `(source, localId)` pair per the Identity section of
[ADR 23](https://github.com/ProfessionalWiki/NeoWiki/blob/master/docs/adr/023-subject-sources.md):
the bare nanoid form stays valid everywhere and means the local source, and a qualified
`source:localId` form can be carried across every boundary. This is a type-and-wire-format change
only: nothing can be done with a sourced id yet, no path creates sourced Subjects, and local
Subjects stay stored bare (slot JSON, Neo4j `id` properties, and API output unchanged, per ADR 22).

Owner design calls made in this PR, within the ADR's boundaries:

- **Serialized form**: `source:localId`, splitting on the first `:` only, so future URI-shaped
  localIds stay expressible. `:` is absent from both the nanoid alphabet and the key grammar, so
  the two forms cannot collide.
- **Source-key grammar**: `[A-Za-z0-9+_-]+`, compared case-sensitively. The authority for what a
  wiki's key can look like is the installer, not `WikiMap` (which concatenates verbatim): MySQL
  dbnames validate as `/^[a-z0-9+_-]+$/i` and SQLite dbnames are unchecked. Case-sensitivity is
  load-bearing rather than style: with `lower_case_table_names=0` (the MySQL/MariaDB default on
  Linux) two dbnames differing only in case are distinct databases, so two wikis' keys can differ
  only in case too, and the graph's `wiki_id` property stores the wiki ID verbatim, which
  forecloses case-folding. Because hand-set dbnames are unconstrained, a configured local key
  outside the grammar never throws; it is compared verbatim, and such keys simply cannot appear in
  qualified input strings.
- **Canonicalization**: an id that explicitly names the local source normalizes to bare at parse
  time, so one Subject never has two textual identities (`equals()` and the text-keyed collections
  stay sound). Parsing happens in the new `SubjectIdParser` (PHP and TS), constructed with the
  local source key (`WikiMap::getCurrentWikiId()` server-side, `wgWikiID` client-side, which is
  the same value by construction) and threaded to the boundary parse sites from the composition
  root. Ids read back from the local graph keep the plain constructor, being canonical by
  construction.
- **Interim localId rule** (until T2 delegates grammar to Sources): non-empty and restricted to
  URL-path-segment-safe characters (RFC 3986 pchar minus percent-encoding). Looser rules cannot
  round-trip: MediaWiki REST rawurldecodes path parameters after routing, so for example `%2F`
  would reach the handler as `/`. `RestSubjectRepository` now also percent-encodes ids in URLs,
  as the schema and layout repositories already do (a byte-for-byte no-op for bare ids).
- **Placeholder hydration**: `{{#view:}}` injects arbitrary wikitext strings into the placeholder
  attribute, and the subject store is keyed by canonical id text, so hydration now parses ids
  instead of constructing them raw: an explicitly-local-qualified id resolves to the same store
  key as its bare form, and invalid input keeps degrading to a skipped view. The hydration logic
  moved to `components/viewData.ts` to be testable.

PHP/TS grammar parity is asserted rather than assumed: both suites consume the shared vector
fixture `tests/subject-id-vectors.json` (a new convention, per the work-order checklist item).
Object-level semantics stay in explicit per-language tests, and
`testLocalOnlyPageSlotJsonRoundTripsByteIdentically` pins that a local-only page's slot JSON
serialization is byte-identical through the new parse path.

> Result of extensive back and forth with @malberts, building on the T1 work order.
> Context: the NeoWiki codebase, ADRs 14/22/23 and the SubjectSources planning doc, the MediaWiki
> core checkout (REST router, installer, ResourceLoader), and an adversarial multi-agent review of
> the implementation plan.


JeroenDeDauw:

PR #1019, opened 2026-07-10
Align the frontend permission hints with the server

JeroenDeDauw:
Follows-up https://github.com/ProfessionalWiki/NeoWiki/pull/1003

Two commits.

## 1. Hint Subject writes on the edit right

The frontend hints disagreed with the authorization the server applies. `canDeleteSubject` hinted on the `delete` right, and the two creation hints on `createpage` plus `edit`, while the server authorizes every Subject write as an `edit` of the page holding the Subject.

`delete` is a sysop right by default, so the delete affordance in the subjects manager was hidden from ordinary editors who are in fact allowed to delete a Subject. The creation hints were wrong the same way: adding a Subject to an existing page creates no page.

All four hints now check `edit`, mirroring `AuthorityBasedSubjectAuthorizer`. Removing a Subject writes a new revision of the page and leaves the old one intact, so it is an edit by the criterion MediaWiki's own rights encode: `delete` is elevated because it takes content out of public history and needs a paired restore right. Nothing here does. The same removal is already reachable through `Special:NeoJson`, which is gated on `edit`, so a stricter gate on the dedicated UI would be bypassable rather than protective.

They remain hints: they check the wiki-global right, whereas the server checks `edit` on the specific page. A user who may edit globally but not the page at hand is still offered affordances the server rejects. Recorded as a TODO — closing it means feeding the server's per-page answer to the frontend.

## 2. Rename the authorizers to permission hints

The frontend has no binding authorization and cannot have any. Naming these classes `Authorizer` implied a security control that does not exist client-side — the same conflation #1003 removed on the server. `SubjectAuthorizer`, `SchemaAuthorizer` and `LayoutAuthorizer` become `*PermissionHints`, implementations, services and factories following. The backend split is not mirrored: there is no frontend counterpart to `SubjectWriteAuthorizer`, because there is nothing for it to authorize.

## Verification

The authorizer spec pinned the old rights, so it had to change. Reverting the production change makes 7 of its assertions fail, including the three naming rights that must *not* be required (`does not need the delete right to delete a subject`, plus the two `createpage` cases).

`make tsci` (vitest, build, lint) passes: 99 test files, 1007 tests. PHP suite unaffected: 1400 tests.

> Direction from @JeroenDeDauw across ~3 rounds: he asked for this as a follow-up PR, first proposed re-adding the PHP `canDeleteSubject` and accepted the correction that the mismatch is frontend-side, then asked for the naming to follow #1003. He also had the `edit`-vs-`delete` question examined from first principles by a separate model, which independently recommended `edit` with high confidence. Diff not yet human-reviewed. Found while auditing why `SchemaAuthorizer` was dead; tests written after the fix, then verified to fail against the old implementation.
> Context: the NeoWiki PHP and TypeScript codebases, and MediaWiki `Authority` / `RateLimiter` internals traced for #1003.
> <sub>Written by Claude Code, `Opus 4.8 (1M context)` (max)</sub>


resources/ext.neowiki/src/persistence/RightsBasedSubjectPermissionHints.ts
```diff
@@ -2,33 +2,37 @@ import type { SubjectAuthorizer } from '@/application/SubjectAuthorizer';
...
-		return rights.includes( 'edit' );
+		return this.canEditPage();
 	}
 
 	public async canDeleteSubject( _subjectId: SubjectId ): Promise<boolean> {
-		// TODO: check delete right on specific page containing the subject.
-		// TODO: or should this be considered a page edit and use 'edit' right instead?
-		const rights = await this.rightsFetcher.getRights();
-		return rights.includes( 'delete' );
+		return this.canEditPage();
```
JeroenDeDauw, line 27: I did conceptual review of this behavior and think it's correct

PR #1036, opened 2026-07-13
Fix subject-format doc path in the validate endpoints

malberts:
Follows-up to https://github.com/ProfessionalWiki/NeoWiki/pull/1034 (based on its branch).

`ValidateSubjectApi` and `ValidateSubjectUpdateApi` point their `statements` param description at
`docs/SubjectFormat.md`, a path that has never existed (it predates the docs reorganisation). The `create` and
`replace` endpoints already reference `docs/api/subject-format.md`; this points the validate pair at the same doc,
so all four subject endpoints agree.

Based on the #1034 branch because that is where `docs/api/subject-format.md` lives (on `master` the doc is still at
`docs/reference/subject-format.md`), so the fix lands together with the reorg and needs no re-touch afterwards.

> <sub>AI-authored — Claude Code, `Opus 4.8 (max)`; @malberts asked to fix this pre-existing reference flagged during the #1034 review, one-line ask, no revisions; diff not yet human-reviewed; target `docs/api/subject-format.md` confirmed present on the base branch, no test exercises these param strings, CI on the branch.</sub>


alistair3149:
Closing in favor of #1037 because it has a doc drift test

PR #1042, opened 2026-07-15
Add the Source contract, registry, and LocalSource

malberts:
For https://github.com/ProfessionalWiki/NeoWiki/issues/993 (work order T2). Builds on #1005 and targets
its branch, so the diff shows only this track.

This PR freezes the Source interface contract that ADR 23 left open. The contract, its whys, and the
alternatives are recorded in `docs/adr/027-source-interface-contract.md`; the highlights:

- **The `Source` contract** (`Domain\Source`): `getSubject( SubjectId )`, `getSchema( SchemaName )`,
  `isEditable()`, `isValidLocalId( string )`, `getBaseUri()`. Sources play no part at query time:
  queryability is materialisation into the graph (ADR 23). Schema resolution enters the contract now
  because interface additions break third-party implementers and T3 needs it.
- **`SourceRegistry`**: byte-exact key comparison (two Linux MySQL/MariaDB dbnames can differ only in
  case, and the graph's `wiki_id` stores the wiki ID verbatim, so case-folding is foreclosed),
  `LogicException` on duplicate registration, bare ids map to the local source.
- **`LocalSource`** wraps the existing local persistence behind lazily constructed lookups and returns
  the per-wiki RDF base (`NeoWikiConfig->rdfBaseUri`, from the native RDF projection) as its base URI.
- **Read-side routing only**: the `subjectLookup` seams (`GetSubjectQuery`, `GetPageSubjectsQuery`,
  both `SubjectResolver` sites) resolve through the registry; write paths and the revision-scoped
  `PointInTimeSubjectLookup` stay direct, since sourced Subjects are read-only per ADR 23.
- **Registration point**: `NeoWikiRegistrar::addSource()` in the `NeoWikiRegistration` hook, with core
  registering `LocalSource` under the wiki ID before the hook fires, so colliding extension
  registrations fail fast. The lazy construction wrappers (`LazySubjectLookup`, `LazySchemaLookup`)
  are load-bearing here: an eagerly built dependency chain fires the registration hook re-entrantly,
  and the hook would observe an empty registry (probe-verified in both process orders).
- **Unresolvable source keys** degrade to not-found plus one logged warning, never fatally. REST GET
  keeps the pinned not-found contract (HTTP 200 with `{"subject":null}`).
- **Deliberately preserved**: on a wiki with no graph backend, `{{#neowiki_value}}` and Lua keep
  failing loudly (`GraphBackendNotConfiguredException` is rethrown through `SubjectResolver`'s
  catches); the softer degradation belongs to the #895 track, not this one.
- **Deferred by design**: no unused write-capability stub. Write-back, when designed, arrives as a
  `WritableSource extends Source` extension, which is non-breaking for implementers.

> <sub>AI-authored — Claude Code, `Fable 5 (max)`; standing directive from @malberts to complete the #993 tracks without blocking on inter-track human review, design calls delegated; diff not yet human-reviewed; TDD with watched failures, three adversarial AI review lenses with live hook-order and backendless probes, `make ci` and `make ts-ci` green locally, branch CI pending at posting time.</sub>

Design, contract decisions, and review adjudication by `Fable 5 (max)`; implementation and the fix
round by `Opus 4.8 (max)` subagents; three parallel `Opus 4.8 (max)` review lenses (contract
fidelity, runtime correctness, test quality) stood in for the usual pre-PR human review at
@malberts' direction. The correctness lens probe-confirmed the hook-ordering hazard and drove the
lazy-construction fix and the loud-backendless preservation described above.

</details>

### DIFF
--- a/docs/adr/027-source-interface-contract.md
+++ b/docs/adr/027-source-interface-contract.md
@@ -1,0 +1,74 @@
+# Source Interface Contract
+
+Date: 2026-07-15
+
+Status: Draft (frozen when the T2 PR merges)
+
+## Context
+
+ADR 23 decided that Subjects come from pluggable Sources and left the Source interface contract as
+an open question. Issue #993's T2 work order assigns freezing it to the T2 PR. The contract must
+serve the wiki-farm case (more registered Sources), the future on-wiki SMW/Wikibase and federation
+adapters, and T3's schema references, without a per-Source capability matrix (ADR 23 Won'ts).
+
+## Decision
+
+A single `Source` interface (Domain\Source) with exactly five members:
+
+- `getSubject( SubjectId $id ): ?Subject` — by-id resolution; the registry routes, the Source
+  interprets its own localIds.
+- `getSchema( SchemaName $schemaName ): ?Schema` — schemas come from Sources; a schema's source is
+  independent of a subject's source.
+- `isEditable(): bool` — the entire capability surface: local editable + versioned, sourced
+  read-only.
+- `isValidLocalId( string $localId ): bool` — each Source owns its localId grammar; the serialized
+  qualified form additionally restricts localIds to URL-path-safe characters at the wire level.
+- `getBaseUri(): ?string` — IRI projection base for this Source's ids. Native RDF projection mints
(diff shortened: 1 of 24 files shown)


JeroenDeDauw:

PR #1044, opened 2026-07-15
Resolve schema references through their Source

malberts:
For https://github.com/ProfessionalWiki/NeoWiki/issues/993 (work order T3). Builds on the T2 Source
contract PR and targets its branch, so the diff shows only this track.

Schema references become `(source, name)`, resolved through the schema's own Source and degrading
gracefully when unresolvable. Reference plumbing only; no non-local schema source exists yet.

Design calls recorded here:

- **A distinct `SchemaReference` value object** with `SchemaName` intact underneath, per the work
  order's own lean: locally the name IS the page title in the Schema namespace, and a
  source-qualified reference is not a valid title.
- **Foreign references serialize as a JSON object `{ "source": ..., "name": ... }`, not a
  `source:Name` string.** Schema names are page titles and may legitimately contain colons, so a
  string-concatenation grammar is ambiguous; subject ids avoid that only because the bare nanoid
  alphabet excludes the separator. Local references stay bare name strings, byte-identical to
  before (regression-pinned on both persisted spots).
- **Canonicalization**: on the subject `schema` field, an object naming the local Source normalizes
  to the bare form at parse time (mirroring `SubjectIdParser`), so one Schema is never referenced
  under two identities. A relation property's `targetSchema` is deliberately not at full parity
  yet: schema save validation still accepts only the bare string form, so the object form has no
  ingress there, and normalization is omitted where it can never run. Widening the schema-save
  contract and canonicalizing that path belongs with sourced-schema authoring; flip it in review if
  you prefer parity now.
- **Resolution** routes through the T2 registry via `SourceRoutingSchemaReferenceResolver`,
  mirroring the subject routing semantics exactly: unknown source keys degrade to null plus one
  logged warning, and unresolvable references land in the pre-existing `schema-not-found` /
  null-schema handling in all five converted consumers (validators, replace action, RDF projector,
  Neo4j updater), never fatally. Name-parameterized paths (REST schema APIs, new-subject creation)
  stay name-based.
- **The Neo4j schema-name node label stays local-only** (work-order constraint), and the frontend
  reduces string-or-object references to the bare name: sourced rendering is deferred by ADR 23, so
  the TS domain keeps plain names end to end.

Docs: `docs/api/subject-format.md` gains the Schema References section, `docs/api/schema-format.md`
documents the `targetSchema` forms and their current acceptance honestly, and the glossary Schema
entry notes the reference form.

> <sub>AI-authored — Claude Code, `Fable 5 (max)`; standing directive from @malberts to complete the #993 tracks without blocking on inter-track human review, design calls delegated; diff not yet human-reviewed; TDD with watched failures per behavior, two adversarial AI review lenses with live probes (hook-ordering re-trace across five accessor orders, byte-identity on seeded content, ingress enumeration for the canonicalization asymmetry), `make ci` and `make ts-ci` green locally, branch CI pending at posting time.</sub>

Design and review adjudication by `Fable 5 (max)`; implementation by an `Opus 4.8 (max)` subagent;
two `Opus 4.8 (max)` review lenses (runtime correctness with container probes, tests/conventions/
docs) stood in for the usual pre-PR human review at @malberts' direction. The review confirmed the
targetSchema asymmetry is unreachable through every current ingress and found only docs overclaims,
which the final commit scopes to shipped behavior.

</details>

### DIFF
--- a/docs/api/schema-format.md
+++ b/docs/api/schema-format.md
@@ -189,9 +189,10 @@ References to other Subjects.
 
 `targetSchema` uses the same reference form as a Subject's `schema`: a bare name string for a local
 Schema, or an object `{ "source", "name" }` for a Schema from another Source. Schema save validation
-currently accepts only the bare string form; the object form becomes saveable when sourced schemas
-arrive. Normalization of references naming the local Source happens on the subject `schema` field
-only, so a `targetSchema` is stored as given.
+currently accepts only the bare string form; saving the object form, and presenting a Subject that
+references a Schema from another Source, are part of the deferred sourced-Schema work (see
+[Schema References](subject-format.md#schema-references)). Normalization of references naming the
+local Source happens on the subject `schema` field only, so a `targetSchema` is stored as given.
 
 Example:
 
--- a/docs/api/schema-format.md
+++ b/docs/api/schema-format.md
@@ -188,8 +188,10 @@ References to other Subjects.
 | `multiple` | boolean | No | `false` | Allow multiple relations |
 
 `targetSchema` uses the same reference form as a Subject's `schema`: a bare name string for a local
-Schema, or an object `{ "source", "name" }` for a Schema from another Source. A reference naming the
-local Source normalizes to the bare form.
+Schema, or an object `{ "source", "name" }` for a Schema from another Source. Schema save validation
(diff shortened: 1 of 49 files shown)


JeroenDeDauw:

PR #1051, opened 2026-07-16
Add SPARQL graph store plugin: projection, query surfaces and dev-stack QLever

JeroenDeDauw:
For https://github.com/ProfessionalWiki/NeoWiki/issues/586

Adds a SPARQL 1.1 graph-store backend that keeps a triple store (QLever, or any SPARQL 1.1 store) in sync with page changes alongside the existing Neo4j backend, and read-only SPARQL query surfaces over it.

A SPARQL store does not yet replace Neo4j: the interactive features (Subject editing UIs, views, value accessors) still require a configured Neo4j backend.

## Sync

Implements the mechanism from `planning/NativeRdfProjection.md`: each page is a named graph identified by the page IRI; a save replaces that graph (`DROP SILENT` + `INSERT DATA`), a delete drops it. The plugin is seeded as a bundled backend in the graph-database fan-out, so it runs behind the same per-plugin failure isolation as Neo4j — a failing store never aborts the triggering edit — and `RebuildGraphDatabases.php` reaches it like any other backend.

## Configuration

```php
$wgNeoWikiSparqlStores = [
    [
        'updateUrl' => 'https://qlever.example/api/neowiki', // required
        'queryUrl' => 'https://qlever.example/api/neowiki',  // optional, defaults to updateUrl
        'accessToken' => 'SECRET',                           // optional, sent as Bearer
        'projection' => 'native',                            // optional, or a Mapping target
    ],
];
```

An entry without a usable `updateUrl` is skipped with a warning rather than failing the wiki. `queryUrl` defaults to `updateUrl` because QLever serves both from one endpoint; a store that separates them (a read replica, or a read-protected endpoint) sets it explicitly.

The configured projection is resolved lazily on each save (through the `newRdfProjection` seam), never at construction, so a store always tracks the current Mapping definitions and plugin construction stays I/O-free. Triples are serialized prefix-less (full IRIs, no `@prefix` directives) so they are valid inside `INSERT DATA`.

## Query surfaces

Three read-only surfaces, registered only when a store is configured, each targeting the first configured store:

- the `{{#sparql_raw}}` parser function
- `nw.sparqlQuery()` in Lua
- `POST /neowiki/v0/query/sparql`

They are read-only *by protocol*: the query is sent as a SPARQL 1.1 *query* operation (`Content-Type: application/sparql-query`), whose grammar has no update forms, so no read-only validator is needed (unlike the Cypher surfaces). The W3C `application/sparql-results+json` document is returned unmodified, with no NeoWiki envelope. The per-tier timeout from `$wgNeoWikiQueryLimits` applies; the `maxRows` cap deliberately does not, since truncating `results.bindings` would alter the W3C document — bound result size with `LIMIT` instead.

## Dev stack, CI and demo data

The dev stack gains a QLever container (`QLEVER_ACCESS_TOKEN` in `Docker/.env.dist` — existing checkouts need it added to their `Docker/.env`, which is not regenerated). CI starts a QLever and runs a system test that drives create, edit and delete against it through the real hook path. `DemoData/SparqlPage/SPARQL_queries.wikitext` demonstrates all three surfaces, and is imported only into wikis that have a store configured.

## Out of scope, for later PRs

RDF import, and multi-store query addressing (the read surfaces use the first configured store).

> <sub>AI-authored — Claude Code, `Opus 4.8 (max)`; detailed spec from @JeroenDeDauw, executed autonomously with no human redirection; AI-reviewed in a later session, not yet human-reviewed; full PHPUnit suite (1738 tests), phpcs and phpstan green locally, CI green, and the sync, rebuild, delete and all three query surfaces driven end-to-end against a live QLever.</sub>

Built by one session from the spec, then in a later session: `master` merged in (the #1034 docs reorganisation moved the docs this branch links to), the diff reviewed, and one blocking bug fixed — `{{#sparql_raw}}` handed its JSON back as wikitext, so the parser autolinked the IRIs every results document carries and corrupted it. That bug was invisible to the unit tests, which assert on what the parser function returns rather than on what a reader gets; the added test parses through a real `Parser`.

</details>

### DIFF
--- a/.github/workflows/ci-php.yml
+++ b/.github/workflows/ci-php.yml
@@ -33,6 +33,11 @@ jobs:
     env:
       NEO4J_URL_OVERRIDE: bolt://neo4j:password@localhost:7689
       NEO4J_URL_READ_OVERRIDE: bolt://mediawiki_read:mediawiki_read@localhost:7689
+      # Live QLever store for the SPARQL query system test. Overrides the dev-stack hostname in
+      # phpunit.xml.dist (PHPUnit <env> does not override an existing environment variable), so the test
+      # reaches the QLever the "Start QLever" step publishes on the runner's localhost.
+      QLEVER_TEST_URL: http://localhost:7019/
+      QLEVER_TEST_ACCESS_TOKEN: neowiki_test_token
 
     services:
       test_neo:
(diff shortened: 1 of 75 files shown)


JeroenDeDauw:
image

Issue #1062, opened 2026-07-16
totalRows count oracle on GET /layouts and GET /schemas


alistair3149:
Low-severity count oracle left out of #1046 (PR #1058). After that PR, `GET /layouts` and `GET /schemas` correctly omit rows for pages the caller cannot read from their returned arrays, but the `totalRows` field still reports an **unfiltered** namespace count:

- `GetLayoutSummariesApi` → `DatabaseLayoutNameLookup::getLayoutCount()`
- `GetSchemaSummariesApi` → `DatabaseSchemaNameLookup::getSchemaCount()`

Both are raw `COUNT(*)` over the namespace. A client can page through the visible rows and infer, from `totalRows − visible`, how many restricted Layouts/Schemas exist. Names and content stay hidden, so this only leaks a **count**.

An honest `totalRows` would require a per-title permission check over the whole namespace (unbounded work). Recommended instead: drop `totalRows`, or report it as the count of the returned (filtered) page, rather than scanning. `GET /schema-names/{search}` has no `totalRows` and is unaffected.

Documented as a known gap in `docs/api/rest-api.md`.

> Split out of #1046 (PR #1058). Directed by @alistair3149.
> <sub>Written by Claude Code, `Fable 5`</sub>

alistair3149:
`GET /neowiki/v0/mappings` (added in #1105) is a third instance of this same `totalRows` count oracle.

`GetMappingSummariesApi::run()` filters unreadable rows correctly — `CachingMappingLookup::getMapping()` returns `null` via `authorizeReadByPageTitle()`, and the handler skips them — but then reports:

```php
'totalRows' => count( $names ),   // $names = getMappingNameLookup()->getMappingNames(), unfiltered
```

so `totalRows − visible` leaks the number of restricted Mappings, exactly as for `/layouts` and `/schemas`. Whatever fix lands here (drop `totalRows`, or report the count of the returned filtered page) should cover this endpoint too.

Unlike the DB-`COUNT(*)` case for Layouts/Schemas, the mappings handler already materializes the full unfiltered name list, so reporting `count( $summaries )` (the filtered page) is essentially free here.

Separately (not this issue, noting for the #1046 family): the Permissions section of `docs/api/rest-api.md` enumerates the read endpoints that enforce per-page `read` (Subject, page-subjects, Schema, Layout, RDF export, entity-dereference) but omits the new Mapping read endpoint, which also enforces it.

[...]

Issue #1060, opened 2026-07-16
GET /subject-labels leaks labels of Subjects on unreadable pages


alistair3149:
`GET /neowiki/v0/subject-labels` was left out of the #1046 read-gate work (PR #1058) because gating it needs a graph-query change. It currently returns the labels of Subjects on pages the caller cannot read, feeding them into autocomplete results.

Root cause: `Neo4jSubjectLabelLookup::getSubjectLabelsMatching()` (`src/GraphDatabasePlugins/Neo4j/Persistence/Neo4jSubjectLabelLookup.php`) returns only `n.id` and `n.name` — **no page identifier reaches PHP**, so per-title filtering is impossible without changing the Cypher to traverse `Page-[:HasSubject]->Subject` and return the page id.

Work required, in order:

1. **Add a `PARAM_MAX` cap on `limit` first.** `GetSubjectLabelsApi` has `PARAM_DEFAULT => 10` but **no `PARAM_MAX`** — `?limit=100000` is accepted today. A per-result permission check without a cap is an unbounded-work DoS. (Contrast `/layouts`, capped at 50.)
2. Traverse to the Page node in the query, resolve `Title::newFromID()` (batch via `LinkBatch`), and filter with `authorizeRead( 'read', $title )`.
3. Over-fetch and re-limit, since filtering after `LIMIT` yields short pages.

> Split out of #1046 (PR #1058). Directed by @alistair3149.
> <sub>Written by Claude Code, `Fable 5`</sub>

PR #1071, opened 2026-07-17
Pin the Schema name search read gate and correct the permissions docs

JeroenDeDauw:
Follows-up to https://github.com/ProfessionalWiki/NeoWiki/pull/1058

Two fixes from an AI review of #1058. Both are small. The review's more significant findings are deliberately **not** in this branch — they need your call, and are reported separately.

### Test the read gate on the Schema name search branch

Both denial tests in `DatabaseSchemaNameLookupTest` call `getSchemaNamesMatching` with an empty search, so they only cover `getFirstSchemaNames`. Deleting `filterReadable` from the search branch — the one `GET /schema-names/{search}` actually uses — passed the full suite. The production code is correct; the gate was simply unpinned. The new test fails without the gate and passes with it.

### Correct the REST permissions documentation

Two inaccuracies:

* It claimed page protection and restricted namespaces apply to reads. Neither does: for the `read` action MediaWiki runs only `checkPermissionHooks`, `checkReadPermissions` and `checkUserBlock`; `read` is not one of the default `$wgRestrictionTypes`; and `$wgNamespaceProtection` is only consulted by `checkSpecialsAndNSPermissions`, which reads skip. Per-page read restrictions come from `$wgWhitelistRead` or a permission extension.
* It stated without qualification that read endpoints enforce per-page `read`. `GET /subject-labels` applies no per-page filter. Verified on a dev wiki with one page read-denied: `GET /subject/{id}` returns `{"subject":null}` while `GET /subject-labels` still returns that Subject's id and label. `totalRows` likewise still counts rows omitted from the list.

This partially restores caveats that 4cb79db trimmed as "temporary not-yet caveats". Flagging rather than assuming your intent: #1058's description says the deferred gaps are "Documented in the new `docs/api/rest-api.md` Permissions section", and the trim removed exactly that, which left the remaining blanket claim false. If you prefer different wording, the only constraint is that the section not assert a guarantee `/subject-labels` does not provide.

### Verification

`make cs` clean; full suite green (1689 tests). New test proven load-bearing by mutation.

> <sub>AI-authored — Claude Code, `Opus 4.8 (max)`; produced by a `/pr-review` run of #1058 for @JeroenDeDauw, no revisions; diff not yet human-reviewed; new test mutation-tested, `make cs` clean, full PHPUnit suite green locally, docs claims checked against MediaWiki 1.43 core and reproduced live on a dev wiki.</sub>

### DIFF
--- a/docs/api/rest-api.md
+++ b/docs/api/rest-api.md
@@ -23,8 +23,9 @@ Demo wiki example: [neowiki.dev/w/rest.php/specs/v0/module/-](https://neowiki.de
 
 ## Permissions
 
-Read endpoints enforce the caller's per-page `read` permission: page protection, restricted namespaces, read
-whitelists, and permission extensions all apply. When you may not read a page, its read endpoints respond as if
+The Subject, Schema, Layout, Mapping and RDF read endpoints enforce the caller's per-page `read` permission: read
+whitelists and permission extensions apply. (MediaWiki's `read` action ignores page protection and
+`$wgNamespaceProtection`, so neither restricts these endpoints.) When you may not read a page, they respond as if
 the data were absent — a `null` value, an empty list, or a `404` — rather than with a `403`. Treat a not-found
 response as "not available": it may mean the data does not exist, or that you may not read it.
 
--- a/tests/phpunit/Persistence/MediaWiki/DatabaseSchemaNameLookupTest.php
+++ b/tests/phpunit/Persistence/MediaWiki/DatabaseSchemaNameLookupTest.php
@@ -141,6 +141,19 @@ class DatabaseSchemaNameLookupTest extends NeoWikiIntegrationTestCase {
 		);
 	}
 
+	public function testUnreadableSchemaNamesAreOmittedFromSearchResults(): void {
+		$denyFirstMatch = static fn ( string $permission, ?PageIdentity $page = null ): bool =>
+			$page === null || $page->getDBkey() !== 'SchemaNameLookupTest21';
+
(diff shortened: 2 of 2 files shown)


docs/api/rest-api.md
```diff
@@ -23,15 +23,21 @@ Demo wiki example: [neowiki.dev/w/rest.php/specs/v0/module/-](https://neowiki.de
...
 
-The Cypher query endpoint is gated only by the `neowiki-query` right, with no per-page filtering (see
-[Query API](query-api.md)).
+Not every read endpoint is per-page filtered:
+
+* `GET /neowiki/v0/subject-labels` returns the labels and ids of Subjects on pages you may not read.
+* The Cypher query endpoint is gated only by the `neowiki-query` right (see [Query API](query-api.md)).
+* On `GET /neowiki/v0/schemas` and `GET /neowiki/v0/layouts`, rows you may not read are left out of the list, but
+  `totalRows` still counts them and the filtering happens after paging — so a page of results can come back short,
+  or empty while later pages still have rows.
```
JeroenDeDauw: Both of the added bullets here seem like something to fix

JeroenDeDauw: Ah, already tracked by
* https://github.com/ProfessionalWiki/NeoWiki/issues/1062
* https://github.com/ProfessionalWiki/NeoWiki/issues/1060

JeroenDeDauw: I'm removing these added bullets. Not sensible to put temporary loopholes in the docs

Issue #1046, opened 2026-07-16
Enforce per-page read permission on Subject, Layout, and RDF read endpoints


alistair3149:
Sub-issue of #87, tracking the backend half of its scope: the Subject, Layout, and RDF read
endpoints never check whether the caller may read the page the data lives on.

The only per-title `read` gates in the backend are `CachingSchemaLookup` and
`CachingMappingLookup` (both `probablyCan( 'read', $title )`), which is why the Schema and
Mapping endpoints are fine. The other read paths apply RevDel audience checks
(`FOR_THIS_USER`) but no per-title `read`, so on a read-restricted wiki they expose data
from pages the caller cannot read. MediaWiki's REST framework checks only the wiki-global
`read` right: a fully private wiki is protected wholesale, while partially restricted
setups (restricted namespaces, whitelist read, permission-filtered farms) are not.

Ungated endpoints, re-verified against master 2026-07-16:

- `GET /subject/{subjectId}` (including `?revisionId=`) — `GetSubjectQuery` reads through the Subject repository with audience checks only.
- `GET /page/{pageId}/subjects` — same path.
- `GET /layout/{layoutName}` — `WikiPageLayoutLookup` receives an `Authority` but uses it only for the audience check.
- `POST /subject/{subjectId}/validate` — reveals Subject existence (404 vs 200) and validates against the Subject's Schema.
- `GET /subject-labels` — `Neo4jSubjectLabelLookup` takes no `Authority`, so labels of Subjects on unreadable pages appear in autocomplete results.
- `GET /page/{pageId}/rdf` — added after the #101 audit, so not in the original #87 scope list. `RdfPageLoader` reads the Subject slot straight off the current revision (`getSlots()->getContent()` without an audience argument) and exports Page Properties (last editor, categories, timestamps). The missing audience argument is largely moot because the current revision cannot be revision-deleted; the per-title `read` gate is the real gap.

Lower priority, same root cause: `GET /layouts` and `GET /schema-names/{search}` enumerate
names from namespaces the caller may not read. In default setups these namespaces are
world-readable, so this only matters under per-namespace read restrictions.

Fix direction: mirror the `CachingSchemaLookup` gate on these read paths. Most already
hold an `Authority` and a page reference or can resolve one via `PageIdentifiersLookup`;
`/subject-labels` additionally needs page identifiers for the returned Subjects to filter
the result list.

[...]

Issue #1040, opened 2026-07-13
Make the subject-to-page lookup MediaWiki-native so Subjects work without a graph backend


alistair3149:
NeoWiki currently requires a configured graph backend for everything Subject-related. Subject ids exist only as keys
inside the `neo` slot JSON, and the only index from a Subject id back to its page is the Neo4j projection
(`Neo4jPageIdentifiersLookup`, a `HasSubject` edge query). `NeoWikiExtension::getPageIdentifiersLookup()` builds it
unconditionally, so Subject CRUD, `{{#view}}`, `{{#neowiki_value}}` and the `mw.neowiki.*` getters all fail on a wiki
with no backend. This is the last open item of #895, and it decides whether a Neo4j-free deployment (e.g. QLever-only
for ECHOLOT, #586) is possible.

Filing for discussion and tracking; direction below is a proposal, not a done decision.

## Proposed direction

Make the subject → page lookup MediaWiki-native and authoritative — the implementation, not a fallback:

```sql
CREATE TABLE neowiki_subject_page (
  nwsp_subject_id VARBINARY(32) NOT NULL,
  nwsp_page_id    INT UNSIGNED  NOT NULL,
  INDEX nwsp_page_id (nwsp_page_id),
  PRIMARY KEY (nwsp_subject_id, nwsp_page_id)
);
```

Written from `OnRevisionCreatedHandler`, next to the existing graph fan-out. `RevisionFromEditComplete` fires inside
`PageUpdater`'s atomic section, so the index commits with the revision or not at all: it can never be staler than the
slot it derives from, and replica reads get the same session-consistency guarantees as the revision itself. Reads join
`page`, so page moves need no index maintenance. Graph backends stay what ADR 19 says they are: rebuildable query
projections.

## Why the index cannot stay in the graph

- **#1028 made projection staleness an accepted state.** Backend failures during edit/delete/undelete are swallowed by
  design so the user's operation commits. An index that authorization and write routing stand on cannot live in
  storage that is allowed to lag: a Subject written during an outage would be unfindable and uneditable until an admin
  rebuilds.
- **Per-plugin failure isolation makes multi-backend identity nondeterministic.** One backend can miss an event
  another received, so two configured backends can hold different subject → page truths, and which one answers becomes
  an accident of wiring order. Tolerable for query results, not for identity.
- **The current lookup has a latent farm bug.** Writes key `Page` nodes on `{id, wiki_id}` (ADR 22), but the reverse
[...]

PR #1082, opened 2026-07-17
Validate relation statement targets server-side

### LINKED ISSUE #1078: Relation values are not validated server-side: target existence, target schema, and single-value cardinality
`RelationType::validate()` only emits `required`. The `targetSchema` attribute is stored and drives the frontend picker, but `RelationProperty::getTargetSchema()` has no other consumers, and nothing checks relation targets on write. Concretely, an API write succeeds when a relation:

* points at a Subject ID that does not exist,
* points at a Subject whose Schema is not the property's `targetSchema`, or
* holds multiple targets on a `multiple: false` property (`SelectType` enforces `single-value-only`; the relation type does not).

The subject editor's picker filters by target schema, so this is mostly unreachable through the UI — but the stored data is only trustworthy as long as nobody uses the API, and API-driven import is a core use case.

Proposed behavior, following the severity model of ADR 026:

* `single-value-only` on a `multiple: false` relation with more than one target — `error`, matching Select.
* Target's Schema differs from `targetSchema` — `error` (new code, e.g. `relation-target-schema-mismatch`).
* Target Subject does not exist — `warning` (new code, e.g. `relation-target-not-found`). Deliberately non-blocking: pointing at a not-yet-created Subject is wiki-native red-link behavior, and imports may legitimately create targets later.

Enforcement continues to ride the existing validation-enforcement mechanism (blocking only newly introduced `error` violations). Statement-scoped validation in `SubjectValidator` (the placement PR #1043 uses for Source resolvability) looks like the natural home, since `PropertyType::validate()` has no access to lookups.

Part of the Relations work tracked in #630.

> <sub>AI-authored — Claude Code, `Fable 5 (max)`; filed at @JeroenDeDauw's direction from a Relations design-review session; not yet human-reviewed; gap verified on origin/master (`validate()` emits only `required`; `getTargetSchema()` has no consumers).</sub>

JeroenDeDauw:
Fixes https://github.com/ProfessionalWiki/NeoWiki/issues/1078

`RelationType::validate()` only emitted `required` and `RelationProperty::getTargetSchema()` had no
consumers, so relation statements went unchecked on write. An API write accepted a relation that
points at a Subject that does not exist, at a Subject whose Schema is not the property's
`targetSchema`, or that holds multiple targets on a `multiple: false` property. The subject editor's
picker hides this, but API-driven import does not go through the picker.

This adds three validations, using the severity model of ADR 26 as actually implemented (blocking vs.
`Violation::NON_BLOCKING_CODES`):

- `single-value-only` (`error`) for a `multiple: false` relation holding more than one target, emitted
  by `RelationType::validate()` exactly as `SelectType` does — no lookups needed.
- `relation-target-schema-mismatch` (`error`, new) when the target Subject exists but its own Schema
  is not the property's `targetSchema`.
- `relation-target-not-found` (`warning`, new, non-blocking) when the target id does not resolve.
  Deliberately non-blocking: pointing at a not-yet-created Subject is wiki-native red-link behaviour,
  and an import may mint the target later.

The last two live in `SubjectValidator` (the placement PR #1043 uses), since they need a lookup and
the writer's-schema `RelationProperty`, which `PropertyType::validate()` cannot reach. Targets are
resolved through the canonical revision-slot-backed `SubjectLookup` the read path uses
(`getSubjectRepository()`), never the secondary Neo4j projection; the target's Schema comes from its
own stored Subject. A missing target short-circuits, so it never also reports a schema mismatch.

Enforcement is unchanged: `relation-target-not-found` joins `Violation::NON_BLOCKING_CODES` (the
warning tier as implemented), and the newly-introduced-blocking-violation gate is untouched.

`RelationInput` already renders server violations generically via `neowiki-field-<code>` messages, so
the two new codes get `i18n/en.json`, `i18n/qqq.json`, and `extension.json` entries and no new UI
code. `docs/api/validation-codes.md` documents both codes and notes `RelationType` as a
`single-value-only` emitter.

ADR 26 (Draft) is intentionally not edited: its per-Constraint severity model is not fully
implemented, and PR #1043 set the precedent of not treating it as a living code catalog. Folding
these two fixed-severity system codes (mismatch = fixed `error`, not-found = fixed `warning`) into
ADR 26 is a follow-up for whoever finalizes it.

A textual conflict with open PR #1043 (which also touches `SubjectValidator`, the same three test
files, `validation-codes.md`, and the i18n / extension.json message list) is expected.

> <sub>AI-authored — Claude Code, `Opus 4.8 (max)`; batch execution of the Relations Phase 1 plan directed by @JeroenDeDauw, from a fixed spec with no redirection; not yet human-reviewed; TDD (red then green), full PHPUnit suite (1801 tests) + `make cs` (phpcs + phpstan) green locally, psalm clean on changed files, production-revert test-verification done; CI green (PHPCS, PHPStan, PHPUnit).</sub>

## Review notes (batch pr-review pass)

A fresh-session review pass fixed a blocking enforcement gap (relation-target violations now carry `valuePartIndex`, 122712bc) and verified all three validations end to end (unit + mutation battery + live endpoint). Two architecture-level follow-ups were identified and deliberately left out of this PR pending a decision:

* The validate path resolves relation targets without page-read authorization, so violation responses can disclose whether an unreadable target exists and what its schema is (the read path gates exactly this per #1046). Likely fix: thread the read authorizer into target resolution and collapse unreadable targets into `relation-target-not-found`.
* Target resolution adds one live lookup per relation target on an anonymous, unrate-limited validate endpoint (no relation cap, no id dedup). Consider a cap, dedup/batching, or a ping limiter before ACL/farm deployments.

> <sub>Review-pass notes added by Claude Code, `Fable 5 (max)`, summarizing a fresh-session `Opus 4.8` review agent's verified findings; not yet human-reviewed.</sub>

### DIFF
--- a/docs/api/validation-codes.md
+++ b/docs/api/validation-codes.md
@@ -208,11 +208,12 @@ Emitted by `SubjectValidator`.
 `propertyName`: the relation property.
 
 A relation targets a Subject that exists but whose own Schema is not the Property's declared
-`targetSchema`. The target is resolved through the canonical revision-slot-backed Subject lookup (the
-same one the read path uses), so the check reflects the target's stored writer's-schema rather than the
-secondary graph projection. This is a blocking `error` (ADR 26): it is rejected at write time when
-`$wgNeoWikiEnforceValidation` is enabled. The subject editor's picker filters candidates by target
-Schema, so this is normally only reachable through the API — for example a bulk import.
+`targetSchema`. The Schema compared is the target's own stored writer's-schema, read from its revision
+slot rather than from a graph node property. Reaching that slot still resolves the target ID through
+the subject-to-page index, which lives only in the graph projection, so this check only runs for
+targets the graph currently knows about. This is a blocking `error` (ADR 26): it is rejected at write
+time when `$wgNeoWikiEnforceValidation` is enabled. The subject editor's picker filters candidates by
+target Schema, so this is normally only reachable through the API — for example a bulk import.
 
 ### `relation-target-not-found`
 
@@ -227,13 +228,23 @@ hardcoded non-blocking set, so it never rejects a write even under enforcement. 
 pointing at a not-yet-created Subject is wiki-native red-link behavior, and an import may legitimately
 mint the target later.
 
+Resolution goes through the subject-to-page index, which lives only in the graph projection, so a
+target that exists in revision slots but is missing from the graph is reported as not found. An
+import populates the slots without updating the projection
+([#1022](https://github.com/ProfessionalWiki/NeoWiki/issues/1022)), so imported targets warn until
+[the graph is rebuilt](../operations/maintenance.md#rebuilding-the-graph). Being non-blocking, this
+misreport never costs a write.
+
 ## Known limitations (Foundation round)
 
-The PHP `SubjectValidator` performs two Subject-level checks:
+The PHP `SubjectValidator` performs these Subject-level checks:
 
 - **`required`** — PHP iterates the Schema's properties to catch absent-required cases.
 - **`type-mismatch`** — PHP compares the Statement's writer's-schema type against the Schema's
   current type and surfaces drift per ADR 11 / ADR 12.
+- **[`relation-target-schema-mismatch`](#relation-target-schema-mismatch)** and
+  **[`relation-target-not-found`](#relation-target-not-found)** — documented above; both need a
+  Subject lookup, which `PropertyType::validate()` has no access to.
 
 ### Deliberate behavior, not a gap
 
--- a/docs/api/validation-codes.md
+++ b/docs/api/validation-codes.md
@@ -204,7 +204,7 @@ Reconciling that asymmetry is left to the enforcement tier (ADR 21).
 Emitted by `SubjectValidator`.
 `args`: `[expectedSchema, actualSchema]` — the target Schema the relation Property declares
 (`targetSchema`), and the Schema the resolved target Subject actually uses.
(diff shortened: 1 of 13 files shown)


docs/api/validation-codes.md
```diff
@@ -227,13 +228,23 @@ hardcoded non-blocking set, so it never rejects a write even under enforcement.
 pointing at a not-yet-created Subject is wiki-native red-link behavior, and an import may legitimately
 mint the target later.
 
+Resolution goes through the subject-to-page index, which lives only in the graph projection, so a
+target that exists in revision slots but is missing from the graph is reported as not found. An
```
JeroenDeDauw, line 232: @alistair3149 does this belong in the docs? Looks like we're now explaining a bunch of implementation details.

alistair3149: We can simplify it to `A a target that exists in revision slots but is missing from the graph is reported as not found`. It is a confusing behavior that should be documented. That will likely be solved with #1040.

JeroenDeDauw: I'm currently in a fierce debate with fable about this text. Its first take is to say the PR is good.

JeroenDeDauw: Ok this took some doing but it now agrees, FWIW

image

JeroenDeDauw: Deslop:
* https://github.com/ProfessionalWiki/NeoWiki/pull/1106

Issue #1087, opened 2026-07-19
Add subject-level RDF export (per-subject RDF API)


JeroenDeDauw:
The only RDF export today is per-page: `GET /rest.php/neowiki/v0/page/{pageId}/rdf` (`src/EntryPoints/REST/ExportPageRdfApi.php`), which dumps every sibling Subject on a page plus page metadata as one document. There is no way to get RDF for a single Subject — the subject route `GET /neowiki/v0/subject/{subjectId}` returns JSON only.

## Why per-subject RDF is worth having

- **Named deliverable.** WP3 T3.2 calls for web APIs for subject-level, bulk-subject, and statement-level access with native RDF output. Page-level export doesn't cover subject-level RDF.
- **Subject is the natural RDF granularity.** Every Subject already has a stable `neo-subj:` IRI. A Linked Data consumer that encounters that IRI (e.g. from another dataset) wants to resolve *that entity's* description, without first discovering which MediaWiki page hosts it. Page-level granularity forces a `subject-IRI → page-ID → page-RDF` indirection an off-wiki client can't perform.
- **Multi-subject pages.** When a page holds several independent Subjects, the page dump is coarser than a consumer wanting one entity needs.

## Proposed shape (open)

Serve RDF for one Subject by reusing the existing native + ontology-mapping projectors filtered to that Subject's triples. Keep parity with the page endpoint's `format` (turtle/trig) and `projection` (`native`, `edm`, …) parameters.

Two ways to expose it, to be decided:

- Content-negotiate / add a `format` parameter to the existing `GET /neowiki/v0/subject/{subjectId}` route (`Accept: text/turtle` / `application/trig`), or
- A dedicated `GET /neowiki/v0/subject/{subjectId}/rdf`.

## Open design questions

- **Bounded-description semantics.** Emit only triples where the Subject is the subject (outbound / CBD), or also inbound relations pointing at it? Default: outbound-only, matching what the page projection already emits per subject.
- **TriG / named graph.** The page endpoint wraps triples in a per-page named graph. For a single Subject, is Turtle the primary format and TriG secondary (or omitted), or does TriG still carry the hosting page's graph IRI?
- **Relation targets** stay as IRIs, not inlined (standard LD).

## Scope

API endpoint only. Full HTTP dereferencing of `neo-subj:` IRIs (routing the IRI itself to this response) is a larger deployment concern and out of scope here.

[...]

Issue #1023, opened 2026-07-11
Basic RDF projection and ontology mapping: feedback-ready milestone (tracking)


JeroenDeDauw:
Tracking issue for implementing the published RDF designs far enough that ECHOLOT partners can try them end to end and give concrete feedback: the native RDF projection synced to a SPARQL store, plus a first, deliberately minimal ontology-mapping capability projecting to EDM.

Designs: [NativeRdfProjection.md](https://github.com/ProfessionalWiki/NeoWiki/blob/master/docs/planning/NativeRdfProjection.md), [OntologyMapping.md](https://github.com/ProfessionalWiki/NeoWiki/blob/master/docs/planning/OntologyMapping.md), [ADR 19](https://github.com/ProfessionalWiki/NeoWiki/blob/master/docs/adr/019-graph-database-architecture.md). Feedback threads: #996, #999.

## Phases

- [ ] #1016 — register graph-database surfaces only when configured (part of #895)
- [ ] #1024 — projection-failure and degradation semantics (part of #895)
- [ ] #1025 — native RDF projection and RDF export
- [ ] #586 — SPARQL store plugin syncing the projection to QLever / any SPARQL 1.1 store
- [ ] #1026 — Ontology Mapping v1 (term substitution, EDM tier)
- [ ] #1027 — end-to-end exercise: person toy model → EDM, demo + feedback round

#1016, #1024 and #1025 are independent of each other; #586 builds on all three; #1026 builds on #1025; #1027 closes the loop.

## Milestone cut line

In scope: per-page named-graph sync to a SPARQL 1.1 store (QLever in the dev stack), per-page RDF export and a bulk dump, Mapping pages with term-substitution rules, per-store projection selection, native + EDM sibling stores demoable over the same wiki, docs updated to as-built.

Deliberately out, as follow-ups: the `{{#sparql}}` / REST query surface, RDF import, intermediate-node synthesis (the CIDOC-CRM tier), SHACL shape generation/validation, RDF-star.

The v1 mapping format will be versioned and explicitly provisional: the mapping-formalism question ([OntologyMapping.md](https://github.com/ProfessionalWiki/NeoWiki/blob/master/docs/planning/OntologyMapping.md) Q1, #995) stays open for partner input.

> Phased plan and scope decisions reviewed and approved by @JeroenDeDauw in-session; issue text written autonomously and posted without human review.
[...]

PR #1088, opened 2026-07-19
Add a View RDF link to the page-tools sidebar

JeroenDeDauw:
The per-page RDF export (GET /rest.php/neowiki/v0/page/{pageId}/rdf) was only
reachable by hand-constructing the URL. Add a "View RDF" link to the NeoWiki
page-tools sidebar section, next to "View JSON", pointing at the current page's
Turtle export.

It is shown only when the page has Subjects, so it never points at the
endpoint's 404-on-empty response. It uses ?format=turtle so the result renders
as readable text in the browser instead of downloading a TriG file, and targets
the native projection. Per-ontology projection links and a per-subject RDF
endpoint (#1087) are out of scope here.

Relates to #1023

> <sub>AI-authored — Claude Code, `Opus 4.8 (max)`; @JeroenDeDauw asked for this sidebar link and to build it via the fix-issue workflow (no revisions during implementation); not yet human-reviewed; new PageToolsBuilder tests + full `make cs` (phpcs + phpstan) green, and verified live on the demo wiki — "View RDF" renders on a Subject page and its target returns `text/turtle`, and is absent on a Subject-less page. Full local PHPUnit is green apart from two pre-existing SPARQL end-to-end tests that need the QLever service (absent from this dev stack, unrelated to this change); CI covers them.</sub>

### DIFF
--- a/i18n/en.json
+++ b/i18n/en.json
@@ -42,6 +42,7 @@
 	"neowiki-page-tools-view-subjects": "View subjects",
 	"neowiki-page-tools-edit-json": "View or edit JSON",
 	"neowiki-page-tools-view-json": "View JSON",
+	"neowiki-page-tools-rdf": "View RDF",
 
 	"neowiki-property-editor-name": "Name",
 	"neowiki-property-editor-type": "Type",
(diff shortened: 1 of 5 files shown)


JeroenDeDauw:
image

PR #1099, opened 2026-07-20
Guard the sidebar RDF link against a missing graph backend

malberts:
Follows-up to https://github.com/ProfessionalWiki/NeoWiki/pull/1088

#1088's new "View RDF" sidebar link gates on `PageToolsBuilder`'s `hasSubjects` flag, which `onSidebarBeforeOutput` computes by building the Subjects lookup. That eagerly constructs the Neo4j client and throws `GraphBackendNotConfiguredException` when no backend is configured, so every content-namespace page view 500s on a backend-less wiki. CI runs with a backend, so it stays green.

Guard the lookup on `getNeo4jPlugin() !== null`, mirroring the no-backend guard #1016 added to `handleContentPage`. A new `NeoWikiSidebarLinkTest` case builds a content page's sidebar under `runWithoutGraphBackend`; it throws without the guard and passes with it.

> <sub>AI-authored — Claude Code, `Opus 4.8 (1M context)`; review of @JeroenDeDauw's #1088 requested by @malberts via /pr-review, blocker found and fixed autonomously (no human steering during the fix); not yet human-reviewed; regression test written failing-first then green, full local PHPUnit (1795 tests) + phpcs + phpstan pass, and the 500 reproduced then confirmed fixed live on a no-backend dev wiki.</sub>

### DIFF
--- a/src/EntryPoints/NeoWikiHooks.php
+++ b/src/EntryPoints/NeoWikiHooks.php
@@ -273,12 +273,17 @@ class NeoWikiHooks {
 			->getNamespaceInfo()
 			->isContent( $title->getNamespace() );
 
+		// Building the Subjects lookup eagerly constructs the Neo4j client, which throws without a
+		// configured backend; guard so content pages still render backend-less (cf. handleContentPage).
+		$hasSubjects = $isContentNamespace
+			&& $extension->getNeo4jPlugin() !== null
+			&& $extension->newPageSubjectsLookup()->pageHasSubjects( $pageId );
+
 		$neoWikiTools = ( new PageToolsBuilder() )->build(
 			title: $title,
 			pageId: $title->getArticleID(),
 			isContentNamespace: $isContentNamespace,
-			hasSubjects: $isContentNamespace
-				&& $extension->newPageSubjectsLookup()->pageHasSubjects( $pageId ),
+			hasSubjects: $hasSubjects,
 			canCreateMainSubject: $hints->canCreateMainSubject( $pageId ),
 			canEditSubject: $hints->canEditSubject( $pageId ),
 			isLatestRevision: self::pageIsLatestRevision( $skin->getOutput() ),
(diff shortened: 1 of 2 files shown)


src/EntryPoints/NeoWikiHooks.php
```diff
@@ -273,12 +273,17 @@ public static function onSidebarBeforeOutput( Skin $skin, array &$sidebar ): voi
...
+			&& $extension->getNeo4jPlugin() !== null
+			&& $extension->newPageSubjectsLookup()->pageHasSubjects( $pageId );
+
 		$neoWikiTools = ( new PageToolsBuilder() )->build(
 			title: $title,
 			pageId: $title->getArticleID(),
 			isContentNamespace: $isContentNamespace,
-			hasSubjects: $isContentNamespace
-				&& $extension->newPageSubjectsLookup()->pageHasSubjects( $pageId ),
+			hasSubjects: $hasSubjects,
```
JeroenDeDauw, line 286: This code doesnt make sense. Also why would it just chek neo4j? I suspect the checks are also happening on the wrong level or wrong place.

JeroenDeDauw (changes requested):

src/EntryPoints/NeoWikiHooks.php
JeroenDeDauw: Just realized this is about the **RDF** link. So the Neo4j check makes even less sense.

malberts: The idea was to match the guard in https://github.com/ProfessionalWiki/NeoWiki/pull/1016/changes#diff-a9fe0c2c52bc8ddb5c946a0a151e1eebdac7975c68a9237dd0e0597dfd159dc1R51-R59.

Currently, the `newPageSubjectsLookup()` constructor eventually calls `new Neo4jPageIdentifiersLookup( $this->getReadOnlyNeo4jClient() )`. So if Neo4j is missing today, then the RDF link generation blows up too. I suppose this minor risk is fine while graph backend work is ongoing.

But yes, hardcoding the Neo4j check is not the right way.

PR #1105, opened 2026-07-20
Add Special:Mappings special page listing ontology Mappings

JeroenDeDauw:
Adds a `Special:Mappings` page that lists and manages the ontology Mapping pages (the `Mapping:` namespace), mirroring `Special:Schemas` and `Special:Layouts`. There is no tracking issue; this description is the spec.

## What this adds

- **`Special:Mappings`** (mount div `ext-neowiki-mappings`), registered in `extension.json` and `i18n/_Aliases.php`.
- **`GET /neowiki/v0/mappings`**: paginated (`limit`/`offset`) list returning `{mappings: [{name, schemas}], totalRows}`, where `schemas` is the names of the Mapping's mapped Schemas, sorted alphabetically server-side.
- **`MappingsPage.vue`**: a Codex table with a **Name** column (linking to `Mapping:<name>`) and a **Mapped schemas** column listing each mapped Schema (linked to its `Schema:` page) followed by the count, e.g. `Artist, Artwork, City, Person (4)`; a Mapping with no mapped Schemas shows `None`. Pagination and an empty state.
- **Create / Edit / Delete**:
  - A **Create** button (gated on `neowiki-mapping-edit`) opens a name-only dialog that creates `Mapping:<name>` with an empty mapping skeleton and opens the new page for editing. Pressing Enter in the name field submits. The reserved name `native` is rejected inline (the backend also rejects a `Mapping:Native` save, so this is a fast inline check rather than the sole guard).
  - Per-row **Edit** (gated on `neowiki-mapping-edit`; opens the page's raw-JSON edit view) and **Delete** (gated on the core `delete` right; confirmation dialog).
- **"All mappings" sidebar link** on Mapping-namespace pages (`t-neowiki-mappings`), parallel to the Schema/Layout links.
- **Messages** in `i18n/en.json` + `i18n/qqq.json` (the frontend ones also registered in the ResourceLoader module).
- **Discovery links**: the Developers Browse list, the RDF-and-ontology-mappings demo page, `Main_Page`'s RDF bullet, `docs/api/rest-api.md`, and a one-line mention in `docs/rdf/ontology-mapping.md`.

## Judgment calls

- **The API returns Schema names, not a count.** The mapped-schemas cell shows which Schemas a Mapping covers; the count is the array length. Sorting is done server-side so the view carries no ordering logic.
- **In-handler pagination.** `MappingNameLookup` exposes only `getMappingNames()` (no SQL limit/offset/count) and is shared with the RDF projection path, so the handler slices the (one-page-per-ontology, small) name list and counts it for `totalRows`. Endpoint behaviour matches the Schema/Layout summaries APIs.
- **Edit navigates; Create seeds and navigates.** Mapping pages are raw JSON with no editor UI, so Edit opens the standard `action=edit` view, and Create writes an empty skeleton and lands on the new page for the user to fill in.
- **Create uses `mw.Api` `action=edit` with `createonly`.** This is atomic and surfaces a "page exists" error as an inline name-taken message; a plain create-or-update save would overwrite an existing mapping.
- **Delete is gated on the core `delete` right** (which alone authorizes page deletion), unlike the Schema/Layout pages which gate delete on their entity-edit right; the same sibling inconsistency is tracked in #1114.

## Verification

`make cs`, `make phpunit`, and the TypeScript gates (vitest, build, lint) pass. Verified live on the dev wiki: the endpoint returns the EDM mapping's four sorted Schema names; the page renders them as comma-separated links with the count, and a zero-Schema mapping as `None`; creating a mapping (via the button or Enter) writes a `NeoWikiMapping` skeleton page and lands on it; `native` is rejected inline; Edit opens the JSON edit view; and the actions column is right-aligned. Delete gating checked with both a non-sysop user (Create and Edit shown, Delete hidden) and a sysop (all three shown, delete succeeds).

> <sub>AI-authored — Claude Code, `Opus 4.8 (max)`; detailed spec from @JeroenDeDauw plus several mid-run refinements (Schema-name column, row Create/Edit/Delete actions, UI-bug fixes, delete-right gating); diff not yet human-reviewed; phpcs/phpstan + PHPUnit + vitest + build/lint green locally, live-verified end-to-end (incl. non-sysop vs sysop gating) on the dev wiki, CI green.</sub>


JeroenDeDauw:
image

JeroenDeDauw:
image

PR #1109, opened 2026-07-20
Drive JSON-namespace content-model lock from a single map

JeroenDeDauw:
NeoWiki's Schema, Layout and Mapping namespaces are each locked to a single
JSON content model. The lock was enforced in two hand-maintained places: the
ContentModelCanBeUsedOn hook, which enumerated the three namespace/model pairs
by hand, and each content handler's canBeUsedOn(). It had no test coverage, so
adding a fourth JSON namespace risked silently shipping it unprotected.

Introduce NamespaceContentModels as the single source of truth mapping each
namespace to its content model, and drive both the hook and the handlers from
it. Behaviour is unchanged: the hook and handlers now read the lock from one
place instead of several hand-kept lists, so they cannot drift out of sync.

Add data-driven tests that iterate the map and assert, for each namespace,
that the matching model is allowed and foreign models (wikitext and other
NeoWiki models) are rejected — covering the hook directly and the real save
path — plus a test pinning the concrete namespace/model pairs.

> <sub>AI-authored — Claude Code, `Opus 4.8 (max)`; detailed spec from @JeroenDeDauw, built autonomously (no interactive human steering); diff not yet human-reviewed; TDD + mutation-checked, independently AI-reviewed (one test-quality fix applied), full PHPUnit + phpcs + phpstan and CI green.</sub>

### DIFF
--- a/src/EntryPoints/Content/LayoutContentHandler.php
+++ b/src/EntryPoints/Content/LayoutContentHandler.php
@@ -12,7 +12,6 @@ use MediaWiki\Content\ValidationParams;
 use MediaWiki\Title\Title;
 use MediaWiki\Parser\ParserOutput;
 use ProfessionalWiki\NeoWiki\Domain\Layout\LayoutName;
-use ProfessionalWiki\NeoWiki\NeoWikiExtension;
 use ProfessionalWiki\NeoWiki\Persistence\MediaWiki\LayoutContentValidator;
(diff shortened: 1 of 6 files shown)


JeroenDeDauw:
Independent Fable max insists this is good, but I think it's nonsense adding accidental complexity

PR #1123, opened 2026-07-21
Add deep-linkable Subject rows and an optional Data-tab dereference target

JeroenDeDauw:
Follows-up to https://github.com/ProfessionalWiki/NeoWiki/pull/1107

Builds two related additions on top of subject-IRI dereferencing (#1107).

### Deep-linkable Subject rows (Data tab)

Arriving on `?action=subjects` with a Subject id as the URL fragment (`#sEpfwJLAtndCccC`) expands that Subject's row, scrolls it into view, and highlights it. This runs on app mount: the Vue rows mount after page load, so native fragment navigation never reaches them. Manually expanding a row writes its Subject id to the address-bar fragment via `history.replaceState` (no new history entry, no `hashchange`), so the address bar is a shareable deep link; collapsing leaves the fragment in place. An unknown or absent fragment changes nothing.

### Data-tab dereference target

New per-wiki `$wgNeoWikiSubjectDereferenceTarget` (`page` default, or `data-tab`). When `data-tab`, the dereference endpoint's HTML branch (a browser `Accept`) redirects to `{canonicalPageUrl}?action=subjects#{subjectId}` instead of the plain page, landing the reader on the Subject's row. The RDF branches are unchanged. An unrecognized value raises a `ConfigException` rather than silently falling back.

### Fragment contract (public)

The Data tab URL fragment is the **bare Subject id** (`#sEpfwJLAtndCccC`), like Wikibase's `#P123`. The dereference endpoint's `data-tab` Location and manual row-expansion both write it, and the Data tab reads it back on mount. The row's DOM id (`ext-neowiki-subject-row-{subjectId}`) is an internal rendering detail, decoupled from the fragment — the mount handler resolves the Subject id to a row itself, so the fragment need not equal a DOM id. Fragment parsing lives in a pure, vitest-covered TS helper (`subjectIdFromHash`); the component only wires it up.

### Tests

Handler integration for both config values and the unrecognized-value error (with a `page` baseline in setUp so the Accept-negotiation tests are config-independent); vitest for the fragment parser, including junk fragments and rejecting the old long form; the existing conneg matrix and drag spec stay green. Full PHPUnit (1937) and vitest (1056) suites plus phpcs/phpstan pass locally; live-verified in the dev wiki (curl + Playwright, logged-out).

### Considered, omitted

- Fragment = row DOM id (an earlier revision of this PR): dropped for the bare Subject id — the mount handler resolves the id to a row itself, so the fragment never needed to equal a DOM id. The old long form is not accepted.
- Class-based highlight, not `:target`, since rows mount after fragment navigation.
- No new UI chrome (no "copy link" button); the shareable link is the address bar itself.

> <sub>AI-authored — Claude Code, `Opus 4.8 (max)`; detailed spec from @JeroenDeDauw, redirected once by his review (fragment → bare Subject id); revised diff not yet re-reviewed; full PHP + TS suites and phpcs/phpstan green locally, live-verified via curl + Playwright; CI green.</sub>


JeroenDeDauw:
Locally tested

PR #1133, opened 2026-07-22
Refine Data-tab subjects UI and fix the missing export icon

alistair3149:
UX refinements to the Data-tab subjects-management UI (`?action=subjects`), plus a real runtime-crash fix uncovered along the way.

## Changes

- **Fix the missing export icon (latent crash from #1097).** The export buttons import `cdxIconDownload`, but it was never added to the `ext.neowiki` icons allowlist (`extension.json`, the `callbackParam` feeding `CodexModule::getIcons`). Because the bundle externalizes Codex and gets icons from that server-generated `icons.json`, the icon resolved to `undefined` at runtime, `CdxIcon` threw `Cannot use 'in' operator to search for 'path' in undefined`, and the render error took out the **entire** export button — no icon, no button, console errors. Adding it to the allowlist fixes all of it. I diffed every `@wikimedia/codex-icons` import in `src` against the allowlist: `cdxIconDownload` was the only gap (the `cdxIcon*` in `src/assets/CustomIcons.ts` are local inline SVGs, not Codex, so they need no entry).
- **"Export all" label for the page-level export button.** New message `neowiki-managesubjects-export-all-button` distinguishes the whole-page export from the per-Subject "Export" buttons, which keep `neowiki-managesubjects-export-button` (its qqq is narrowed to per-Subject).
- **ID/IRI identifiers footer → description list.** The identifiers block is now a `<dl>` (each pair grouped in a `<div>`; labels `<dt>`, copy-button values `<dd>`), which is the real term/value content. The row itself stays a `<details>`/`<summary>` disclosure — deliberately not a `<dl>`.
- **Scope the small/subtle typography to the identifiers.** Moved `font-size`/`color` off `__row-footer` onto `__row-identifiers`, reset the default `<dl>`/`<dd>` margins, and dropped the identifiers `gap` (line-height carries the spacing). Side effect (intended): the export button no longer inherits the footer's x-small size.

`resources/ext.neowiki/dist/` is gitignored, so this diff is source-only; rebuild the bundle with `make ts-build`.

## Manual Browser Check

On a page with subjects (e.g. the Rijksmuseum demo) at `?action=subjects`, after `make ts-build` + a hard refresh:

1. The top-right export control shows a **download icon** and reads **"Export all"** (before this fix it failed to render at all).
2. Open the console — **no** `CdxIcon` / "Cannot use 'in' operator" errors.
3. Expand a Subject row. In the footer, **ID:** and **IRI:** render as a labeled list, small and subtle, on their own lines with comfortable spacing.
4. Click the ID and IRI copy buttons — each copies and shows a success notification (aria-labels intact).
5. The per-Subject **Export** button in the footer is normal-sized (not shrunk to the ID/IRI text size), and still opens the format menu.

## Verification

- `make ts-lint` clean, `make ts-build` succeeds, `make ts-test filter=SubjectsManager` 5/5.
- Live-verified on the dev wiki: console clean; identifiers render as `DL`/`DT`/`DD`; typography moved (identifiers 12px subtle, export button back to 14px); `<dd>`/`<dl>` margins zeroed.
- Reviewed by a code-reviewer subagent: no Critical/Important issues.

<sub>AI-authored — Claude Code, `Opus 4.8 (1M context)`; ad-hoc UX refinements requested by @alistair3149 across one session, which surfaced the #1097 icon-allowlist crash; ts-lint + ts-build + vitest (SubjectsManager 5/5) green locally, live-verified in-browser on the dev wiki; reviewed by a code-reviewer subagent (no blocking issues). CI not yet run.</sub>


alistair3149:
image

Issue #1063, opened 2026-07-16
Parse-path Schema/Mapping reads vary by user but the parser cache does not (ADR)


alistair3149:
Pre-existing structural tension, flagged (not caused) by the #1046 gate work (PR #1058). ADR-shaped; recording so it isn't lost.

`CachingSchemaLookup` / `CachingMappingLookup` gate reads with `authorizeRead('read', $title)` and are reachable from **page parsing** — the Lua library (`nw` schema access) and, transitively, anything that reads Schema/Mapping content during parse. But MediaWiki's parser cache key does **not** vary on user identity (`ParserOptions` tracks dateformat, thumbsize, userlang, etc. — there's an explicit `@todo Track this for caching somehow without fragmenting the cache`), and NeoWiki registers no parser option and never calls `disableCache`.

So a per-user gate on a parse-reachable lookup creates a latent cache-coherency problem:

- **Leak:** a privileged first parse bakes restricted values into the user-agnostic cache; unprivileged readers get the cached page.
- **Silent breakage:** an anon/job first parse gets a gated `null`, caches empty output, and privileged users then see silently-empty pages.

`{{#view}}` is immune by construction (it emits a client-side placeholder div; Subject/Layout data is fetched per-user over REST — the model the rest of #1046 relies on), but the Lua/parser-function reads are not.

This PR's gate upgrade (probablyCan→authorizeRead) neither created nor worsened this — the pre-existing `probablyCan` gate already varied by user — but it's the moment to write the invariant down. Likely direction: pin parse-path reads to a fixed **public** authority so parser-cached output is authority-independent by construction (making today's "uniformly public parse output" intentional rather than accidental), and keep per-user gating strictly on the REST/client-fetch path. Wants an ADR.

Also in this bucket (same root cause, pre-existing): `Neo4jQueryLimits::forUser()` in the Cypher parser function / Lua runner already varies parse output by `apihighlimits` and writes it to the user-agnostic cache.

> Split out of the #1046 review (PR #1058). Directed by @alistair3149.
> <sub>Written by Claude Code, `Fable 5`</sub>

JeroenDeDauw:
Related:
* https://github.com/ProfessionalWiki/NeoWiki/issues/1059#issuecomment-5078718794

JeroenDeDauw:
Consolidated into #1059: any permission gate on a parse-time surface immediately raises the cache question recorded here, so the two resolve as one. Substance folded into that issue's body.

Issue #1059, opened 2026-07-16
Parse-time surfaces: permission checks and parser-cache semantics


alistair3149:
Parse-time surfaces have permission gaps — the raw query functions let anyone who can get wikitext parsed read the whole graph — and any fix interacts with MediaWiki's user-agnostic parser cache: parse output that varies by the requesting user either leaks via the shared cache or breaks caching. Permission gates and cache behavior therefore resolve together; this issue also carries what was #1063.

Current behavior per surface:

* `{{#cypher_raw}}`, `{{#sparql_raw}}`, `nw.query`, `nw.sparqlQuery`: no check at all — no right, no per-page filter; a preview suffices, no save needed. This bypasses the per-page gates from #1046; the REST query endpoints, by contrast, require the `neowiki-query` right. Their result-size limit (`Neo4jQueryLimits::forUser()`) already varies parse output by `apihighlimits`.
* `{{#neowiki_value}}` and the `nw` data accessors: only revision-deletion visibility is honored; the page's `read` permission is not checked.
* Schema/Mapping lookups (`CachingSchemaLookup`/`CachingMappingLookup`): gated per user (`authorizeRead`) but reachable from parse — a privileged first parse bakes restricted values into the shared cache; an anon/job first parse caches silently-empty output for everyone.
* `{{#view}}`: leak-free by construction — parse emits a placeholder; data is fetched per user over REST.

To decide: what may vary by the parsing user, and what that means per surface — add a per-user gate, evaluate as a fixed public authority (cached output becomes user-independent), disable the surface, or make it configurable. Product input so far: the stakeholder deployment driving MVP scope disables the parser cache anyway, leaving little MVP reason for cache-safety machinery ([comment](https://github.com/ProfessionalWiki/NeoWiki/issues/1059#issuecomment-5078718794)).

The decision lands in the access-control ADR's parse-time TODO ([#1134](https://github.com/ProfessionalWiki/NeoWiki/pull/1134)); this issue stays open for the implementation.

> <sub>AI-authored — Claude Code, `Fable 5`; split out of the #1046 review, directed by @alistair3149; rewritten 2026-07-25 to absorb #1063 (`Fable 5 (max)`, directed by @JeroenDeDauw); final wording not human-reviewed; claims cross-checked against both prior bodies and the #1058 review.</sub>

JeroenDeDauw:
Same for `{{#neowiki_value}}`

[...]

PR #1134, opened 2026-07-22
Add ADR 27: Access Control

JeroenDeDauw:
Records the access-control model the recent permission work implements: MediaWiki as the sole
permission authority, the page as the unit of access control (per-subject ACLs are an explicit
non-goal), denial as absence, per-row filtering for page-attributable results, scoping keys (not
ACL state) in graph projections, whole-store read semantics for raw queries with server-side
filter injection to come, and no permission checks in projection or dump generation (their output
can contain restricted content and must not be exposed past its audience).

Still open, recorded as explicit TODOs: the parse-path rule, cross-wiki subject display, and the
default grant of `neowiki-query`.

Related: #1046, #1059, #1063.

> <sub>AI-authored — Claude Code, `Fable 5 (max)`; directed ask from @JeroenDeDauw, with several in-session review rounds steering structure and content, plus his own edits on the branch; draft read by @JeroenDeDauw, no formal review yet; claims checked against master code/docs and the issue tracker; the per-subject-ACL non-goal and the projection decision were each backed by a dedicated sweep of the requirements record for contrary demand (none found); CI green.</sub>

### DIFF
--- a/docs/adr/027-access-control.md
+++ b/docs/adr/027-access-control.md
@@ -6,9 +6,9 @@ Status: Draft
 
 ## Context
 
-NeoWiki data leaves the system through several kinds of surface: REST lookup endpoints, the REST query endpoints that
-execute a caller-supplied Cypher or SPARQL query ([query-api.md](../api/query-api.md)), parse-time accessors (parser
-functions and Lua), RDF export, and projection into graph and SPARQL stores.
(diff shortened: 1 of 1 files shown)


JeroenDeDauw:
The **Open decisions** section is meant to be resolved before we undraft the ADR, with entries either going to the **Decision** (and perhaps **Consequences**) section, or going out of scope of the ADR with perhaps tracking via dedicated issues.

PR #1137, opened 2026-07-23
Redesign the Data-tab export control as a format-first two-level menu

alistair3149:
Redesigns the Data-tab export control (`?action=subjects`), following up on #1133.

## Why

The export control was a quiet `CdxMenuButton` with two problems:

- **The menu didn't scale.** It listed a flat `format × projection` set (JSON + Turtle/TriG *per* projection = `2N+1` items). Projections are `native` plus one per ontology Mapping — a user-grown, permission-filtered set — so a dozen mappings reached ~25 flat items.
- **Quiet weight, invisible hover, no normal-weight option.** `CdxMenuButton` hardcodes a quiet trigger, and its quiet hover washed out against the footer's background.

## What

A reusable **`DataExportButton`**: a normal-weight framed `CdxButton` (visible hover) plus a manually-controlled two-level `CdxMenu` (`useFloatingMenu`), bounded to the trigger width:

- **Level 1 — serialization:** `JSON`, `Turtle ›`, `TriG ›`. JSON downloads immediately; Turtle/TriG (marked with a trailing chevron) reveal the projection list without navigating.
- **Level 2 — projection:** a `‹ Back` item plus one entry per projection (`Native` + each mapping). Selecting one downloads that projection in the chosen serialization (new tab, `noopener`).

Decoupling the two axes makes the list `N`, not `2N`, and the projection list appears only after a deliberate RDF-format choice. `DataExportMenu.ts` was refactored from flat `MenuButtonItemData[]` builders to `{ jsonUrl, rdfUrl(projection, format) }` URL builders.

Also in this change:
- **Keyboard:** Enter/Space open, arrow/Home/End open-and-highlight, Enter/Space select or advance a level, Escape closes. A native button's `detail`-0 activation click is ignored so keyboard activation never double-fires (which previously flashed the menu open→closed).
- **Screen readers:** an `aria-live` region announces the in-place level change ("Choose a projection" / "Choose a format").
- Registered `CdxMenu` / `useFloatingMenu` / `useGeneratedId` in the `ext.neowiki` `codexComponents` array and `cdxIconDownload` / `cdxIconNext` / `cdxIconPrevious` in the icons allowlist (these throw "not listed" at runtime otherwise — the same registration class as the icon allowlist in #1133).
- Removed the now-unused templated `…-export-turtle` / `…-export-trig` messages.
- Driveby: tightened the subjects-manager control-actions gap to `@spacing-75`.

## Manual Browser Check

On a page with subjects and at least one ontology Mapping (e.g. the Rijksmuseum demo) at `?action=subjects`, after `make ts-build` + a hard refresh:

1. The top **"Export all"** and each per-Subject footer **"Export"** button are **framed** (normal weight) with a **visible hover**.
2. Open one (click, or focus + Enter/Space): a **bounded dropdown** (menu width ≈ button, not the full row) shows **JSON**, **Turtle ›**, **TriG ›**.
3. **JSON** downloads (new tab) in one action.
4. **Turtle** (or TriG) swaps to the projection list — `‹ Back` plus `Native` and each mapping — without navigating or closing.
5. Pick a projection → it downloads that projection in the chosen serialization; the menu closes.
6. Reopen → back at level 1 (not the stale projection list).
7. **Keyboard:** Tab to the button; Enter/Space open (no flash); ArrowDown/Up highlight; Enter selects/advances; `‹ Back` returns to level 1; Escape closes.
8. Console is clean (no `codexComponents` / `CdxIcon` errors).

## Verification

- `make ts-lint` clean, `make ts-build` green.
- Full `make ts-test` suite green: **1127/1127**. `DataExportButton` **9/9** (including Enter-opens, Escape-closes, and a flash-guard test); `DataExportMenu` and `SubjectsManagerPage` suites green.
- End-to-end browser check on the dev wiki — mouse and keyboard: framed buttons with visible hover, bounded menu, chevron affordances, correct download URLs, Back + reset-on-reopen, and the aria-live level announcements firing at each transition.
- Two independent code reviews (on the pre-refinement state): merge-ready, no blocking issues.

## Follow-up (not in this PR)

A structural test diffing `src`'s Codex imports (components + icons) against `extension.json`, so a missing `codexComponents` / icon-allowlist entry fails CI instead of only at runtime. The build and unit tests can't catch it (Codex is externalized; tests resolve the full `node_modules` copy).

<sub>AI-authored — Claude Code, `Opus 4.8 (1M context)`; brainstormed the direction with @alistair3149 (format-first two-level menu; export dialog deferred as YAGNI), implemented via subagent-driven-development (per-task implementer + two-stage spec/quality review), then refined over several rounds of @alistair3149's manual review (bounded menu width, chevron affordance, matching Back chevron, keyboard handling incl. the Enter flash, screen-reader announcement); full `ts-test` (1127) + `ts-lint` + `ts-build` green locally, end-to-end browser-verified (mouse + keyboard) on the dev wiki, two independent code reviews merge-ready. CI not yet run.</sub>


alistair3149:
image

PR #1144, opened 2026-07-23
Add a "Copy link" row action to the Data tab

JeroenDeDauw:
Adds a "Copy link" action to each Subject row on the Data tab (`?action=subjects`) that copies a deep link to that row — the page URL with the Subject id as the fragment — to the clipboard. It completes the deep-linkable rows from #1123, which offered no UI to obtain such a link.

The row controls have two responsive halves, so the action is added to both: an inline icon button in the action cluster at desktop/tablet widths (hover-revealed, like the existing edit/delete buttons) and an item in the overflow (⋯) menu on narrow screens. Neither is permission-gated — every user, including read-only ones whose cluster and menu were otherwise empty, gets the affordance at every width.

Expanding a row now keeps its controls visible rather than hover-revealed, at both widths (the inline cluster on desktop, the ⋯ button on narrow screens), via a new `--expanded` row modifier that joins the existing `--highlighted` reveal rule. Expansion is a user-initiated focus on the row, and persistent controls aid discoverability — most for read-only users, whose only control is Copy link. Focus reveals the controls only for keyboard focus (`:has( :focus-visible )` rather than `:focus-within`): a mouse click — on a copy button, or on the summary when toggling a row — no longer pins them visible after the pointer leaves.

- Both orderings keep the destructive Delete last: promote, edit, copy link, delete on other rows; edit, copy link, delete on the main row.
- The link is built from `location.href` with the fragment set to the bare Subject id, so it works unchanged under both default and pretty (`$wgActionPaths['subjects']`) URL styles. Copying does not mutate the address bar, row expansion, or highlight.
- Success and failure show a notification, mirroring the existing copy-id and copy-iri helpers; the success toast omits the (long) URL.

Three shared messages (`neowiki-managesubjects-row-copy-link`, `-link-copied`, `-link-copy-error`) are added to `en.json`, `qqq.json`, and the `ext.neowiki` module.

Vitest specs cover, for a read-only user, the menu item and the inline button on both row types; that triggering copy link (via the menu and via the inline button) writes a URL whose fragment is the Subject id and shows the success toast; that a clipboard rejection shows the error toast; and that the `--expanded` modifier class is present only while a row is open. No PHP behavior changes.

> <sub>AI-authored — Claude Code, `Opus 4.8 (max)`, final qqq touch-up by `Claude Fable 5 (max)`; detailed spec from @JeroenDeDauw, two follow-up corrections (add the desktop-inline affordance, since the ⋯ menu alone is mobile-only; then keep controls visible on expanded rows); not yet human-reviewed; vitest (1136), PHPUnit (1950), phpcs/phpstan and eslint green locally and in CI, new tests mutation-/fail-first-checked, browser-verified at desktop and narrow widths, logged out and as sysop.</sub>


JeroenDeDauw:
image

PR #1147, opened 2026-07-23
Add per-Constraint validation severity levels (ADR 26, backend)

alistair3149:
Implements the backend half of [ADR 26](https://github.com/ProfessionalWiki/NeoWiki/blob/master/docs/adr/026-validation-severity-levels.md): every Constraint carries a severity (`error` / `warning`), and enforcement blocks only when a write introduces a **new** `error` violation.

## What changes

A Constraint is now written either as the bare scalar shorthand (severity defaults to `warning`) or as an object carrying the severity:

```json
{
  "type": "number",
  "required": { "severity": "error" },
  "minimum": 0,
  "maximum": { "value": 100, "severity": "error" }
}
```

* `Violation` carries a `severity`, stamped where it is raised — Constraint-backed codes take the severity their Constraint configures, system conditions take the fixed severity from the ADR.
* `isBlocking()` now means `severity === error`, replacing the hardcoded non-blocking code list.
* Every serialized violation gains an always-present `severity` field, in the dry-run 200 bodies and the enforcement 422 alike.
* Severity is deliberately **not** part of `ViolationDiff` identity, so changing a Constraint's severity never makes a pre-existing violation count as newly introduced.

### Behaviour change worth noting

`required` (and the other authorable Constraints) now default to `warning`, so with `$wgNeoWikiEnforceValidation` on, an **unannotated** Schema blocks nothing beyond the fixed-`error` system conditions. That is the ADR's intent — invalid Subjects are a normal, supported state, and blocking is what a schema author opts into. The enforcement fixtures in the test suite were annotated `error` to keep testing the blocking path.

## Design note

Severity is stored in a name-keyed map on `PropertyCore`, populated by a generic `SeverityNormalizer` at the JSON boundary and read via `PropertyDefinition::severityOf()`. The concrete `*Property` classes are untouched.

The alternative — a `minimumSeverity` sibling field per Constraint on each Property class — was rejected because Property Types are an extensibility surface: this way an existing custom type (e.g. RedHerb's `ColorType`) keeps working with **zero** changes and opts into severity with a single `severityOf()` call, and the name-keyed map generalises to the ADR's deferred "constraints as data" direction. The accepted cost is that severity is not visible as a typed field on `NumberProperty`; the code↔constraint link stays local at each `validate()` site.

Severities annotated on declared Display Attributes are dropped (they are not Constraints, and re-emitting them would change the attribute's serialized shape), and the content schema rejects a malformed severity at save.

## ADR amendment

ADR 26's fixed-severity table omitted three codes the validator already produces. They are documented at the severity that preserves their existing behaviour: `unregistered-type` → `warning`, `relation-target-not-found` → `warning`, `relation-target-schema-mismatch` → `error`.

## Docs

`docs/` publishes to neowiki.ai, so the user-facing contracts this PR changes are updated alongside it: the always-present `severity` field and a per-code severity in [Validation codes](docs/api/validation-codes.md) (whose "three codes are non-blocking, every other code blocks" rule this PR replaces), the scalar-or-object Constraint form in [Schema format](docs/api/schema-format.md), the Constraint entry in the [Glossary](docs/glossary.md), and the corrected `$wgNeoWikiEnforceValidation` description in both the installation docs and `extension.json`.

## Scope

Backend only. The follow-ups the ADR anticipates in its Cons are tracked separately:

* #1148 — the frontend must parse the dual Constraint form and style `warning` vs `error`. Until it lands, a Schema saved with object-form Constraints is mis-read by the schema editor.
* #1149 — the per-Constraint severity control in the Schema editor. Until then, `error` severity is authored by hand-editing Schema JSON.

## Verification

`make phpunit` (targeted suites) and `make cs` green locally: SeverityNormalizer 9, Property 369, Presentation 62, Violation 131, Type 205, Validation 88, SchemaContentValidator 21, CreateSubjectApi 17, ReplaceSubjectApi 14; phpcs 593/593 and phpstan clean.

Smoke-tested live against a dev wiki over the REST API — object-form Schemas save, a typo'd severity is rejected at save, Constraints round-trip (`error` keeps the object form, unannotated stays shorthand, a Display Attribute stays a bare scalar), and `POST /subject/validate` returns:

```json
[{"propertyName":"Score","code":"max-value","args":[100],"severity":"error"},
 {"propertyName":"Code","code":"required","args":[],"severity":"error"},
 {"propertyName":"Note","code":"required","args":[],"severity":"warning"}]
```

## Manual Browser Check

N/A — backend only. There is no UI surface in this PR; the warning-vs-error styling and the severity authoring control arrive in #1148 and #1149 respectively. The behaviour above was verified over the REST API instead.

<sub>AI-authored — Claude Code, `Opus 4.8 (1M context)`; implemented from the pre-existing ADR 26, with the domain-representation fork (generic normalization vs per-Constraint fields) explored and decided by @alistair3149; reviewed by a three-lens agent panel (design soundness / elegance / correctness) and its findings applied; full targeted PHPUnit suites and phpcs/phpstan green locally, live-verified via the REST API; CI not yet run.</sub>

### DIFF
--- a/docs/adr/026-validation-severity-levels.md
+++ b/docs/adr/026-validation-severity-levels.md
@@ -87,6 +87,9 @@ carve-out:
 | `invalid-url` / `invalid-date` / `invalid-datetime` | the value cannot be interpreted as its type | `error` | type integrity, comparable to a parse error in a linter, which is fatal regardless of rule configuration |
 | `single-value-only` | multiple values on a single-value property | `error` | `multiple` declares the value's shape, like `type` itself, so it is a shape condition rather than a Constraint |
 | `schema-not-found` | the Subject's Schema page is missing | `warning` | severity configuration lives in the Schema and there is none to read from; a warning keeps the Subject editable |
+| `unregistered-type` | the Statement's Property Type is not registered (extension disabled) | `warning` | degraded wiki state, not a bad edit; keeps the Subject editable, like `schema-not-found` |
+| `relation-target-not-found` | a relation value points to a Subject that does not exist | `warning` | red-link philosophy: the target may be minted later (e.g. during import) |
+| `relation-target-schema-mismatch` | a resolvable relation target's own Schema is not the declared `targetSchema` | `error` | a resolvable target with the wrong shape is a genuine integrity error |
 
 The fixed `error` for uninterpretable values (`invalid-url`, `invalid-date`, `invalid-datetime`) is the most
 debatable of these classifications: messy cultural-heritage imports contain exactly such values. It stays workable
--- a/docs/api/schema-format.md
+++ b/docs/api/schema-format.md
@@ -38,8 +38,37 @@ Every property definition carries the common fields below plus the type-specific
 |-------|------|----------|---------|-------------|
 | `type` | string | Yes | - | The property type. See [Property Types](#property-types). |
 | `description` | string | No | `""` | Human-readable description of the property |
-| `required` | boolean | No | `false` | Whether a value is required for this property |
+| `required` | boolean or object | No | `false` | Whether a value is required. Accepts a [severity](#constraint-severity). |
 | `default` | varies | No | `null` | Default value when none is provided |
+
+## Constraint severity
+
+Every Constraint carries a severity of `error` or `warning`, which decides whether violating it can
+block a write. Write a Constraint either as the bare value, which keeps the default `warning`, or as
+an object carrying the severity:
+
+```json
+{
+  "type": "number",
+  "required": { "severity": "error" },
+  "minimum": 0,
+  "maximum": { "value": 100, "severity": "error" }
+}
+```
+
+Boolean Constraints (`required`, `uniqueItems`) take no `value` in the object form — writing the
+object at all implies `true`. Every other Constraint carries its value under `value`, including
+`options`, whose `value` is the options array.
+
+The Constraints that accept a severity are `required`, `minimum`, `maximum`, `minLength`,
+`maxLength`, `uniqueItems`, and `options`. Severity is a Constraint concept, so it does not apply to
+Display Attributes such as `precision`, where it is discarded, nor to the shape-declaring fields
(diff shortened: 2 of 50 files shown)


JeroenDeDauw:
I am still reviewing this. Likley to complete within 24h

docs/adr/026-validation-severity-levels.md
```diff
@@ -87,6 +87,9 @@ carve-out:
 | `invalid-url` / `invalid-date` / `invalid-datetime` | the value cannot be interpreted as its type | `error` | type integrity, comparable to a parse error in a linter, which is fatal regardless of rule configuration |
 | `single-value-only` | multiple values on a single-value property | `error` | `multiple` declares the value's shape, like `type` itself, so it is a shape condition rather than a Constraint |
 | `schema-not-found` | the Subject's Schema page is missing | `warning` | severity configuration lives in the Schema and there is none to read from; a warning keeps the Subject editable |
+| `unregistered-type` | the Statement's Property Type is not registered (extension disabled) | `warning` | degraded wiki state, not a bad edit; keeps the Subject editable, like `schema-not-found` |
+| `relation-target-not-found` | a relation value points to a Subject that does not exist | `warning` | red-link philosophy: the target may be minted later (e.g. during import) |
+| `relation-target-schema-mismatch` | a resolvable relation target's own Schema is not the declared `targetSchema` | `error` | a resolvable target with the wrong shape is a genuine integrity error |
```
JeroenDeDauw, line 92: Weird to add to an already-accepted ADR and also weird it seems to contain a reference. Isn't that for docs elsewhere? ADR isn't to document the implementation.

JeroenDeDauw: So potentially a bunch of stuff already in this ADR that should not be here

alistair3149: I agree, it was the natural place to track it first because the ADR was used as a proposal.
But now we should probably put the references elsewhere.

PR #1152, opened 2026-07-23
Project url property values as IRI objects

### LINKED ISSUE #1151: Project url property values as IRI objects instead of string literals
Current behavior (verified on the demo wiki, 2026-07-24): url-type property values project as typed string literals, e.g. Vermeer's EDM export contains `foaf:page "https://en.wikipedia.org/wiki/Johannes_Vermeer"^^xsd:anyURI`.

Two problems:

1. String literals are inert in RDF: SPARQL cannot join on or follow them. Emitting IRI objects is what lets a statement point at an external semantic-web entity (Wikidata, VIAF, Getty) and be part of the global graph. Requested for ECHOLOT in https://github.com/ProfessionalWiki/NeoWiki/discussions/996 (Q5: "using a triple value IRI from another semantic web resource without duplicating it", asked as an "at least" baseline).
2. It is nonconformant where target vocabularies expect resources: `foaf:page` has range `foaf:Document`; `edm:isShownBy` expects an `edm:WebResource`.

Fix: url property values project as IRI objects (`<https://…>`) in the projections, native and ontology alike. Values that are not valid IRIs keep the current string-literal form, so nothing is lost.

Out of scope, deliberately: external entities as first-class relation targets / federation (ADR 23 territory, being explored via questions in #996), retaining external URIs on import (#306), and any new external-identifier property type.

> <sub>AI-authored — Claude Code, `Fable 5 (max)`; filed at @JeroenDeDauw's direction from the #996 Q5 exchange; not line-reviewed; current behavior verified against the live demo export.</sub>

JeroenDeDauw:
Fixes https://github.com/ProfessionalWiki/NeoWiki/issues/1151

url property values projected to RDF as typed string literals (`"https://…"^^xsd:anyURI`), which are inert: SPARQL cannot follow them, and they are nonconformant where a vocabulary expects a resource (e.g. `foaf:page`, `edm:isShownBy`). They now project as IRI objects (`<https://…>`) in both the native and ontology (Mapping) projections. A value that is not a valid absolute IRI (e.g. the scheme-less values the url type still accepts) keeps the `xsd:anyURI` literal form, so nothing is lost.

Explicit datatype wins: when a Mapping property entry sets `datatype` on a url-typed property, that override is deliberate configuration and takes precedence, so the value is emitted as a literal with that datatype rather than an IRI object.

The value-mapper seam (`RdfValueMapperRegistry`, and the `addRdfValueMapper` extension point) now returns `RdfTerm[]` (IRIs or literals) instead of `Literal[]` — a backward-compatible widening, since a `Literal` is an `RdfTerm`.

The IRI-validity check reuses the one that validates authored Mapping terms. It moves from `CurieExpander::isSafeAbsoluteIri` to `Iri::isSafeAbsolute` (its natural home on the value object); `CurieExpander` delegates, unchanged in behavior. This lets the native value mapper reuse the check without `Domain\Rdf` depending on `Domain\Mapping`, which would introduce a dependency cycle.

Docs updated: `rdf/rdf-export.md`, `rdf/ontology-mapping.md` (datatype-override note), `planning/NativeRdfProjection.md`, `extending/extending.md`.

Considered, omitted: `docs/examples/person-to-edm.md` needed no change (its shown output contains no url literal). The SPARQL/QLever tests round-trip only labels, so none asserted the url literal form.

> <sub>AI-authored — Claude Code, `Opus 4.8 (max)`; detailed spec from @JeroenDeDauw, executed autonomously with no human revisions; diff not yet human-reviewed; full PHPUnit suite, phpcs and phpstan green locally and in CI, new tests mutation-checked, native + EDM RDF export verified live on the worktree demo wiki.</sub>


JeroenDeDauw:
pr-review did not find anything

PR #1154, opened 2026-07-23
Show button focus rings for keyboard users only

JeroenDeDauw:
Clicking a NeoWiki button with the mouse left the Codex focus ring on it until focus moved away — on the Data-tab row actions (copy link, edit, delete), the export buttons, the infobox edit button, and dialog actions. Codex 1.14 styles button focus on bare `:focus`, which a mouse click matches even though the browser does not treat it as `:focus-visible`.

Focus indication is now keyboard-only. For `:focus:not( :focus-visible )` a scoped override drops the inset ring and restores each button variant's resting border-color, mirroring Codex's own per-variant focus selectors. Keyboard focus (`:focus-visible`) is left intact, as are the pressed (`:active`), toggled (`--is-active`) and Windows-High-Contrast states. The two reveal-on-focus lists in the schema and layout editors move from `:focus-within` to `:has( :focus-visible )` for the same reason, extending the row-actions change in #1144.

The override applies under `.ext-neowiki-ui`, added to every Vue app mount root through a shared mount helper and to each dialog — Codex teleports dialogs out of the mount tree, so the class rides on the CdxDialog and reaches the close and footer buttons — plus the existing `.ext-neowiki-view` page placeholder. It never touches Codex buttons from MediaWiki core or other extensions. The rationale lives in the override file header (`assets/keyboard-focus.less`).

Verified in the browser across the button variants and scopes: mouse focus leaves no ring once the pointer moves away, keyboard focus still shows it, and core skin buttons are unaffected.

> <sub>AI-authored — Claude Code, `Opus 4.8 (max)`; detailed spec from @JeroenDeDauw, no mid-task steering; diff not yet human-reviewed; lint + vitest (1136 tests) + phpcs/phpstan green locally, before/after confirmed in-browser with Playwright, CI green.</sub>

### DIFF
--- a/resources/ext.neowiki/src/assets/keyboard-focus.less
+++ b/resources/ext.neowiki/src/assets/keyboard-focus.less
@@ -1,0 +1,59 @@
+/**
+ * Keyboard-only focus rings for NeoWiki's Codex buttons.
+ *
+ * Codex 1.14 styles button focus on bare `:focus`, so a mouse click — which
+ * focuses the button but, per the browser heuristic, does not match
+ * `:focus-visible` — leaves the progressive focus ring painted until focus
+ * moves away. Focus indication should be shown to keyboard users only.
+ *
+ * For `:focus:not( :focus-visible )` we drop the inset ring and restore each
+ * variant's resting `border-color`, mirroring the per-variant focus selectors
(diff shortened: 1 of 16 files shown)


JeroenDeDauw:
Looks good locally. Did not review the code yet beyond a cursory scan

PR #1155, opened 2026-07-23
Default Subject IRI dereference to the Data tab

JeroenDeDauw:
Default `$wgNeoWikiDereferenceSubjectsToDataTab` to `true`, so a browser dereferencing a Subject's concept URI (`/entity/{id}`, served by `ResolveSubjectIriApi`) lands on the hosting page's Data tab opened on the Subject's row, rather than the plain hosting page. RDF `Accept` negotiation (TriG/Turtle) is untouched — only the `text/html` target changes.

## Why

Only data-context users ever hold a concept URI — they come from RDF exports, SPARQL results, and reconciliation payloads — so the dereference audience wants the structured record. A page renders its Main Subject in the infobox, but child Subjects may appear nowhere on the page view, making the Data-tab row the only landing that is correct for *every* Subject. It also matches the expectation Wikibase users bring: a concept URI resolves to the entity's data view. The plain page stays one click away, and wikis that prefer it keep the setting (set it `false` on the `MediaWiki:NeoWiki` page or in `LocalSettings.php`).

## Changes

- `extension.json`: value `false` → `true`; description reworded so the stated default matches.
- `src/Application/WikiConfig/ConfigExample.php`: the `MediaWiki:NeoWiki` preload example is documented as "set to the defaults", so its `dereferenceSubjectsToDataTab` flips to `true` too.
- `src/EntryPoints/REST/ResolveSubjectIriApi.php`: class doc inverts the default/opt-out framing.
- `docs/rdf/rdf-export.md`: default column and the HTML-target prose inverted.
- Tests: the plain-page cases in `ResolveSubjectIriApiTest` now disable the setting explicitly; a new default-behavior test (`testHtmlDereferenceRedirectsToTheSubjectsDataTabRowByDefault`, no override) asserts the Data-tab `Location` (subjects action + bare Subject-id fragment). `InWikiConfigDereferenceTest`'s pin comment corrected (the false pin is now the opposite of the page value, not "the default").

No behavioural change for anyone who has set the value explicitly. The on-wiki `MediaWiki:NeoWiki` override and `LocalSettings.php` continue to take precedence as before.

## Verification

Live, on the worktree dev wiki, against a real demo Subject (`s1demo1aaaaaaa1`, hosted on `ACME_Inc`):

- Default (no override), `Accept: text/html` → `303` → `.../wiki/ACME_Inc/subjects#s1demo1aaaaaaa1` (Data-tab row).
- Setting `false` via `MediaWiki:NeoWiki` (also confirming the on-wiki override still wins), `Accept: text/html` → `303` → `.../wiki/ACME_Inc` (plain page).
- `Accept: text/turtle` → `303` → `.../subject/s1demo1aaaaaaa1/rdf?format=turtle` (RDF unaffected).

Full PHPUnit suite (2020 tests), phpcs, and phpstan all green locally.

> <sub>AI-authored — Claude Code, `Opus 4.8 (max)`; detailed spec from @JeroenDeDauw, no redirection; diff not yet human-reviewed; TDD test restructure (default test seen red before the flip), full PHPUnit suite (2020 tests) + phpcs + phpstan green locally, all three dereference cases verified live on the worktree wiki; mutation-tested; CI green.</sub>

### DIFF
--- a/docs/rdf/rdf-export.md
+++ b/docs/rdf/rdf-export.md
@@ -107,8 +107,8 @@ content-negotiates it and answers `303 See Other` with an absolute `Location`:
 TriG wins when both RDF types are acceptable; the RDF redirects use the native projection. A Subject that is absent or
 on a page the caller may not read returns one indistinguishable `404`; a malformed id `400`.
 
-The HTML target is that page's Data tab (`?action=subjects`) opened on the Subject's row (`#{subjectId}`) by default,
-or the plain hosting page when `$wgNeoWikiDereferenceSubjectsToDataTab` is disabled.
+The default HTML target is that page's Data tab (`/Example_Page/subjects`) opened on the Subject's row.
+When `$wgNeoWikiDereferenceSubjectsToDataTab` is set to `false`, the target is the wiki page itself (`/Example_Page`).
 
 The negotiator is always reachable at the REST path, which needs no server configuration:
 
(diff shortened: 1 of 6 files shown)


JeroenDeDauw:
LGTM

PR #1157, opened 2026-07-24
Set proper filename with pageid for RDF-Export

LibrErli:
Add Content-Disposition header for proper RDF filename download

Currently, RDF exports from `/rest.php/neowiki/v0/page/{id}/rdf` are
downloaded by browsers as a generic file named "rdf".

This PR adds a `Content-Disposition: attachment; filename="{pageId}.rdf"`
header so files download with proper names like `36.rdf`, `42.rdf`, etc.


malberts:
Should we rather return an extension that matches the output format? This endpoint currently returns Turtle (`.ttl`) or TriG (`.trig`), not XML-based `.rdf`.

The format is already known in `RdfFormatNegotiation::rdfResponse()`, where the content type header is set, so the content disposition probably also belongs there instead.

JeroenDeDauw:
Follow up:
* https://github.com/ProfessionalWiki/NeoWiki/pull/1161

JeroenDeDauw (approved):

PR #1164, opened 2026-07-25
Give the demo wiki a main menu tailored to its content

JeroenDeDauw:
The demo wiki ships MediaWiki's stock sidebar — Main page, Recent changes, Random page, MediaWiki help — so a visitor who lands on a subject page rather than the Main Page has no route to the datasets or the feature demos, and a quarter of the menu points off-site.

Add `MediaWiki:Sidebar` to the demo data: the four dataset hubs in the order the Main Page presents them, then the feature demos, then the two exits an evaluator wants. Labels match the Main Page's names for the same pages, so both surfaces teach one vocabulary. Features runs data → model → display → export → query → build, opening on the Rijksmuseum Data tab, which the Main Page was the only route to. Sidebar lines containing a pipe run through `MessageCache::transform`, so the `{{fullurl:}}` that reaches an action view resolves against whichever wiki serves the demo data.

The demo data had no route into the MediaWiki namespace, so `DemoData/MediaWiki/` joins `Module/` as a directory whose files import under a namespace prefix.

Considered, omitted: Recent changes and Random page — anonymous visitors cannot edit a sandbox that resets periodically, and the corpus holds enough thin reference subjects that random landings undersell it. Special:Layouts and Special:Mappings — both are one hop away via the Main Page and the Developers hub, where Special:Schemas alone is not. Validation Demo — its payoff is an editing flow anonymous visitors cannot run.

Note that Vector 2022 keeps the main menu behind the hamburger for anonymous visitors, so on neowiki.dev this menu is one click away until https://github.com/ProfessionalWiki/DockerWiki/pull/970 grows a matching line for `main-menu-pinned`.

> <sub>AI-authored — Claude Code, `Opus 5 (max)`; open-ended ask from @JeroenDeDauw ("optimize the main menu for the DemoData", then a pointer that the Data tab is a set piece), menu content designed by a `Fable 5 (max)` subagent and reviewed by a fresh-context text-review subagent, whose out-of-scope prose edits I reverted; diff not yet human-reviewed; full PHPUnit suite and `make cs` green locally, CI green, and every link checked against a worktree wiki — all 13 resolve, none red.</sub>

### DIFF
--- a/DemoData/AGENTS.md
+++ b/DemoData/AGENTS.md
@@ -14,12 +14,16 @@ partners, knowledge managers, MediaWiki ecosystem evaluators, and live-demo audi
 | `Page/<Name>.wikitext` | Free-form wiki pages (hubs, references) | Main namespace, `<Name>` |
 | `SparqlPage/<Name>.wikitext` | Pages demoing the SPARQL surfaces. Imported only when `$wgNeoWikiSparqlStores` is non-empty (elsewhere `{{#sparql_raw}}` is unregistered and would render literally). | Main namespace, `<Name>` |
 | `Module/<Name>.lua` | Scribunto modules | `Module:<Name>` |
+| `MediaWiki/<Name>.wikitext` | Interface pages, such as `Sidebar` (the main menu) | `MediaWiki:<Name>` |
 
 `ImportDemoData.php` reseeds the demo set: it creates and updates pages from these directories, and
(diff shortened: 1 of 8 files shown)


JeroenDeDauw:
image

JeroenDeDauw:
LGTM

Issue #1163, opened 2026-07-25
RDFS/OWL self-description of Schemas in the native projection


JeroenDeDauw:
Emit an RDFS/OWL definition for each Schema (as a class) and each Property Definition (as a property with domain and range), so a wiki's own Schemas are readable as RDF alongside the data. [NativeRdfProjection.md Q10](https://github.com/ProfessionalWiki/NeoWiki/blob/master/docs/planning/NativeRdfProjection.md#open-questions); split out of #1027, where it was a stretch item.

Why it is wanted: NeoWiki has no global properties, so typing lives per Schema and the native projection's flat predicate IRIs claim no equivalence across Schemas. An RDFS export publishes that per-Schema typing in standard terms, which is what mapping authors need as input — a wiki's Schemas are effectively its own ontology. Requested by takin (T2.3) on 2026-07-03 as authoring input for ontology mappings.

**Gated on partner confirmation.** The idea was put back to @Habennin and @lozanaross in https://github.com/ProfessionalWiki/NeoWiki/discussions/999 (2026-07-22) as a question — whether an RDFS/OWL self-description is what they need when authoring mappings. Wait for that answer before designing, so the shape follows the actual use.

Open when picked up: which vocabulary (RDFS only, or OWL), what a `relation` property's range is, where it is served (part of the page/dump export, a dedicated endpoint, or a Schema-page projection), and how it relates to the generated shape exports in [ShapeLanguages.md](https://github.com/ProfessionalWiki/NeoWiki/blob/master/docs/planning/ShapeLanguages.md).

> <sub>AI-authored — Claude Code, \`Opus 5 (max)\`; one-line ask from @JeroenDeDauw to assess #1027, this is the stretch item split off so closing #1027 does not drop it; text not human-reviewed before posting; no code involved, demand and status traced to the cited doc and discussion.</sub>

JeroenDeDauw:
The gate on this issue is satisfied: partner confirmation arrived at the ECHOLOT T2.3 meeting on 2026-08-10. @Habennin asked for exactly this — an RDFS export of an instance's Schemas as a T-box view — rating it high-value for semantic-web audiences and likely needed as input for mapping tooling. Accepted caveat from the same discussion (OEAW): the export is necessarily lossy — validation semantics do not survive RDFS. Related earlier demand: the domain/range thread in #999.

[...]

Issue #1053, opened 2026-07-16
Include the projection in named-graph IRIs so multiple projections can share one triple store


JeroenDeDauw:
For [discussion #996](https://github.com/ProfessionalWiki/NeoWiki/discussions/996) (Lozana's Q1/Q2): requiring one triple store per target ontology is an operational burden, and multiple mappings should be queryable in the same store.

## Problem

Both projectors place every quad in the same named graph — the page IRI `{$base}/page/{pageId}`: `RdfNamespaces::page()` is used as the graph by both `RdfPageProjector` (native) and `OntologyMappingProjector`. Two projections written into the same store therefore collide: under the per-page replace sync planned for #586 (`DROP SILENT GRAPH <g>; INSERT DATA { GRAPH <g> { … } }`), whichever projection writes a page last wipes the other's triples for that page, and concatenated dumps merge into mixed graphs.

The current design papers over this with "a store holds exactly one projection; wanting the same data in two vocabularies means configuring two stores" (OntologyMapping.md) — which is the burden #996 Q1 calls out. For QLever that literally means one server process per vocabulary: a QLever server serves exactly one index (no Fuseki-style multi-dataset hosting).

## Change

Qualify named-graph IRIs by projection, uniformly (native included):

```
{$base}/graph/{projection}/page/{pageId}
e.g. {$base}/graph/native/page/42 and {$base}/graph/edm/page/42
```

- The page resource IRI (`neo-page:42`, used in triples) stays unchanged; only the graph name moves.
- Subject/Relation/property/Schema IRIs stay projection-independent (deliberate: shared Subject IRIs are what make co-querying work).
- Projection names are IRI-encoded with the existing name-encoding regime (targets are author-controlled strings).
- The page id stays in the graph IRI, so per-page provenance and the per-page replace sync are unchanged.
- Projection-first ordering gives each projection a prefix-abbreviable graph namespace and allows prefix-based scoping (`STRSTARTS`) for the rare "restrict to one projection" query.

With this, "a store holds one projection" softens to "a store holds one **or more** projections, each in its own graph family" — and the #586 store config should be endpoint + list of projections. Multiple stores remain a deployment option (hard isolation, independent scaling), not a requirement.

[...]

PR #1165, opened 2026-07-25
Demo the native and EDM projections as siblings in one SPARQL store

### LINKED ISSUE #1027: End-to-end EDM projection exercise: person toy model, mapping, demo
Part of the #1023 milestone; the milestone's proof point. Exercise the projection + mapping stack end to end, as proposed in [OntologyMapping.md](https://github.com/ProfessionalWiki/NeoWiki/blob/master/docs/planning/OntologyMapping.md): implement the neutral person model natively, map it, and project it to EDM — proving or disproving the local-schemas-plus-mappings approach by doing.

Scope:

- Person and City Schemas implementing the shared toy model (the "Neutral Person to Many Standards" spreadsheet linked from OntologyMapping.md).
- An EDM Mapping page for them.
- Demo setup with two sibling stores over the same wiki (native projection + EDM projection), with example SPARQL queries for both.
- Update the planning docs where the implementation settled previously open points.
- Draft feedback-round posts for #996 and #999 (human-reviewed before posting).
- Stretch: RDFS export of Schemas (NativeRdfProjection.md Q10), requested as authoring input for mapping work.

Depends on #586 and #1026.

> Scope approved by @JeroenDeDauw in-session as part of the #1023 plan; issue text written autonomously and posted without human review.
> Context: the RDF planning docs, their feedback threads, and the shared toy-model spreadsheet reference.

JeroenDeDauw:
Fixes https://github.com/ProfessionalWiki/NeoWiki/issues/1027

The person toy model, the EDM Mapping, the per-page and bulk EDM export, the SPARQL store sync, and the feedback round
on #996 and #999 all landed earlier. What the milestone still promised was the demo: native and EDM as sibling
projections over one wiki, with example queries for both. The dev stack held only the native projection, so nothing on
a running wiki showed that a store can hold several projections, let alone that a query can join across them.

The dev stack now configures two store entries at the one QLever endpoint, `native` and `EDM`. Each projection
writes its own family of per-page named graphs (#1053), so they coexist; a rebuild fills both. The query surfaces
target the first entry's endpoint, which both entries share, so both projections are reachable.

Demo data gains an `EDM queries` page: two queries in EDM terms, and one that, for the same entity IRI, joins the EDM
projection's birth date with `neo-prop:Source`, which `Mapping:EDM` leaves unmapped. Its result carries both graph
IRIs, so the reader sees one page in two projections in one store.

A wiki whose store holds only the native projection must not render an example that silently returns no rows. Such a
wiki still imports `Mapping:EDM`, so the gate cannot be the Mapping's existence: the new `EdmSparqlPage/` directory is
imported only when the store the read surfaces query holds the EDM projection *and* the native one, since the page
claims both and its last query joins them. The pointers to it from `SPARQL queries` and `RDF and ontology mappings`
are `{{#ifexist:}}`-gated, and `SparqlPage`, whose examples are all native, now requires `native` rather than merely
some usable store. Verified live both ways: with the EDM entry the page imports and all three queries return rows;
without it the page is deleted on re-import and both pointer sentences disappear, leaving no redlink.

Two documentation errors are fixed along the way: the `projection` example value was lowercase `'edm'`, which
normalises to `Mapping:Edm` and does not resolve, and `planning/OntologyMapping.md` described a store config that
carries a list of projections.

Adding a store entry does not project pages already saved, so `installation.md` and `Docker/README.md` now say to
rebuild, and `maintenance.md` records that a rebuild feeds every configured backend rather than only Neo4j. On the
dev stack the EDM entry depends on the demo data's `Mapping:EDM` page: without it every Subject save logs an
unknown-projection error and a rebuild reports every page as failed, which `Docker/README.md` now warns about.

Making the live neowiki.dev demo match needs the same second entry in its own deployment configuration plus a
redeploy, which is outside this repository.

The stretch item on #1027, an RDFS/OWL export of Schemas, is split off to #1163 and is not part of this.

> <sub>AI-authored — Claude Code, `Opus 5 (max)`; one-line ask from @JeroenDeDauw to assess #1027 and implement what was left, no redirections; diff not yet human-reviewed; new unit tests seen failing first and mutation-checked, live SPARQL system tests added and run against the dev-stack QLever (they also run in CI), full PHPUnit suite and `make cs` clean locally, and the demo pages verified rendering real rows on a dev wiki both with and without the EDM store entry; a fresh-context prose review and a fresh-context code review were run and applied in the second commit.</sub>


JeroenDeDauw:
LGTM

Issue #350, opened 2023-10-10
Investigate slot security


JeroenDeDauw:
AC: Answer questions:
* Can we restrict READ access to slots? For instance, disallow access to the `neo` slot. 
* Does the web API allow for WRITE access to non-main slots? And if it does, can we restrict this?

https://www.mediawiki.org/wiki/API:Revisions

http://localhost:8484/index.php/Special:ApiSandbox#action=query&format=json&prop=revisions&titles=Professional_Wiki&rvprop=timestamp%7Cuser%7Ccomment%7Ccontent&rvslots=main%7Cneo

![image](https://github.com/ProfessionalWiki/NeoWiki/assets/146040/8737a446-0ec4-485b-b546-15dd8f20dde3)

JeroenDeDauw:
Both questions answered below, against MediaWiki 1.43.6 source with the behaviour reproduced on a 1.43.8 dev wiki.

**Summary: there is no unguarded write path into the `neo` slot, and slot read cannot be restricted separately from page read because MediaWiki has no per-slot read primitive. Page read is necessarily the unit of access control.**

## Can we restrict READ access to slots?

**No**, and not by design choice — core offers no mechanism, on four independent axes:

- **`SlotRoleHandler`** exposes `getRole`, `getOutputLayoutHints`, `isDerived`, `getNameMessageKey`, `getDefaultModel`, `isAllowedModel` and `supportsArticleCount`. None takes an `Authority` or `User`. `isAllowedModel()` is a write-time content-model check, called only from `PageUpdater`.
- **`RevisionRecord`'s `$audience`** resolves to `audienceCan( DELETED_TEXT, … )` — the revision-wide RevDel bitfield, with no per-role dimension. `RevisionSlots::getContent()` takes no audience argument at all, and `RevisionRecord::getSlots()` is documented as "no audience checks applied".
- **No config setting.** The only slot-related one is `$wgRevisionSlotsCacheExpiry`, a cache TTL.
- **No hook.** The only slot-named hooks are `GetSlotDiffRenderer` and `TextSlotDiffRendererTablePrefix`, both diff rendering.

The slot is consequently readable on every core surface that exposes revision content, each gated on per-page `read`:

| Surface | Exposes `neo` |
|---|---|
| `action=query&prop=revisions&rvslots=neo` | yes |
| `action=raw&slot=neo` | yes — raw JSON served as `text/plain` |
| `Special:Export` | yes — as an MCR `<content>` element |
[...]

PR #1167, opened 2026-07-25
Correct ADR 004's claim that Subject JSON is only accessible via NeoWiki interfaces

### LINKED ISSUE #1166: Correct ADR 004's claim that Subject JSON is only accessible via NeoWiki interfaces
[ADR 004: Use a Dedicated MediaWiki Revision Slot](https://github.com/ProfessionalWiki/NeoWiki/blob/master/docs/adr/004-use-dedicated-slot.md) lists this consequence:

> The structured data JSON is only accessible via our own interfaces. This improves security and control but also means
> that we need to create our own web APIs and interfaces.

The first two clauses are false. MediaWiki's generic revision surfaces expose the `neo` slot without touching any NeoWiki interface, verified against 1.43 and reproduced on a dev wiki:

- `action=raw&slot=neo` returns the raw Subject JSON as `text/plain`. `SubjectContent` extends `JsonContent` extends `TextContent`, so it passes `RawAction`'s `instanceof TextContent` filter.
- `action=query&prop=revisions&rvslots=neo&rvprop=content` returns the slot content.
- `Special:Export` emits the slot as a full MCR `<content>` element in the XML.
- `action=compare&slots=neo` and the ordinary diff view render it.

Every one of those is gated on the page's `read` permission, so the *behaviour* is safe. What is wrong is the stated rationale: the slot is not an access-control boundary, and "improves security and control" invites reasoning that does not hold. The third clause is still true — core has no write path into the slot and no structured read of it.

Suggested replacement for that bullet:

> * The structured data JSON is reachable through MediaWiki's generic revision surfaces (`action=raw`, `prop=revisions`,
>   `Special:Export`, diffs), each gated on the page's `read` permission. The slot is not an access-control boundary.
>   Editing it and reading it as structured data still require our own web APIs and interfaces.

Answered while investigating #350, which asked whether slot read access can be restricted separately from page read. It cannot: MediaWiki has no per-slot read primitive.

> <sub>AI-authored — Claude Code, `Fable 5 (max)`; spun out of the #350 investigation, a one-line ask from @JeroenDeDauw, no revisions; not yet human-reviewed; the ADR claim was checked against MediaWiki 1.43.6 source and reproduced on a 1.43.8 dev wiki.</sub>

JeroenDeDauw:
Fixes https://github.com/ProfessionalWiki/NeoWiki/issues/1166

ADR 004 listed as a consequence that the structured data JSON "is only accessible via our own
interfaces", and that this "improves security and control". MediaWiki's generic revision surfaces
reach the slot without touching a NeoWiki interface: `action=raw&slot=neo` serves the raw JSON as
`text/plain`, `prop=revisions&rvslots=neo` returns it, `Special:Export` emits it as an MCR
`<content>` element, and diffs render it.

Each of those is gated on the page's `read` permission, so the behaviour is safe. The rationale was
not: the slot is not an access-control boundary, and MediaWiki has no per-slot read primitive at all
— `SlotRoleHandler` never sees an `Authority`, revision audience checks are RevDel-only and
role-agnostic, and there is no config setting or hook for it.

The decision itself is unaffected, so the status line is unchanged.

Considered, omitted: the specific core classes and line numbers, which belong in the investigation
record rather than a decision record; a note that import and rollback can write non-main slots,
which concerns write paths rather than this bullet's claim.

> <sub>AI-authored — Claude Code, `Fable 5 (max)`; follow-up to the #350 investigation, approved by @JeroenDeDauw as a one-line go-ahead, no revisions; not yet human-reviewed; the corrected claims were checked against MediaWiki 1.43.6 source and reproduced on a 1.43.8 dev wiki, docs-only change with no tests referencing ADR files.</sub>

### DIFF
--- a/docs/adr/004-use-dedicated-slot.md
+++ b/docs/adr/004-use-dedicated-slot.md
@@ -16,5 +16,7 @@ wikitext pages as usual and then add the JSON to a dedicated revision slot.
 ## Consequences
 
 * Improved usability compared to Wikibase by having the data on the same page as the wikitext.
-* The structured data JSON is only accessible via our own interfaces. This improves security and control but also means
-  that we need to create our own web APIs and interfaces.
+* The JSON is reachable through MediaWiki's generic revision surfaces — `action=raw`, `prop=revisions`,
+  `Special:Export` and diffs — each gated on the page's `read` permission. The slot is not an access-control boundary:
+  MediaWiki has no way to restrict read access to a single slot separately from the page.
+* Editing the JSON, and reading it as structured data, require our own web APIs and interfaces.


JeroenDeDauw:
LGTM

Issue #952, opened 2026-06-27
Render historically accurate on history pages


JeroenDeDauw:
> Identified by @JeroenDeDauw, who recalled the point-in-time mechanism and worked out the three
> affected cases (including the Layout→Schema one) through extended back-and-forth.
> Context: a trace of the NeoWiki rendering/versioning paths — the PHP persistence layer and the Vue display components.
> <sub>Written by Claude Code, `Opus 4.8`</sub>

## Summary

When viewing an old revision of a page, NeoWiki resolves the **Subject data** point-in-time correctly, but renders it
with the **current Schema and Layout** rather than the versions in force when the revision was saved. Only data
interpretation is versioned (the per-Statement writer's-schema type, ADR 011); display configuration is not.

This is a presentation-fidelity gap, **not** data loss — the historical data is faithful. Filing it to decide whether
historical views should also pin display configuration.

## Background: what already works

- **Subject data is point-in-time.** `PointInTimeSubjectLookup` is built with the viewed revision and resolves the
  Subject from that revision's `neo` slot, or — when the Subject lives on another page — via `findRevisionAtOrBefore()`,
  which picks the owning page's newest revision with `rev_timestamp <=` the viewed revision's timestamp
  (`src/Persistence/MediaWiki/Subject/PointInTimeSubjectLookup.php:28-31,71-91`). Auto main-subject rendering threads
  the oldid through (`src/EntryPoints/NeoWikiHooks.php:58-63`).
- **Writer's-schema type is pinned** per Statement, so values render with their original type even if the property type
  later changed (ADR 011).
- **Schema and Layout _own_ history pages already render correctly.** Their frontend init reads `wgRevisionId` and
  fetches `/v1/revision/{id}`, rendering that revision's content
  (`resources/ext.neowiki/src/neowiki.ts:81-98` schema, `:130-143` layout).

## What's missing

Three cases where a referenced definition is loaded at *current* version on a historical view:

1. **Schema version on Subject history pages** — *substantive*. Drives property ordering and Display Attributes
   (precision, color, required, description). The no-Layout fallback uses schema-defined order.
2. **Layout version on Subject history pages** — *substantive*. Drives which properties are shown, their order, and
   per-property Display Attribute overrides (Display Rules).
3. **Schema version on Layout history pages** — *cosmetic only*. The Layout renders faithfully from its old revision;
   the current Schema is used solely for a type icon + tooltip next to each display rule
[...]

PR #1169, opened 2026-07-25
Name a Statement's property type `propertyType` on every surface

### LINKED ISSUE #1168: Subject reads and writes use different Statement shapes, so Subjects cannot round-trip
The Subject read and write REST endpoints disagree on the Statement field name, and the write side discards what it does not recognise without reporting it.

The read endpoint and the stored slot JSON both use `type`. From `GET /neowiki/v0/page/24/subjects` on a demo wiki:

```json
"Population": { "type": "number", "value": 921402 }
```

[`docs/api/subject-format.md`](https://github.com/ProfessionalWiki/NeoWiki/blob/master/docs/api/subject-format.md) documents `type` as a required field of the Statement object, and `CreateSubjectApi`'s `statements` parameter description points readers at exactly that page: *"Nested shape matches the subject JSON format documented at https://neowiki.ai/docs/api/subject-format."*

`StatementListBuilder::build()` accepts only `propertyType` (`src/Application/StatementListBuilder.php:38-41`):

```php
if ( !is_array( $entry ) || !isset( $entry['propertyType'] ) ) {
    continue;
}
```

## Consequence

A Subject cannot be round-tripped. `POST /page/{pageId}/mainSubject` with a body copied from the read endpoint returns `201 {"status":"created"}` and stores `"statements": {}`. `PUT /subject/{subjectId}` likewise answers `200 "updated"` having changed nothing. Both reproduced on a dev wiki.

The only symptom is indirect: required properties then report `required` violations, which reads as a Schema problem rather than as discarded input.

This runs through `CreateSubjectAction`, `ReplaceSubjectAction`, `ValidateSubjectQuery` and `ValidateSubjectUpdateQuery` — every write and validate path. The REST tests pass because they use `propertyType` throughout, so the shape they pin is the undocumented one.

## Two separable defects

1. **Field name.** The write path should accept `type`, matching the documentation, the stored format and the read endpoint. There is no production data, so no alias is needed — changing the builder and the tests is enough.
2. **Silent drop.** An unrecognised or malformed entry should be an error, not a `continue`. Returning a success status while discarding input is what makes this expensive to diagnose, and it matters most for batch import, where a partial write is indistinguishable from a complete one.

Dropping *empty* values is a separate, deliberate case (see the comment on `SubjectValidator::validateRequiredProperties`) and should stay.

Related: #546 covers the same theme in `StatementListPatcher`.

Found while probing REST behaviour for #350.

> <sub>AI-authored — Claude Code, `Fable 5 (max)`; incidental finding from the #350 investigation for @JeroenDeDauw, no revisions; not yet human-reviewed; reproduced end to end against a 1.43.8 dev wiki and confirmed against the source on `master`.</sub>

JeroenDeDauw:
Fixes https://github.com/ProfessionalWiki/NeoWiki/issues/1168

The Subject REST endpoints disagreed with themselves: writes read a statement's property type from `propertyType`,
while the stored slot JSON and every read emitted `type`. Both shapes were documented, but a body taken from
`GET /page/{id}/subjects` could not be sent back — the statements were dropped, the response was still
`201 {"status":"created"}`, and the only symptom was a `required` violation that reads as a Schema problem.

Reads and writes now use one shape, so Subjects round-trip. That is the pattern import and migration clients are
built on, and it is what the asymmetry cost.

`propertyType` is the name on every surface. A bare `type` takes its meaning from the object holding it: inside a
Schema's Property Definition it is right, because the object *is* the property. Inside a Statement it is not — a
Statement has no type of its own, and the sibling value does, so `type` invites exactly the wrong reading. The
domain model has always used `propertyName` and `propertyType` for this reason, and the glossary already documented
the field as `propertyType`; the serializer was dropping the qualifier on the way out.

- Stored slot JSON, REST reads, REST write input, and `statementsToJson()` in the frontend all use `propertyType`.
- The Lua accessor's statement table uses `propertyType` too, so the term is not split across surfaces. Its `values`
  key is unchanged, and the shipped `Module:NeoWikiDemo` is updated with it.
- `StatementDeserializer` accepts `propertyType` and falls back to `type`. Revisions written before the rename still
  hold the old key and revision content cannot be migrated, so without this, history pages and diffs of older
  revisions would lose their statements (#952). Writes only ever emit `propertyType`.
- A Property Definition's `type` is unchanged, in JSON and in Lua alike — there the container is the property.
- DemoData Subject fixtures are regenerated onto the new key.

Compatibility, in rough order of who notices:

- **Lua modules reading `.type` off a statement break on parse.** This is the only part that breaks *wiki content*
  rather than a client, and there is no shim: Lua output is computed fresh each parse, so nothing is unmigratable.
  The stale seeded demo module reproduced it precisely — `attempt to concatenate field 'type' (a nil value)` — until
  it was updated.
- **Consumers reading statements from the REST API** see the key change, including anything built on
  `GET /subject/{id}` or `GET /page/{id}/subjects`.
- **Clients writing** with `propertyType`, the long-documented write shape, are unaffected.

There is no production data, so nothing needs migrating beyond the above.

## Tests

`ReplaceSubjectApiTest::testStatementsReadFromTheApiAreAcceptedVerbatim` feeds `GetSubjectApi`'s response straight
into `ReplaceSubjectApi` and asserts the write landed, so it fails if either end's key drifts again.
`StatementDeserializerTest` covers a legacy `type` payload, and `StatementListBuilderTest` pins that write input
rejects it. `NeoWikiLibraryTests.lua` exercises the Lua shape through Scribunto.
`StatementListBuilderTest::testEmptyValueIsDropped` closes a gap found on the way: removing the documented
empty-value drop had left every test passing.

Verified live on a dev wiki beyond CI: the REST round-trip of a verbatim `GET` body through `PUT` and `POST`
including relation and select values; create, edit and clear in the Subject editor; a genuinely pre-rename revision
imported with a `type` slot still rendering its statements; and the Lua table reporting `propertyType` with values
intact.

## Deferred

A malformed entry is still dropped rather than reported; saying so needs a validation code rather than a throw. Split
out as https://github.com/ProfessionalWiki/NeoWiki/issues/1170, which now also records the sharpest case — a client
still sending `type` on write gets `200` and silently loses every statement — plus a pre-existing sibling symptom
where a value of the wrong type for its declared type answers `500`.

Considered, omitted: renaming the "Writing Subjects" heading to "Replacing Subjects", now that it is the superset
holding the only body example.

> <sub>AI-authored — Claude Code, `Opus 5 (max)`; implemented from specs written by an orchestrating Claude session on @JeroenDeDauw's instruction, whose first naming choice he overruled mid-PR and who then extended the rename to Lua; diff not yet human-reviewed; PHPUnit, vitest, phpcs/phpstan and ts-lint/ts-build green in CI, plus the live checks listed above.</sub>

The commit history renames in two directions on purpose. The first commits resolved the read/write split onto `type`,
on the argument that a Schema's Property Definition already uses that key for the same concept. Jeroen overruled it:
a bare field name is read in the scope of its container, and a Statement's container makes `type` mean the wrong
thing. The later commits rename to `propertyType` across storage, reads, writes and the frontend, add the legacy-key
read fallback that keeping storage on `type` had not required, and finally extend the rename to Lua, which the
earlier passes had missed.

</details>

### DIFF
--- a/DemoData/Module/NeoWikiDemo.lua
+++ b/DemoData/Module/NeoWikiDemo.lua
@@ -61,7 +61,7 @@ function p.subject( frame )
 			end
 		end
 		rows[#rows + 1] = '|-'
-		rows[#rows + 1] = '| ' .. name .. ' || ' .. stmt.type .. ' || ' .. table.concat( vals, ', ' )
+		rows[#rows + 1] = '| ' .. name .. ' || ' .. stmt.propertyType .. ' || ' .. table.concat( vals, ', ' )
 	end
 
 	rows[#rows + 1] = '|}'
--- a/DemoData/Subject/ACME_Amsterdam_HQ.json
+++ b/DemoData/Subject/ACME_Amsterdam_HQ.json
@@ -6,11 +6,11 @@
 			"schema": "Office",
 			"statements": {
 				"Description": {
-					"type": "text",
+					"propertyType": "text",
 					"value": [ "Global headquarters. Houses executive leadership, engineering, and the original Anvil product team." ]
 				},
 				"City": {
-					"type": "relation",
+					"propertyType": "relation",
 					"value": [
 						{
 							"id": "r1demo7aaaaaar7",
@@ -19,11 +19,11 @@
 					]
 				},
 				"Address": {
-					"type": "text",
+					"propertyType": "text",
 					"value": [ "Herengracht 1, 1015 BB Amsterdam" ]
 				},
 				"Opened": {
-					"type": "number",
+					"propertyType": "number",
 					"value": 2005
 				}
 			}
--- a/DemoData/Subject/ACME_Engineering.json
+++ b/DemoData/Subject/ACME_Engineering.json
@@ -6,11 +6,11 @@
 			"schema": "Department",
 			"statements": {
 				"Description": {
-					"type": "text",
+					"propertyType": "text",
 					"value": [ "Builds and maintains ACME's product portfolio, from the original Anvil through TNT." ]
 				},
 				"Headcount": {
-					"type": "number",
+					"propertyType": "number",
 					"value": 184
 				}
 			}
--- a/DemoData/Subject/ACME_Inc.json
+++ b/DemoData/Subject/ACME_Inc.json
@@ -6,29 +6,29 @@
 			"schema": "Company",
 			"statements": {
 				"Founded at": {
-					"type": "number",
+					"propertyType": "number",
 					"value": 2005
 				},
 				"Websites": {
-					"type": "url",
+					"propertyType": "url",
 					"value": [
 						"https://acme.example"
 					]
 				},
 				"Status": {
-					"type": "select",
+					"propertyType": "select",
 					"value": ["o1demo1aaaaaaa1"]
 				},
 				"World domination progress": {
-					"type": "number",
+					"propertyType": "number",
 					"value": 42
 				},
 				"Is public": {
-					"type": "boolean",
+					"propertyType": "boolean",
 					"value": true
 				},
 				"Products": {
-					"type": "relation",
+					"propertyType": "relation",
 					"value": [
 						{
 							"id": "r1demo1aaaaaar1",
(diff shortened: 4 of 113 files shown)


JeroenDeDauw:
LGTM

PR #1174, opened 2026-07-25
Require an explicit pick in the schema selector

### LINKED ISSUE #980: New Subject: Existing schema selector should not auto-select when typing
From https://github.com/ProfessionalWiki/NeoWiki/pull/953#issuecomment-4860611708

When creating a new subject with an existing schema, the schema selector picks the schema immediately when typing a matching name. The matching schemas should be shown in the dropdown instead, and the user should still trigger the selection manually (via click or enter).

[Screencast_20260702_003228.webm](https://github.com/user-attachments/assets/edb7e80b-04ec-4732-8daa-63e7fddb8c23)

JeroenDeDauw:
Fixes https://github.com/ProfessionalWiki/NeoWiki/issues/980

Two fixes to the schema selector, used both when creating a subject from an existing
schema and when choosing a relation's target schema.

Typing a schema name out in full no longer selects it. The keystroke that completed the
name used to commit the match, which in the subject creator jumped to the next step while
the user was still typing. A schema is now selected only by clicking its entry or
confirming the highlighted one with the keyboard.

The schema list now opens as soon as the field is focused. It used to stay closed until
the user left the field and came back, because the field was focused before the schemas
had finished loading.

The field is a Codex Lookup rather than a Combobox; the Lookup reports typing and picking
separately. Its expander chevron is gone as a result, and the field appears only once the
schemas have finished loading — a Lookup opens its menu on focus only if it was created
with entries.

Considered, omitted: a placeholder control for the moment before the schemas arrive. The
wait is one microtask once the list is cached, but on a wiki with hundreds of schemas the
first open pages through them serially, so this is worth a look if that gap becomes
visible.

> <sub>AI-authored — Claude Code, `Opus 5 (max)`; detailed spec from an orchestrating Claude session on @JeroenDeDauw's instruction, which named both bugs and their causes and asked for the Lookup and Combobox routes to be weighed against the installed Codex, no redirections; diff not yet human-reviewed; regression tests written first against the real Codex component and seen failing on master, 12 mutations run against the picker and all caught, full vitest suite (1158) plus ts-lint and ts-build green locally, and every acceptance case driven in a real browser on a dev wiki (subject creator and relation target, 17 checks, no console errors). Fresh-context code, security, test, design and edge-case reviews were run and applied in the third commit; the design verdict was good-enough, with CdxLookup judged forced rather than preferred.</sub>

### DIFF
--- a/resources/ext.neowiki/src/components/common/SchemaPicker.vue
+++ b/resources/ext.neowiki/src/components/common/SchemaPicker.vue
@@ -1,8 +1,10 @@
 <template>
 	<div class="ext-neowiki-schema-picker">
-		<CdxCombobox
-			ref="comboboxRef"
-			v-model:selected="selectedSchema"
+		<CdxLookup
+			v-if="schemasLoaded"
+			ref="lookupRef"
+			v-model:selected="pickedSchema"
+			v-model:input-value="inputText"
 			:menu-items="menuItems"
 			:placeholder="$i18n( 'neowiki-schema-picker-placeholder' ).text()"
 			@input="filterSchemas"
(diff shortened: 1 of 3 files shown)


JeroenDeDauw:
Tested locally. Both the subject creation flow and the Relation defintion editor in schema editor

JeroenDeDauw:
@malberts you assigned this

PR #1188, opened 2026-07-26
Cache Schema lookups per request instead of per Subject

JeroenDeDauw:
Backend validation and the graph projection each resolved a Subject's Schema from scratch, so
saving a page with 10 Subjects on one Schema resolved that Schema around a dozen times — once per
Subject in the projection, plus once or twice per written Subject in validation.
`Neo4jSubjectUpdater::updateSubject` carried a `TODO` for its own per-Subject fetch.

`CachingSchemaLookup` already existed, but its WANObjectCache tier stores nothing on a wiki with
`$wgMainCacheType = CACHE_NONE` (the dev stack, and any wiki without a shared object cache), and
even warm it pays serialization per call. `NeoWikiExtension::getSchemaLookup()` also built a new
lookup on every call, so the validator and the projection shared nothing regardless.

## Change

- `CachingSchemaLookup` gains a process-local tier in front of the shared cache, keyed by the same
  article-id + latest-revision-id key. Editing a Schema still yields a new key, so the
  no-stale-Schemas guarantee holds unchanged and needs no invalidation hook — including during a
  `RebuildGraphDatabases.php` run, where the lookup lives for the whole process. The null from a
  Schema whose content does not deserialize is memoized too, so the projection does not re-parse
  it per Subject; a missing or unreadable page returns before the tier.
- `NeoWikiExtension::getSchemaLookup()` pins one lookup on first use rather than building a new one
  per call, so the validator and the graph projection share that tier for the rest of the request or
  run. `getNeo4jPlugin()` already pins its own lookup the same way.
- `CachingMappingLookup`'s docblock now says it mirrors only `CachingSchemaLookup`'s shared tier.

Memory is bounded by the number of Schemas a process touches; ADR 29 places 500 at the top of its
acceptable range.

## Permission semantics

Unchanged: the process-local tier sits behind the read gate, and `authorizeReadByPageTitle` still
runs on every call, ahead of both tiers. It measures 0.0011 ms, so skipping it would buy nothing.
The tier holds user-independent Schema content under the key the shared cache already used, so it
shares nothing the WAN tier did not.

The lookup resolves Schema content as the Authority of its first use and keeps it. The only
realistic mid-process Authority switch is weaker to stronger — a login or account creation mutating
the main context — where reusing the anonymous Authority's resolutions can only under-serve, never
over-serve, and post-login content renders in a fresh request after the redirect anyway.

## Measurements

5000 iterations, single PHP process, dev stack (`CACHE_NONE`), 12-statement Subject on a
12-property Schema:

|                                                      | before                    | after                |
|------------------------------------------------------|---------------------------|----------------------|
| `ProposedSubjectValidator::validate`                 | 0.32-0.38 ms (2.6-3.1k/s) | 0.028 ms (36.4k/s)   |
| `SchemaLookup::getSchema`                            | 0.34-0.38 ms              | 0.020 ms             |
| `SubjectValidator::validate`, Schema already resolved | 0.006 ms                  | 0.006 ms             |

Schema resolution was ~98% of the cost of validating a Subject and is ~71% of a 12x smaller total.
Underlying Schema fetches when projecting a page with N Subjects on one Schema: N to 1.

With a warm shared cache (production `CACHE_ACCEL`/memcached) schema resolution already cost
~0.028 ms before this change, so the win there is ~1.2-1.4x, not 12x. The large win is on wikis
without a shared object cache, and on the first request after any Schema edit.

## Found, not fixed here

Validating a Subject with relation statements costs 0.71 ms per relation target on the same stack:
each target is resolved through `SubjectLookup::getSubject`, one graph round trip apiece,
un-batched. A Subject with a handful of relations therefore costs several ms whatever happens to
Schema lookup, and is the more likely source of a multi-millisecond per-Subject validation figure.
Batching those resolutions is a change to `SubjectValidator` and its lookup interface; it wants its
own issue.

## Considered and omitted

Keying the pinned lookup by Authority and rebuilding on a switch. It guards only the
weaker-to-stronger case above, which cannot over-serve, so it was structure with no threat to
justify it. It did, accidentally, isolate tests from one another — on master nothing persisted
between calls — so it is replaced by an explicit per-test `resetInstance()` in
`NeoWikiIntegrationTestCase`, as a `final` `@before` hook because 52 of the 76 `setUp()` overrides
in tests/phpunit never call `parent::setUp()`. No test class outside that base resolves Schemas
through the pinned lookup.

## Overlap

One line of `Neo4jSubjectUpdater` (the `TODO`) and no Cypher, so no conflict surface with #1185 or
#1171.

> <sub>AI-authored — Claude Code, `Opus 5 (max)`; commissioned by @JeroenDeDauw, detailed AI spec, one design redirection in review (per-Authority guard dropped for pin-first); design-reviewed in session by @JeroenDeDauw, who drove that revision and flagged a docblock for trimming; full diff not yet human-reviewed; profiled before and after on a local dev stack, tests written test-first and confirmed failing without the fix, 9 mutations applied to the cached paths with no survivors, AI-reviewed in a second commit, PHPUnit hook order probed to confirm the test-isolation reset fires despite setUp overrides; full local PHPUnit + phpcs + phpstan green.</sub>

### DIFF
--- a/docs/operations/maintenance.md
+++ b/docs/operations/maintenance.md
@@ -23,13 +23,16 @@ longer has. Run it to:
 - Recover after a backend wipe or restore.
 - Fix any drift between a backend's copy and the canonical revision slots.
 
-Two things to plan around:
+Three things to plan around:
 
 - It reconciles pages, not stray data. Anything the projection never knew about — a node written directly to Neo4j,
   say, or a named graph left behind in a SPARQL store — is not something the rebuild can find. For a guaranteed-clean
   result, empty the backend before rebuilding: wipe Neo4j's data volume, or drop the wiki's named graphs from the store.
 - It runs as a single sequential process with no batching or resume, so the time scales with the number of pages. Plan
   downtime on large wikis.
+- Schema edits made while it runs do not reach it. It resolves each Schema once and keeps that copy for the rest of the
+  run, so a Schema edited mid-run is projected as it stood when the rebuild first read it. Edit Schemas before starting,
+  or re-run the rebuild afterwards.
 
 ## Upgrades
 
--- a/src/Application/Schema/Exception/SchemaContentUnavailableException.php
+++ b/src/Application/Schema/Exception/SchemaContentUnavailableException.php
@@ -1,0 +1,28 @@
+<?php
+
+declare( strict_types = 1 );
+
+namespace ProfessionalWiki\NeoWiki\Application\Schema\Exception;
+
+use RuntimeException;
+
+/**
+ * Thrown when the content of a Schema page that exists could not be read at all, as opposed to
+ * being read and found not to be a valid Schema.
+ *
+ * The two produce the same null for callers, but only the second is a property of the revision.
+ * An unreadable blob is a transient condition — an external store or replica hiccup — so a cache
+ * keyed by revision id must not remember it: the key would not change until someone edited the
+ * Schema, pinning it as missing for the life of the entry. Content that does not deserialize will
+ * not deserialize on the next call either, and is safe to remember.
+ *
+ * Callers are expected to have established that the page exists; {@see CachingSchemaLookup} does so
+ * before it reaches the inner lookup.
+ */
+class SchemaContentUnavailableException extends RuntimeException {
+
+	public static function forName( string $schemaName ): self {
+		return new self( 'Schema content could not be read: ' . $schemaName );
+	}
+
+}
--- a/src/GraphDatabasePlugins/Neo4j/Persistence/Neo4jSubjectUpdater.php
+++ b/src/GraphDatabasePlugins/Neo4j/Persistence/Neo4jSubjectUpdater.php
@@ -26,7 +26,6 @@ class Neo4jSubjectUpdater {
 	}
 
 	public function updateSubject( Subject $subject, bool $isMainSubject ): void {
-		// TODO: we should make sure this schema retrieval is cached
 		$schema = $this->schemaRepository->getSchema( $subject->getSchemaName() );
 
 		if ( $schema === null ) {
--- a/src/NeoWikiExtension.php
+++ b/src/NeoWikiExtension.php
@@ -1028,8 +1028,14 @@ class NeoWikiExtension {
 		);
 	}
 
-	// Pinned to the authority of the first use, like getNeo4jPlugin(): the realistic mid-process
-	// authority switch is a login, where serving the earlier anonymous resolutions only under-serves.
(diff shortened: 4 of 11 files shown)


docs/operations/maintenance.md
```diff
@@ -23,13 +23,16 @@ longer has. Run it to:
...
+Three things to plan around:
 
 - It reconciles pages, not stray data. Anything the projection never knew about — a node written directly to Neo4j,
   say, or a named graph left behind in a SPARQL store — is not something the rebuild can find. For a guaranteed-clean
   result, empty the backend before rebuilding: wipe Neo4j's data volume, or drop the wiki's named graphs from the store.
 - It runs as a single sequential process with no batching or resume, so the time scales with the number of pages. Plan
   downtime on large wikis.
+- Schema edits made while it runs do not reach it. It resolves each Schema once and keeps that copy for the rest of the
+  run, so a Schema edited mid-run is projected as it stood when the rebuild first read it. Edit Schemas before starting,
+  or re-run the rebuild afterwards.
```
JeroenDeDauw, line 35: Seems to detailed and arguably should be omitted entirely

alistair3149: Trim it to just "Schema edits made while it runs do not reach it."?
We can investigate into more graceful handling if needed.

JeroenDeDauw: Could do that trim. Could also just omit. Why would you make Schema edits during a rebuild? And if you did, you'd probably expect issues, as clearly the already rebuild stuff wont reflect your changes.

One argument for keeping is that AI Agent can cheaply hold it into account. But then AI Agent could just look at the code and get a better picture, so IMO makes sense to optimize for human readers.

alistair3149: > Why would you make Schema edits during a rebuild? And if you did, you'd probably expect issues, as clearly the already rebuild stuff wont reflect your changes.

I am not sure if that is an expected behavior for the wiki admin. Without the point being raised, users would expect that the schema change during the run would still be projected automatically or gracefully, instead of requiring another run.

JeroenDeDauw: they would expect it to be project for the part of the run that hasn't already happened. Point being is that presumably they would expect the run would not end up fully reflecting the new version of the schema. Which seems good enough.

spending too much time on this; either options is fine; I expect 0 real issues if we omit the line

JeroenDeDauw: best part is no part. if you never re-add something you are not removing enough

we'd also not have had this in there unless AI added it

alistair3149: Done in #1204

PR #1189, opened 2026-07-26
Add performance-testing tooling for big synthetic wikis

JeroenDeDauw:
[ADR 29](docs/adr/029-scalability-targets.md) sets Subject-count targets, and the first
measurements against them needed a throwaway harness to build big wikis. This makes that
capability a committed tool: generate a deterministic N-Subject wiki, time the import, and
snapshot the result so a large environment rehydrates into any worktree stack in minutes instead
of being regenerated or kept running.

## What lands

- `maintenance/GeneratePerformanceDump.php` writes a MediaWiki XML dump of synthetic Subject pages
  plus the Schema they use, so an import is self-contained. Options: `--pages` (required),
  `--subjects-per-page` (10), `--seed` (1), `--output` (stdout). The shape sits at the scale ADR 29
  calls typical: 10 Subjects per page (1 main + 9 children) and 12 Statements each, three of them
  relations to Subjects on other pages. Mixing negative and positive page offsets means the import
  meets both targets that already exist and targets it has to stub-create and later upgrade in
  place. Output is deterministic, and page titles and ids both carry the seed, so dumps from
  different seeds can be imported into one wiki.
- `make perf-generate`, `perf-import`, `perf-snapshot` and `perf-restore`. The first two follow the
  Makefile's dual-mode proxy convention and work from the host or inside the container; the other
  two drive the containers, so they are host-only. `perf-import` runs
  `importDump.php --no-updates` and prints elapsed time, pages/sec and Subjects/sec; that line is
  the benchmark. The snapshot is a `mariadb-dump` plus a `neo4j-admin database dump`, and restore
  loads both back.
- A README section under Development, and `perf/` gitignored.

The generator is PHP rather than a script in another language so it runs in-container with no host
dependencies.

## Verified end to end

On a dev stack: generate, import, snapshot, tear the stack down including volumes, reseed clean,
restore. The restored wiki matches the pre-wipe state exactly — 2085 Subject nodes, 8170
relationships, 321 pages — and the Subject REST endpoint answers for a generated id. Deleting 100
Subject nodes and restoring brings the count back. Three seeds imported into one wiki keep their
own page sets (20 + 20 + 50 pages).

Three behaviours the targets have to handle, all covered in the code and documented:

- `neo4j-admin` has to run as the `neo4j` user, not the root that `docker compose exec` gives you.
  Loading as root leaves a store the server cannot write, and the database then stays offline while
  `START DATABASE` still reports success — so both targets probe the database with a query and fail
  if it does not answer.
- Since #1185 merged, `update.php` creates the Neo4j uniqueness constraints, stack bring-up runs it,
  and a Neo4j dump carries them through a restore — so no flow measures an unindexed graph, and the
  README's earlier rebuild-first step is dropped.
- QLever is outside the snapshot, since its index is filled at runtime over SPARQL and persists
  across a restore. A restored wiki therefore keeps whatever the SPARQL store already held.

## Tests

`GeneratePerformanceDumpTest` covers the dump structure, Schema validity through
`SchemaContentValidator`, that the Schema defines exactly the properties the Subjects use and that
every property type is registered, statement counts through NeoWiki's own slot deserialization,
relation-target closure and off-page targeting at three run sizes, determinism, seed disjointness,
the highest accepted seed, and both rejected-option paths. Checked by mutation: breaking the
relation offsets or their clamp, the seed's contribution to ids or titles, one statement, or the
Schema page each fails the matching test and nothing else.

## Considered, omitted

- A `--statements-per-subject` knob. ADR 29 treats Statements per Subject as a dimension, but
  varying it means generating variable Property Definitions, and no measurement needs it yet.
- A round-trip import integration test. The import path is covered by the end-to-end run above;
  running MediaWiki's importer inside PHPUnit would be slow for what it adds.
- Adding `maintenance/` to the phpcs and phpstan targets. It is outside both today, so this script
  gets neither, but bringing the directory in scope would surface the two existing scripts as well.

> <sub>AI-authored — Claude Code, `Opus 5 (max)`; detailed spec from @JeroenDeDauw, no redirects; diff not yet human-reviewed; full PHPUnit suite and `make cs` green locally, CI green, plus generate/import/snapshot/wipe/restore runs against a live dev stack, a five-mutation check of the new tests, and an adversarial code and prose review whose findings are in the second and third commits.</sub>


JeroenDeDauw:
Not yet scrutinized by me

PR #1194, opened 2026-07-27
Stop the schema selector getting stuck in a pending state

malberts:
Follows-up to https://github.com/ProfessionalWiki/NeoWiki/pull/1174

One regression that PR introduced by moving the selector from CdxCombobox to CdxLookup,
plus a test for a line it left uncovered. It targets that branch rather than master because
master has neither.

CdxLookup treats a change to its menu items as the answer to the user's keystroke, and
marks the field as loading until one arrives. Its own source says as much: the menu's
pending indicator is "meant to indicate that new menu items are being fetched or computed",
and the watcher that ends it reads a menu-items change as "in response to user input". The
picker was not always answering. Its filter held the typed text trimmed and lower-cased, so
any edit that left that form unchanged wrote the same string, the menu-items computed did
not re-evaluate, and CdxLookup went on waiting: a striped background on a field that is
loading nothing, surviving blur and refocus, clearing only on a later edit that did reach
the menu. The same silence keeps the list shut until you leave the field and return. A
single space typed into an empty field reaches it, as does emptying the field after picking
a schema, since picking already resets the filter. Every edit now invalidates the menu,
including one that leaves the text exactly as it found it: pasting a name over its own
selection, or a cancelled composition.

Separately, resetting the filter when the surrounding editor commits a schema had no test.
Removing that line left every test passing, while the field went blank for a schema the
active filter hides.

Reproduced in a browser against the dev wiki, on the subject creator and on a relation's
target schema, and confirmed absent from `master`, where CdxCombobox has no loading state
at all and so carries no such contract.

Considered, omitted: a guard against Codex renaming `cdx-lookup--pending`, which the two
loading-state tests name. A healthy field carries no modifier class, so an absence
assertion cannot tell a fixed field from a renamed class, and the only way to pin the name
is a test of Codex rather than of this component. Worth revisiting if the class moves.

> <sub>AI-authored — Claude Code, `Opus 5 (max)`; found and fixed during a review of #1174 requested by @malberts, no redirections; diff not yet human-reviewed, though @malberts has independently reproduced the regression on both surfaces; reproduced in a real browser before and after each fix, every fix pinned by a test seen failing first and by a distinct killed mutant, and the frontend suite plus ts-lint and vue-tsc green locally. Adversarial review of an earlier round found that it closed the reported paths but still let an edit that changes nothing strand the field; that is closed and covered here.</sub>

### DIFF
--- a/resources/ext.neowiki/src/components/common/SchemaPicker.vue
+++ b/resources/ext.neowiki/src/components/common/SchemaPicker.vue
@@ -1,8 +1,10 @@
 <template>
 	<div class="ext-neowiki-schema-picker">
-		<CdxCombobox
-			ref="comboboxRef"
-			v-model:selected="selectedSchema"
+		<CdxLookup
+			v-if="schemasLoaded"
+			ref="lookupRef"
+			v-model:selected="pickedSchema"
+			v-model:input-value="inputText"
 			:menu-items="menuItems"
 			:placeholder="$i18n( 'neowiki-schema-picker-placeholder' ).text()"
 			@input="filterSchemas"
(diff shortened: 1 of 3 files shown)


malberts:
Fixes:
* Click Add Subject
* Type space
* Click away
* Input field keeps showing pending state animation

[Screencast_20260728_134410.webm](https://github.com/user-attachments/assets/bee74b17-579f-48dd-a726-8caf594245d4)

malberts:
Fixes:
* Edit schema
* Select different Target schema
* Select all via keyboard and press Delete or Backspace
* Click away
* Input field keeps showing pending state animation

[Screencast_20260805_202351.webm](https://github.com/user-attachments/assets/fd43a82b-74fd-4822-ace9-0aafd5997af8)

PR #1205, opened 2026-07-27
Point the demo-data guide at commands that work

JeroenDeDauw:
`DemoData/AGENTS.md` was added in #814 naming `load-test-data` and `reinstall-db`. Both had been deleted from the dev-environment repo's Makefile days earlier, and neither ever existed here as a target, so the guide's `make` instructions have never worked in the repo the file lives in. The current targets are `import-demo-data` and `reset`.

Corrected along with the names:

- `reset` reseeds the demo data — which the two-command form it replaced made visible and a bare `make reset` does not. "Full clean-slate" also over-claimed: the QLever index volume is declared in the dev compose file, so `make remove` never sees it and it survives.
- `rebuild-graph-databases` rebuilds every configured backend's projection, not only Neo4j.
- The wiki port is auto-allocated, so `8484` is the default rather than the address, and is wrong in every worktree.

`make reset` then turned out to abort before the reseed it now promises. `remove` loads the base compose file only, so `test-neo4j-data` survives it, and `setup-test-neo` failed issuing a bare `CREATE USER` against an instance that already had that user — taking `reset` down two steps before the import and leaving the wiki installed but empty. `IF NOT EXISTS` makes the step idempotent.

> <sub>AI-authored — Claude Code, `Opus 5 (max)`; one-line ask from @JeroenDeDauw, then an AI review pass and one round of adjustments at his request; diff not yet human-reviewed; ran `make reset` end-to-end before and after the Makefile fix (aborted leaving a 1-page wiki → exit 0, 120 pages reseeded), plus `make import-demo-data`, `make phpunit` (2143 pass) and `make cs` locally.</sub>

### DIFF
--- a/DemoData/AGENTS.md
+++ b/DemoData/AGENTS.md
@@ -159,12 +159,12 @@ From the repo root:
 # whose source files are gone. Enough on its own after renames or deletions.
 make import-demo-data
 
-# Full clean-slate reset (wipes the DB and Neo4j volumes, recreates the dev stack). Rarely needed.
+# Wipes the wiki database and Neo4j, then reseeds. Rarely needed.
 make reset
 
-# Reproject the Neo4j graph if Cypher results look stale.
+# Reproject the graph stores if Cypher or SPARQL results look stale.
 make rebuild-graph-databases
 ```
 
-A successful import ends with `Import finished` and zero `FAILED` lines. The wiki runs at
-`http://localhost:8484/`.
+A successful import ends with `Import finished` and zero `FAILED` lines. The wiki runs at the
+port `make dev` printed (default `http://localhost:8484/`).
--- a/DemoData/AGENTS.md
+++ b/DemoData/AGENTS.md
@@ -21,7 +21,7 @@ partners, knowledge managers, MediaWiki ecosystem evaluators, and live-demo audi
 
 `ImportDemoData.php` reseeds the demo set: it creates and updates pages from these directories, and
 deletes pages a previous import created whose source file is now gone. So renaming or removing a
-file and re-importing is enough — no `make reinstall-db` needed. Only pages the import itself created
+file and re-importing is enough — no `make reset` needed. Only pages the import itself created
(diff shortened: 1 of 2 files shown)


malberts:
`make reset` does not complete for me on this branch, but it stops before the step this PR fixes, so I
think this is an environment difference rather than anything in the diff.

Reproduced on two freshly bootstrapped worktrees, each running nothing but `make dev` and then
`make reset`. `make dev` is green both times: 119 pages, `Import finished`. Before the reset the stack
is fully healthy, including the Neo4j read path: `/wiki/Rijksmuseum` and `/wiki/Claude_Monet` return
200 with their infoboxes, `/wiki/Museum_collection` renders 10 rows in its Cypher table, and a
nonexistent title 404s. The first `make reset` after that exits 2 at `install-db`:

```
MediaWiki 1.43.9 Updater

Could not find composer.lock file. Have you run "composer install --no-dev"?
```

`load-neo4j-users`, `setup-test-neo` and `import-demo-data` are never reached, so the wiki comes back
installed with no demo data, and with `load-neo4j-users` skipped the Neo4j read user is missing and the
Cypher tables on the demo hubs render empty.

Second difference, and I think it is the same root: `make remove` here destroys all six project
volumes, `test-neo4j-data` included (measured, 6 before and 0 after). So the surviving-user abort this
PR fixes cannot occur on my setup at all, while you evidently hit it.

Both fit one explanation, that my teardown goes deeper than yours. `down --volumes` takes the dev-file
volumes, and the recreated `mediawiki` container loses the `composer.lock` that
`dev-entrypoint.sh:26-33` wrote into its writable layer, which the host-side marker
`Docker/mediawiki/vendor/.neowiki-merged` then stops it from rewriting.

For reference, this was on podman 5.8.4 with podman-compose 1.6.0 (Python provider, via the `docker`
shim).

> <sub>AI-authored — Claude Code, `Opus 5 (xhigh)`; review requested by @malberts; not human-reviewed;
> figures from local runs on this branch.</sub>

PR #1213, opened 2026-07-29
Support MediaWiki 1.44 through 1.47

malberts:
Makes the extension work on MediaWiki 1.44 through 1.47 and puts those versions under CI, which previously ran the minimum supported release and the development branch with nothing between them. Both defects fixed here live in that gap.

## CI

MediaWiki 1.46 removed the `tests/phpunit/phpunit.php` entry point the workflow invoked, so the runner becomes a composer script the workflow and the Makefile share. It runs unchanged on every targeted version: `composer phpunit:config` generates the config 1.46 and later require, and is tolerated as a no-op where that script does not exist.

The script keeps the extension's own `phpunit.xml.dist` and passes core's bootstrap explicitly, which the removed entry point used to inject. Selecting tests by path instead would inherit core's config, silently adopting its `forceCoversAnnotation`, `failOnWarning` and `failOnRisky` settings and dropping the `<env>` block that points the Neo4j and QLever tests at the right endpoints.

MediaWiki is cloned rather than unpacked, because `phpunit.xml.template` is `export-ignore`d and so absent from the archive while the generator that reads it is present. The `master` row is marked experimental so core breakage surfaces without blocking, matching the convention in Network, WikibaseFacetedSearch and Chameleon. PHPStan stays pinned to one version.

## Removed and changed MediaWiki APIs

`getText()` went in 1.45, and `setRawText()` and `getRawText()` in 1.47. `runOutputPipeline()`, `setContentHolderText()` and `getContentHolderText()` replace them and all three exist in 1.43, so no version branch is needed. Both forms drive the same output pipeline with the same stages and defaults, so the parser-function tests still assert against the same HTML.

Two constructor changes are decided by inspecting the API being adapted to rather than by a version string. Testing for the `ModuleManager` class would be wrong: it ships from 1.46, a release before `Router` accepts one, so the class is present while the old signature is still in force.

## Deprecations

Implicit nullable parameters gain their explicit `?Type`, required from PHP 8.4 and accepted since 7.1. `getRevisionByTitle()` receives a `Title` so it takes the `PageIdentity` path rather than the `LinkTarget` one deprecated in 1.45, and `SpecialNeoJson` overrides `isListed()` instead of passing the constructor flag deprecated in 1.46.

## Test environment

MediaWiki refuses to give an IP address an actor once temporary accounts are on, which the 1.47 test wiki has, so every anonymous edit the suite made threw. It now uses MediaWiki's `TempUserTestTrait` to pin the feature off. Whether NeoWiki behaves correctly *for* a temporary account is a separate question this does not answer.

`SpecialNeoJsonTest` no longer supplies its own `wpEditToken`. MediaWiki 1.44 moved `HTMLForm`'s CSRF check from the user's session onto the form context's request, which is the session `SpecialPageExecutor` mints its token from, and a form that rejects the token re-renders silently rather than raising. The rate-limit test was therefore green without the write it names ever running.

`markPageTableAsUsed()` is gone. `$tablesUsed` has been deprecated since 1.41 and 1.43 already detects changed tables by itself, so registering one had no effect on any supported version.

`NeoWikiLibraryTest` names its Lua engine. Scribunto's `LuaEngineTestBase` multiplexed a test class across every available engine through a static `suite()`; from 1.46 that mechanism is gone, `makeSuite()` returns an empty suite, and each concrete class names its engine through `getEngineName()`, which throws until overridden. The Lua cases therefore skipped rather than ran on 1.46 and master, and the suite still reported OK. LuaStandalone is the engine named because it needs no PHP extension and so is present wherever the suite runs. One class rather than a subclass per engine: a LuaSandbox subclass would only ever skip in CI, which installs no `luasandbox`. Scribunto 1.43 through 1.45 have no `getEngineName()` and still choose engines through `suite()`, so they are unaffected.

## Not covered

Four pre-existing issues this work surfaced are left for separate changes: `SpecialNeoJson` authorizes on its own authority but writes as the main request context's; the `onSubmit()` authorization rejection branch is unreachable from the current suite; NeoWiki's behaviour for a temporary account is untested; and `ModuleSpecHandlerNeoWikiTest` is `@coversNothing` because much of it asserts MediaWiki's contract rather than ours.

> <sub>AI-authored — Claude Code, `Opus 5 (xhigh)`; from an open question by @malberts on unblocking CI, steered over ~14 turns, with the scope split out of the now-closed #862 and the commits regrouped on request; UI spot-checked by @malberts, diff not line-reviewed; suite run locally against MediaWiki 1.43, 1.46 and 1.47, Playwright UI smoke tests on all three, CI green on all five matrix rows.</sub>

### DIFF
--- a/.github/workflows/ci-php.yml
+++ b/.github/workflows/ci-php.yml
@@ -21,12 +21,27 @@ jobs:
   test:
     name: "PHPUnit: MW ${{ matrix.mw }}, PHP ${{ matrix.php }}"
 
+    continue-on-error: ${{ matrix.experimental }}
+
     strategy:
       fail-fast: false
       matrix:
         include:
           - mw: 'REL1_43'
             php: '8.3'
+            experimental: false
+          - mw: 'REL1_44'
+            php: '8.3'
+            experimental: false
+          - mw: 'REL1_45'
+            php: '8.3'
+            experimental: false
+          - mw: 'REL1_46'
+            php: '8.4'
+            experimental: false
+          - mw: 'master'
+            php: '8.5'
+            experimental: true
 
     runs-on: ubuntu-latest
 
@@ -76,7 +91,7 @@ jobs:
             mediawiki
             !mediawiki/extensions/
             !mediawiki/vendor/
-          key: mw_${{ matrix.mw }}-php${{ matrix.php }}
+          key: mw_${{ matrix.mw }}-php${{ matrix.php }}-v2
 
       - name: Cache Composer cache
         uses: actions/cache@v5
@@ -146,7 +161,8 @@ jobs:
           echo "QLever did not become ready in time; dumping logs:"; docker logs test_qlever; exit 1
 
       - name: Run PHPUnit
-        run: php tests/phpunit/phpunit.php -c extensions/NeoWiki/phpunit.xml.dist
+        run: composer phpunit
+        working-directory: mediawiki/extensions/NeoWiki
 
   phpcs:
     name: "PHPCS with PHP 8.3"
@@ -203,7 +219,7 @@ jobs:
             mediawiki
             !mediawiki/extensions/
             !mediawiki/vendor/
-          key: mw_${{ matrix.mw }}-php${{ matrix.php }}
+          key: mw_${{ matrix.mw }}-php${{ matrix.php }}-v2
 
       - name: Cache Composer cache
         uses: actions/cache@v5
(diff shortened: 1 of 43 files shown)


malberts:
Not manually reviewed yet, and not manual tested, just a few random spot checks in a real MW 1.47 environment:

[Screencast_20260729_235328.webm](https://github.com/user-attachments/assets/6da9ef88-a898-4f65-91d4-c163a61c9247)

PR #1215, opened 2026-07-29
Give the frontend stores registry semantics (#1211)

### LINKED ISSUE #1211: Frontend stores act as caches without an invalidation model
Follow-up to #1177 / #1182.

## Two omissions, copied three times

`SchemaStore`, `LayoutStore` and `SubjectStore` each hold copies of server state and each hand-implement the same `fetchX` / `getOrFetchX` / `saveX` trio, copied between them with the same two omissions in every copy:

1. **A read assigns its awaited result to shared state unconditionally**, with no check that the value is still wanted. This is #1177: a summaries pagination that started before a save writes its pre-save list into the cache after the save cleared it.
2. **A write invalidates only on its own success path, and only the state that action happens to know about.**

## Concrete defects on master

- [ ] `SchemaStore.saveSchema` invalidates after a bare `await`, so a rejected save invalidates nothing. `MediaWikiPageSaver.savePage` rejects on the HTTP response, not on revision state, so a save that committed and lost its reply leaves a stale list for the rest of the session. Move the two lines into a `finally`.
- [ ] `LayoutStore` never invalidates `layouts`. Not on save, not on delete, nowhere in `src`. `saveLayout` writes the client's own object into the map and stops.
- [ ] `SubjectStore.deleteSubject` removes from `subjects` and leaves `pageSubjects` listing the deleted Subject — which is why `SubjectsManagerPage.vue` falls back to `window.location.reload()`, with a comment conceding the point. Worth noting because this delete *is* store-routed: routing deletes through a store does not by itself fix this.
- [ ] `DeletePageDialog` posts `action=delete` via `mw.Api` and imports no store or repository. It is the delete path for Schemas, Layouts *and* Mappings, so a deleted Schema keeps appearing in pickers. All three `@deleted` handlers wire only a private refetch.

## No decision was ever recorded

[ADR 016](docs/adr/016-frontend-state-management.md) settles *which components read the stores* (props-down/events-up for editing UIs, stores for display UIs) and describes the result as "one page-wide set of state ... used by everything that needs to be fully reactive". That is a registry, not a cache. Fetch-on-miss and write-through-on-save accreted afterwards with nothing written down, and two of the three copies have no invalidation tests at all.

## The caching is already worked around

- `getOrFetchSchema` — the cache-consulting read — has **two** callers (`SubjectCreatorDialog`, `SubjectEditorDialog`).
- `fetchSchema` — an unconditional refetch that ignores the cache — has **four** (`SchemaDisplay`, `SchemasPage`, `SubjectsManagerPage`, `Infobox`). `SubjectsManagerPage.openEditor` says why: *"Refresh both subject and schema so the editor never opens against stale data."*
- `getOrFetchLayout` has **no** callers outside the store; `layouts` is seeded by `StoreStateLoader` and read.
- `SchemasPage`, `LayoutsPage` and `MappingsPage` skip the stores entirely and hand-build the same REST URL as `RestSchemaRepository.getSchemaSummaries`.
- Two flows use `window.location.reload()` as their invalidation primitive.

So two directions:

**A. Make the cache correct.** Add a generation check to every write-back, make every mutation invalidate — including deletes and failures — and route the writers outside the store (`StoreStateLoader`, and `SubjectStore.loadPageSubjects` calling `schemaStore.setSchema`) through the guarded path. The guard pattern already exists in-house and is tested: `useSubjectValidation` and `SubjectLookup` both capture a monotonic sequence before the await and drop the write if it moved. Downside: it automates an obligation rather than removing it, and every future `fetchX` has to remember it.

**B. Drop the cache semantics, keep the registry.** Keep the page-scoped shared state ADR 016 actually describes; keep in-flight dedup for concurrently mounting relation pickers, which is the one piece doing real work; delete `getOrFetch*` and the save write-through. Costs no extra requests at today's call sites, since both `getOrFetchSchema` callers are already freshly seeded and the server sends the page's schemas with the payload.

I lean B, with one guard kept inside `setX` so a late refetch cannot clobber a newer value. Either way the decision belongs in a revision to ADR 016 — its silence on this is what let the template be copied three times.

Adding cache headers to the REST read endpoints (#1212) is worth doing but is not an alternative to either direction: these are ordering and consistency defects, and HTTP caching neither orders responses nor dedupes concurrent ones.

> <sub>AI-authored — Claude Code, `Opus 5 (1M context)`; written from a review of #1182 plus a read-only audit of the store layer by several agents, with an adversarial pass that corrected an over-broad first diagnosis; text not yet human-reviewed; every code claim above re-verified directly against `origin/master`, no runtime reproduction attempted.</sub>

alistair3149:
## Summary

Frontend stores (`SchemaStore`, `LayoutStore`, `SubjectStore`) treated themselves as caches with
no invalidation model — reads served stale values to avoid a fetch, writes only invalidated their
own success path, and deletes invalidated nothing. This gives them registry semantics instead,
per the approved [#1211](https://github.com/ProfessionalWiki/NeoWiki/issues/1211) design (see
[ADR 30](docs/adr/030-frontend-store-registry-semantics.md)):

1. **Reads are explicit.** `fetchX` always hits the network and records the result; `getX` reads
   the registry synchronously. No fetch-on-miss, with two exceptions:
   - `SubjectStore.getOrFetchSubject` — display-only label resolution over the page payload's
     seeded, referenced Subjects (avoids N requests for N relation displays).
   - `SchemaStore.fetchAllSchemaSummaries` — concurrent callers share one in-flight pagination;
     results are not retained, so each new caller cohort is fresh.
2. **Mutations keep the registry own-write consistent on every path.** Each store has a
   `mutationEpoch`, bumped when a write is acknowledged. Saves write through; deletes remove from
   all registry state; invalidation that must survive a rejected save happens in `finally`.
3. **Async write-backs are epoch-guarded.** Every fetch that writes into a store snapshots the
   store's `mutationEpoch` before its await and discards the result if the epoch moved, since a
   response that raced a mutation may predate it.

`getOrFetchSchema` and `getOrFetchLayout` are gone from the public API; callers now fetch
explicitly and read the registry. The persistent schema-summaries cache (`allSummaries`) is
dropped in favor of in-flight dedup only. Deletes (Schema, Layout, Subject) now remove the entity
from all registry state instead of relying on a page reload, and the dead
`window.location.reload()` / pending-notification write for Subject deletion is removed
accordingly.

Fixes #1211

## Manual Browser Check

1. Open `Special:Schemas`, create a schema `PickerFreshness`, then open any page's
   subject creator (Add subject): the picker lists `PickerFreshness` without a reload.
2. Delete `PickerFreshness` from `Special:Schemas`, reopen the subject creator: the
   picker no longer lists it.
3. On a page with subjects, open the manage-subjects view (`?action=subjects`), delete a
   child subject: the row disappears and a success notification shows, with no page reload.
4. Edit a schema used by an infobox on another tab, then on the first tab open that
   infobox's subject editor: the editor shows the updated schema (fresh fetch on open).
5. Open the schema editor for a schema with two or more relation properties: the network
   tab shows one summaries pagination, not one per picker.

<sub>AI-authored — Claude Code, Fable 5; implementation from the approved #1211 design (direction B refined); verified via the tsci checks run individually (vitest 1168/1168 with a capped worker pool, vite build, eslint — the single make tsci invocation OOMs in this dev sidecar) and the manual browser checks above.</sub>

### DIFF
--- a/docs/adr/016-frontend-state-management.md
+++ b/docs/adr/016-frontend-state-management.md
@@ -2,7 +2,7 @@
 
 Date: 2025-04-17
 
-Status: Accepted
+Status: Accepted. Extended by [ADR 30](030-frontend-store-registry-semantics.md), which adds fetching and invalidation semantics.
 
 ## Context
 
--- a/docs/adr/016-frontend-state-management.md
+++ b/docs/adr/016-frontend-state-management.md
@@ -2,7 +2,7 @@
 
 Date: 2025-04-17
 
-Status: Accepted. Extended by [ADR 30](030-frontend-store-registry-semantics.md), which adds fetching and invalidation semantics.
+Status: Accepted. Amended by [ADR 30](030-frontend-store-registry-semantics.md), which adds fetching and invalidation semantics and scopes the editing-UI rule to reads.
 
 ## Context
 
--- a/docs/adr/030-frontend-store-registry-semantics.md
+++ b/docs/adr/030-frontend-store-registry-semantics.md
@@ -2,115 +2,100 @@
 
 Date: 2026-07-31
 
-Status: Accepted
+Status: Draft
 
-Amends [ADR 16](016-frontend-state-management.md), which settles which components read the
-stores but is silent on fetching, caching and invalidation: its editing-UI rule now applies
-to reads, while mutations and validation still go through store actions.
+Amends [ADR 16](016-frontend-state-management.md), which settles which components read the stores but is silent on
+fetching, caching and invalidation. Its editing-UI rule now applies to reads; mutations and validation still go through
+store actions.
 
 ## Context
 
-The frontend keeps server state in Pinia stores (`SchemaStore`, `LayoutStore`, `SubjectStore`).
-Each store grew the same trio of actions: `fetchX`, `getOrFetchX` and `saveX`. The trio treated
-the store as a cache: reads consulted stored values to avoid fetches, and saves wrote the
-client's object through. An invalidation model was never decided
-([#1211](https://github.com/ProfessionalWiki/NeoWiki/issues/1211)). Reads assigned awaited
-results to shared state unconditionally
-([#1177](https://github.com/ProfessionalWiki/NeoWiki/issues/1177)), writes invalidated only on
-their own success path, and deletes invalidated nothing. The components already routed around
-the cache: editing flows fetched fresh data before opening, and two flows reloaded the whole
-page as their invalidation primitive.
+The frontend keeps server state in Pinia stores (`SchemaStore`, `LayoutStore`, `SubjectStore`). Each store grew the
+same trio of actions — `fetchX`, `getOrFetchX` and `saveX` — treating the store as a cache: reads consulted stored
+values to avoid fetches, and saves wrote the client's object through. An invalidation model was never decided
+([#1211](https://github.com/ProfessionalWiki/NeoWiki/issues/1211)). Reads assigned awaited results to shared state
+unconditionally ([#1177](https://github.com/ProfessionalWiki/NeoWiki/issues/1177)), writes invalidated only on their
+own success path, and deletes invalidated nothing. The components already routed around the cache: editing flows
+fetched fresh data before opening, and two flows reloaded the whole page as their invalidation primitive.
 
-A client-side cache can only be consistent with the session's own writes. Edits from other tabs
-and other users stay invisible until the client fetches again. Query libraries such as TanStack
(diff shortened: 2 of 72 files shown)


JeroenDeDauw:
Follow-up targeting this branch: https://github.com/ProfessionalWiki/NeoWiki/pull/1222 — it moves the editing reads off the stores onto the repositories (per ADR 16) and shrinks ADR 30 accordingly. Merge-order note: since we squash-merge, #1222 should be merged into this branch before this PR lands (or be rebased onto master afterwards).

PR #1216, opened 2026-07-30
Add REST endpoints for setting and removing one Statement

### LINKED ISSUE #591: REST API endpoint for updating a single Statement

malberts:
Fixes https://github.com/ProfessionalWiki/NeoWiki/issues/591

Third-party integrators can now change a single property without fetching and re-sending the whole Subject:

```
PUT    /neowiki/v0/subject/{subjectId}/statements/{propertyName}
DELETE /neowiki/v0/subject/{subjectId}/statements/{propertyName}
```

Both reuse the machinery behind `PUT /subject/{subjectId}`: its read-then-write authorization, select-value
resolution, and validation with enforcement. They answer with the same body it does, carrying the persisted Subject
and its Schema, and `422` under enforcement. Violations always cover the whole Subject, not just the property
written.

## Design decisions

**The body nests the Statement under a `statement` key** rather than putting `propertyType` and `value` at the top
level as the issue sketched:

```json
{ "statement": { "propertyType": "url", "value": ["https://example.com"] }, "comment": "..." }
```

MediaWiki's REST body validator types every declared field, and it has no type that accepts both the scalars and
the arrays a Statement value can be (`ArrayDef` rejects `2019`, and there is no "any" type). Every other endpoint
that carries values nests them inside an `array`-typed param for the same reason, so a top-level `value` would have
had to either reject numbers and booleans or bypass validation. The nesting also keeps the edit summary out of the
Statement's own field namespace.

**`propertyType` is optional** (the issue's first open question). When omitted, the type the Subject's Schema
currently gives the property is stored as the writer's schema, which is the honest answer for a client that did not
state one. A property the Schema does not define, sent without a type, is rejected with `400` rather than guessed
at. An explicitly sent type still wins, so writer's-schema drift stays expressible.

**A value that is empty for its type removes the Statement**, matching how the whole-Subject write treats one, so a
client that clears a field gets the same result on either endpoint. `DELETE` on a Statement the Subject does not
have succeeds and writes nothing; MediaWiki's null-edit handling means no revision is created.

**`UpdateStatementAction` builds the proposed Subject as a copy** (`Subject::withStatements`) instead of mutating
the loaded one, so a write that enforcement rejects leaves the Subject untouched by construction rather than by the
save never being reached.

## Type-mismatched values

Exercising the new endpoint surfaced a pre-existing hole in the shared `StatementListBuilder`: a value of the wrong
shape for the type it declares reached the typed value-object constructors, and the TypeError escaped as a `500`.
It was equally reachable through `PUT /subject/{subjectId}` before this branch, for a number sent to a `text`
property, a string sent to `number` or `boolean`, a scalar sent to `relation`, or a Statement carrying no `value`.

The builder now reports those as bad input, so both write endpoints answer `400` with the offending property and
type named. A new public endpoint that `500`s on a wrong-shaped value is not shippable, and putting the check
anywhere but the builder would duplicate the type knowledge that lives there, so the fix is in the shared class and
`PUT /subject/{subjectId}` improves with it.

## Answering like the whole-Subject write

Two write-path changes landed on master while this branch was open, and the single-Statement endpoints follow the
endpoint they sit beside on both.

A Subject on a page the caller cannot read answers not-found rather than reaching the write check, which would
answer `403` where a restricted page answers `404` and so tell a caller lacking the wiki-global `edit` right which
of the Subject ids they hold exist.

Write responses carry the persisted Subject and the Schema it instantiates. That matters more here than on the
whole-Subject write: a client using these endpoints has deliberately not fetched the Subject, so without the
bundle it would have to read one back.

## Property names in the path

`{propertyName}` is URL-encoded, so `Founded at` is `Founded%20at`. A property name containing `/` is not
addressable this way on servers that reject `%2F` in paths; the docs point those cases at `PUT /subject/{subjectId}`.

## Verification

Alongside the unit and integration tests, the endpoints were exercised over real HTTP against a local dev stack,
which covers the route registration, path decoding, and CSRF handling that handler tests bypass: set with an
explicit type, set with the type omitted, set through a URL-encoded multi-word property name, delete, delete again
(no second revision), `400` for an undefined property with no type, `400` for each type-mismatched value shape,
`404`, `403`, and `400` for a missing `statement`. Statement order and the Subject's other statements survived each
write, and the new routes appear in the generated OpenAPI spec. That run predates the rebase; the write-path
alignment above is covered by tests only.

Every new test was checked against mutations of the code it covers, including full removal of the set, remove,
authorization, read-gate, enforcement, select-resolution, property-type-fallback, and response-bundling logic.

Considered, omitted: a dedicated `RestUpdateStatementPresenter` unit test, since the API tests assert the presenter's
whole contract (both status codes and body shapes) through the real object.

### DIFF
--- a/docs/api/rest-api.md
+++ b/docs/api/rest-api.md
@@ -29,8 +29,8 @@ pagination.
 Subject write endpoints require per-page `edit` permission and answer `403` when you may read the page but not edit it.
 Denial of `read` answers `404` instead, so that a page you may not read stays indistinguishable from one that is absent.
 The write endpoints keyed by page id ([Pages and Subjects](#pages-and-subjects)) return that `404` for a page you may
-not read and for a page id that does not exist; `PUT /neowiki/v0/subject/{subjectId}` returns it for a Subject on a page
-you may not read and for a Subject id that does not exist.
+not read and for a page id that does not exist; the write endpoints keyed by Subject id return it for a Subject on a
+page you may not read and for a Subject id that does not exist.
 
 The Cypher query endpoint is gated only by the `neowiki-query` right, with no per-page filtering (see
 [Query API](query-api.md)).
--- a/docs/api/rest-api.md
+++ b/docs/api/rest-api.md
@@ -55,6 +55,8 @@ Read, change, and validate Subjects. New Subjects are created on a page — see
 | `GET /neowiki/v0/entity/{subjectId}` | Dereference a Subject's concept URI. `303` to the Subject's RDF (`Accept: application/trig` or `text/turtle`) or to the hosting page (otherwise). See [Dereferencing subject IRIs](../rdf/rdf-export.md#dereferencing-subject-iris). |
 | `PUT /neowiki/v0/subject/{subjectId}` | Replace a Subject's label and statements. |
 | `DELETE /neowiki/v0/subject/{subjectId}` | Delete a Subject. |
+| `PUT /neowiki/v0/subject/{subjectId}/statements/{propertyName}` | Set one Statement, leaving the Subject's label and other Statements as they are. |
+| `DELETE /neowiki/v0/subject/{subjectId}/statements/{propertyName}` | Remove one Statement. |
 | `POST /neowiki/v0/subject/validate` | Check whether a new Subject is valid, without saving it. Returns `{violations: [...]}` — see [Validation codes](validation-codes.md). |
 | `POST /neowiki/v0/subject/{subjectId}/validate` | Check whether a change to a Subject is valid, without saving it. Returns `{violations: [...]}` — see [Validation codes](validation-codes.md). |
 | `POST /neowiki/v0/subject-ids` | Mint a batch of unused Subject IDs to assign on create, e.g. to wire relations across an interlinked import. Body `count` (1–1000). |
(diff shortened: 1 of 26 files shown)


JeroenDeDauw:
Conflicts with master

JeroenDeDauw:
LGTM w/ https://github.com/ProfessionalWiki/NeoWiki/pull/1270

PR #1219, opened 2026-07-30
Treat an empty color part as absent rather than malformed

### LINKED ISSUE #1208: RedHerb ColorType treats an empty string part as malformed rather than absent
Found while reviewing #1199. Real but low severity, and **not** introduced by that PR.

## What

`ColorType::validate` tests emptiness as `$value->strings === []`. A one-element `StringValue( '' )` (or a whitespace-only part) is therefore not `[]`, so it skips the `required` branch, falls into the hex loop, fails `HEX_COLOR_REGEX`, and raises `invalid-color` — which since #1199 carries a fixed `Severity::Error` and blocks under `$wgNeoWikiEnforceValidation`.

An empty string part is *absent* data, not *malformed* data. Every core string-valued type classifies it that way:

| type | `''` under an unannotated `required` |
|---|---|
| text | `required` / warning |
| url | `required` / warning |
| date | `required` / warning |
| dateTime | `required` / warning |
| **color** | **`invalid-color` / error, blocking** |

`TextType` and `UrlType` get there via a trim-based `$hasContent` plus skipping empty parts in the validation loop; `DateType`/`DateTimeType` via `extractFirstString`, which returns null when part 0 trims empty. `SelectType` shares ColorType's unguarded `=== []` shape but escapes blocking because its per-part code takes `severityOf( 'options' )` rather than a hardcoded severity.

Two further consequences: it also fires on an *optional* property (the `required` branch is skipped either way, so no severity annotation changes the outcome), and a padded-but-valid `' #aabbcc '` blocks where `UrlType` would trim first.

## Why it is low severity

- No shipped client can produce it. `newStringValue` (`resources/ext.neowiki/src/domain/Value.ts`) does `.map( trim ).filter( part => part !== '' )`, so a cleared field serialises to `null` and the statement is dropped.
- `$wgNeoWikiEnforceValidation` defaults to `false`.
- `color` is defined only by RedHerb, which is autoloaded under `TestAutoloadNamespaces`.

Reachable from a hand-written REST client or an import path only.

## Why fix it anyway

`docs/api/validation-codes.md` names ColorType the worked example for plugin authors adding a violation code. A wrong emptiness test propagates from there into third-party Property Types that *will* run under enforcement.

## Not a regression from #1199

Before #1147, `isBlocking()` was a blocklist of three codes not including `invalid-color`, so this input was rejected then too. It became a silent warning during #1147 and blocks again after #1199. The genuinely new state is an intra-type split that existed in neither era: on the same property, clearing to `[]` or omitting gives `required`/warning and persists, while clearing to `""` gives `invalid-color`/error and 422.

**Do not fix this by reverting the `Severity::Error` stamp** — that is what ADR 26 mandates for a fixed system condition, and reverting would leave RedHerb inconsistent with every core type.

## Fix

Mirror `UrlType` on both halves — a trim-based `hasContent` for the `required` test, and `continue` on an empty part inside the loop. Doing only the second half leaves a required-but-empty color with no violation at all.

`ColorTypeValidateTest` has no empty-part coverage (every `StringValue(` call site is `()` or a literal). `UrlTypeValidateTest` has the three cases to copy: `testRequiredAndSingleEmptyStringPartReturnsRequiredViolation`, `testRequiredAndWhitespaceOnlyPartReturnsRequiredViolation`, `testEmptyStringPartIsSkippedWithNoViolation`.

malberts:
Fixes https://github.com/ProfessionalWiki/NeoWiki/issues/1208

RedHerb's ColorType tested emptiness with `$value->strings === []`, so a one-element StringValue holding `''` or
whitespace was neither empty nor a valid hex color: it skipped the `required` branch and raised `invalid-color`, a
code `Violation::isBlocking()` treats as blocking, so the write is rejected under `$wgNeoWikiEnforceValidation`.

An empty string part is absent data, not malformed data. Every core string-valued type classifies it that way: `text`
and `url` via a trim-based content check plus skipping empty parts, `date` and `dateTime` via `extractFirstString`.
ColorType now mirrors `url` on both halves, which also stops a whitespace-padded but otherwise valid `' #aabbcc '`
from being rejected. What the client sent is still what comes back in the violation `args`: the trimmed value is used
for the hex check and the allow-list lookup only, since `invalid-option` is a core code documented as carrying the
offending part and core SelectType passes it unchanged.

For a required property the write was already rejected either way, since `required` is blocking too; what changes
there is that the reported code now names the actual problem. The rejection itself goes away for an optional
property, and for a padded but valid color.

ColorType is the worked example that `docs/api/validation-codes.md` points plugin authors at, so a wrong emptiness
test propagates from there into third-party Property Types that do run under enforcement.

The content check lives in a `hasContent()` helper instead of being inlined the way UrlType has it, to stay under the
cyclomatic complexity limit phpcs enforces.

Note for whoever lands #1147: that branch rewrites the same three lines (adding `severity:` arguments) and still
carries the `=== []` test, so a conflict resolved in its favour silently undoes this.

> <sub>AI-authored — Claude Code, `Opus 5`; `/fix-issue` run against the issue as written, no redirection; diff not yet human-reviewed; new tests seen failing before the fix and passing after, mutation-tested, full PHPUnit suite plus phpcs and phpstan green locally and on CI.</sub>

### DIFF
--- a/tests/RedHerb/src/ColorType.php
+++ b/tests/RedHerb/src/ColorType.php
@@ -66,7 +66,7 @@ class ColorType implements PropertyType {
 				$violations[] = new Violation(
 					propertyName: null,
 					code: 'invalid-color',
-					args: [ $color ],
+					args: [ $part ],
 					valuePartIndex: $index,
 				);
 				continue;
@@ -76,7 +76,7 @@ class ColorType implements PropertyType {
 				$violations[] = new Violation(
 					propertyName: null,
 					code: 'invalid-option',
-					args: [ $color ],
+					args: [ $part ],
 					valuePartIndex: $index,
 				);
 			}
--- a/tests/RedHerb/src/ColorType.php
+++ b/tests/RedHerb/src/ColorType.php
@@ -46,7 +46,7 @@ class ColorType implements PropertyType {
 			return [];
 		}
 
-		if ( $definition->isRequired() && $value->strings === [] ) {
+		if ( $definition->isRequired() && !$this->hasContent( $value ) ) {
 			return [ new Violation( propertyName: null, code: 'required' ) ];
 		}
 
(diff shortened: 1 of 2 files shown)


JeroenDeDauw:
Closing together with https://github.com/ProfessionalWiki/NeoWiki/issues/1208: https://github.com/ProfessionalWiki/NeoWiki/pull/1290 made the empty-part half of this unreachable (StringValue canonicalizes at construction), and the padded-value half is strictness we are accepting in the example type for now.

PR #1221, opened 2026-07-31
Refresh npm lockfiles weekly

JeroenDeDauw:
Follows-up to https://github.com/ProfessionalWiki/NeoWiki/pull/1220

Adds a scheduled workflow that runs `npm update` against resources/ext.neowiki and tests/RedHerb, verifies the result,
and opens a lockfile-only pull request when anything moved.

Dependabot raises pull requests for dependencies named in a manifest. Nested transitive packages keep whatever version
their parent last resolved, so they drift until an advisory lands on them — as happened with brace-expansion and
js-yaml, whose fixes were already inside the ranges the manifests declare.

The job runs `npm ci`, the tests, the build and the linters against the refreshed tree before opening the pull request,
so a broken refresh leaves no pull request behind. The pull request itself carries no checks, because GitHub does not
start workflow runs for events raised with `GITHUB_TOKEN`.

> <sub>AI-authored — Claude Code, `Opus 5 (max)`; asked to automate the dependency refresh after the manual bump in #1220; no revisions; diff not yet human-reviewed; the job's own logic was dry-run against master in a container (npm update, then npm ci, tests, build and linters for both projects, all passing), but the workflow has not yet run on GitHub.</sub>

### DIFF
--- a/.github/workflows/refresh-npm-lockfiles.yml
+++ b/.github/workflows/refresh-npm-lockfiles.yml
@@ -1,0 +1,94 @@
+name: Refresh npm lockfiles
+
+# Runs `npm update` against both npm projects so transitive packages move to their
+# latest in-range release. Dependabot opens pull requests for dependencies named in
+# a manifest; nested transitive packages otherwise sit at whatever version was
+# resolved when their parent was last touched, until an advisory lands on them.
+#
+# The job verifies the refreshed tree before opening the pull request. That pull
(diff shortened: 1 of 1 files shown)


JeroenDeDauw:
Superseded by https://github.com/ProfessionalWiki/NeoWiki/pull/1223, which gets the same weekly lock file refresh from Renovate's `:maintainLockFilesWeekly` instead of a bespoke workflow.

PR #1222, opened 2026-07-31
Editing reads via repositories; Subject writes return the persisted state

JeroenDeDauw:
Follows-up to https://github.com/ProfessionalWiki/NeoWiki/pull/1215

## Summary

Editing reads already ran through store fetch actions before #1215; keeping them there is what forced the epoch
guard's "resolved but may have recorded nothing" contract onto every fetch, and a stale-data error toast on
Special:Layouts for a situation users should never see. This PR makes two moves, recorded in
[ADR 30](https://github.com/ProfessionalWiki/NeoWiki/blob/refactor/1211-reads-via-repositories/docs/adr/030-frontend-store-registry-semantics.md)
(which now amends
[ADR 16](https://github.com/ProfessionalWiki/NeoWiki/blob/refactor/1211-reads-via-repositories/docs/adr/016-frontend-state-management.md):
its editing-UI rule applies to reads; mutations and validation still go through store actions):

1. **Editing reads go through the repositories.** `fetchSchema`, `fetchLayout` and `fetchSubject` are gone from
   the stores; the code that opens an editor fetches via the repository and passes props. `SubjectEditorDialog`
   now takes the schema as a required prop — breaking for third-party hosts of the exported dialog. It mirrors
   the prop internally, so a schema saved through the nested schema editor applies without the host passing a
   new one down. `getOrFetchSubject` keeps its epoch guard inline on the miss path; `SubjectCreatorDialog`
   picks up a `requestSequence` guard for its now-local schema fetch. Components acquire repositories via
   `NeoWikiServices` injection (the pre-existing idiom); `Service.SubjectRepository` is newly registered.
2. **Subject writes answer with the persisted state.** The create and replace endpoints return the canonical
   Subject as stored — byte-identical to the read endpoints' output, asserted end to end — plus the Schema it
   instantiates, so clients need not re-fetch after a write. Additive for existing API consumers; documented in
   [`docs/api/subject-format.md`](https://github.com/ProfessionalWiki/NeoWiki/blob/refactor/1211-reads-via-repositories/docs/api/subject-format.md).
   Replace also gains the read gate create already had, so an unreadable page is indistinguishable from an
   absent one on every subject write.
3. **The registry holds only server-produced Subjects.** Store write-throughs record the mutation response's
   entity — which carries the page context the editor's client-built copy lacks, closing the bug class where a
   save put a context-less Subject in the registry and broke relation-label rendering — and the bundled Schema
   lands epoch-guarded. Schema and Layout saves still write the authored object through; the editor is its sole
   author. Displays catch up on commit: a schema changed in another session renders after your next save, and
   cancelling an editor leaves the display on page-load state.

#1215's mutation core is untouched: `removeSchema`/`removeLayout`, the `deleteSubject` re-sync invariant, the
summaries in-flight dedup, and the epoch guards on the seeding paths all stay as they are on the base branch.

Also fixed on the way, previously unreported: opening an editor on a subject and then deleting that subject
crashed the manage-subjects page (`editingSubject` was a store-read computed over a possibly-deleted id; it is
now local state), and an editor save could revert a label renamed in another tab whenever the epoch guard
discarded `openEditor`'s refresh (the save now starts from the fetched subject). The
`neowiki-layouts-edit-stale-error` message is removed together with the code path that showed it, and infobox
editors now run their validation dry-run when opened (the dialog mounts open, which the existing dry-run
watcher was written for).

## Merge notes

- Squash workflow: merge this into the #1215 branch before that PR lands (or rebase onto master after).
- Merging to master conflicts in `RestCreateSubjectPresenterTest.php` / `RestReplaceSubjectPresenterTest.php`:
  master's #1147 added `severity` to the expected violation arrays these tests rewrite. Keep both sides — the
  new `subject`/`schema` assertions *and* the `severity` keys.

## Manual browser check

1. Infobox pencil → edit → save: the new values render immediately, no reload.
2. In a second tab, add a property to the schema; in the first tab open the infobox editor, fill the new field,
   save: the value renders without a reload. Cancel instead of save: the display keeps its page-load state.
3. Special:Schemas / Special:Layouts: edit opens on fresh data.
4. `?action=subjects`: edit and delete a child subject; the listing stays correct with no reload.

<sub>AI-authored — Claude Code; implementation and fixes `Opus 5 (max)` from written specs; independent AI code-review and text-review passes (`Opus 5 (max)`) with findings applied; orchestration and final review `Fable 5 (max)`. PHPUnit (2190) + phpcs/phpstan + vitest (1197) + build + lint green locally and in CI; the review pass verified the save/cancel flows over live HTTP and in the browser, including the byte-identity of write responses with the read endpoints on a namespaced page.</sub>

### DIFF
--- a/docs/adr/016-frontend-state-management.md
+++ b/docs/adr/016-frontend-state-management.md
@@ -2,7 +2,7 @@
 
 Date: 2025-04-17
 
-Status: Accepted. Extended by [ADR 30](030-frontend-store-registry-semantics.md), which adds fetching and invalidation semantics.
+Status: Accepted. Amended by [ADR 30](030-frontend-store-registry-semantics.md), which adds fetching and invalidation semantics and scopes the editing-UI rule to reads.
 
 ## Context
 
--- a/docs/adr/030-frontend-store-registry-semantics.md
+++ b/docs/adr/030-frontend-store-registry-semantics.md
@@ -2,115 +2,100 @@
 
 Date: 2026-07-31
 
-Status: Accepted
+Status: Draft
 
-Amends [ADR 16](016-frontend-state-management.md), which settles which components read the
-stores but is silent on fetching, caching and invalidation: its editing-UI rule now applies
-to reads, while mutations and validation still go through store actions.
+Amends [ADR 16](016-frontend-state-management.md), which settles which components read the stores but is silent on
+fetching, caching and invalidation. Its editing-UI rule now applies to reads; mutations and validation still go through
+store actions.
 
 ## Context
 
-The frontend keeps server state in Pinia stores (`SchemaStore`, `LayoutStore`, `SubjectStore`).
-Each store grew the same trio of actions: `fetchX`, `getOrFetchX` and `saveX`. The trio treated
-the store as a cache: reads consulted stored values to avoid fetches, and saves wrote the
-client's object through. An invalidation model was never decided
-([#1211](https://github.com/ProfessionalWiki/NeoWiki/issues/1211)). Reads assigned awaited
-results to shared state unconditionally
-([#1177](https://github.com/ProfessionalWiki/NeoWiki/issues/1177)), writes invalidated only on
-their own success path, and deletes invalidated nothing. The components already routed around
-the cache: editing flows fetched fresh data before opening, and two flows reloaded the whole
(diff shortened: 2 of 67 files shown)


JeroenDeDauw:
LGTM

PR #1223, opened 2026-07-31
Run Renovate for dependency updates

JeroenDeDauw:
Supersedes https://github.com/ProfessionalWiki/NeoWiki/pull/1221

Adds a nightly Renovate run and its configuration, modelled on the one DockerWiki uses. Nothing proposes dependency
updates here today: Dependabot version updates are off, and its security updates only fire once an advisory lands.

The configuration keeps the pull request volume low. npm and composer minor and patch updates each roll into one
grouped pull request, GitHub Actions get their own, and majors go to the dependency dashboard instead of straight to a
pull request. Image tags in the dev stack are treated that way from minor upwards, because they choose the PHP,
database and Node versions development and CI run on.

`:maintainLockFilesWeekly` regenerates the npm lock files every Monday. That is what moves nested transitive packages,
the drift behind the brace-expansion and js-yaml alerts.

Needs `RENOVATE_TOKEN_PROWIKIEXPERTS` to be readable from this repository. The workflow is otherwise the one DockerWiki
runs, on the current action release.

> <sub>AI-authored — Claude Code, `Opus 5 (max)`; asked to replace the bespoke lockfile workflow in #1221 with Renovate after pointing at DockerWiki's setup; no revisions; diff not yet human-reviewed; `renovate-config-validator` passes on the config, but Renovate itself has not run against this repository.</sub>

### DIFF
--- a/.github/workflows/renovate.yml
+++ b/.github/workflows/renovate.yml
@@ -1,0 +1,30 @@
+name: Renovate
+
+on:
+  schedule:
+    - cron: '17 2 * * *'
+  workflow_dispatch:
+
+permissions:
+  contents: write
+  pull-requests: write
+  issues: write
+
+concurrency:
+  group: renovate
+  cancel-in-progress: true
+
+jobs:
+  renovate:
+    name: "Renovate"
+
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Run Renovate
+        uses: renovatebot/github-action@v46.2.0
+        with:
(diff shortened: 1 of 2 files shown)


JeroenDeDauw:
@alistair3149 you want to continue this?

alistair3149:
The PR looks done but we need `RENOVATE_TOKEN_PROWIKIEXPERTS` which I don't have

PR #1226, opened 2026-07-31
Project every page into the graph databases

### LINKED ISSUE #1190: Persist Page Properties for pages without Subjects
## Problem

NeoWiki projects a graph Page node (its Page Properties and Subjects) only for a page whose latest revision carries the NeoWiki subject slot. A page with no Subject gets no Page node, so its Page Properties (`name`, `namespaceId`, `creationTime`, `lastUpdated`, `categories`, `lastEditor`, and any an extension contributes via `PagePropertyProvider`, such as approval state) are not persisted and cannot be queried.

This blocks using NeoWiki as a general page-metadata store, for example querying approval state or system metadata across pages that are not themselves Subjects. It builds on the refresh-without-an-edit work in #889.

No ADR ratified the subject-slot gate. It is an emergent consequence of the slot-based projection ([ADR 004][adr004], [ADR 019][adr019]), and it conflicts with [ADR 023][adr023], which routes approval and page metadata to Page Properties on the page node on the assumption they are queryable. A page that never had a Subject has no slot; a page whose Subjects were all deleted keeps an empty slot and currently still gets a bare node. So the effective gate is "has the slot", which is incidental.

## Proposed change

Add an opt-in setting taking a list of namespace IDs, default empty. Project a page as a Page node if it has at least one Subject, or its namespace is listed. The subject slot stops being the projection gate.

The setting name used throughout this issue, `NeoWikiPageDataNamespaces`, is a working placeholder; see Open questions.

A namespace list rather than a boolean: it bounds the cost to the chosen namespaces (whole-wiki projection is the big-wiki performance concern) and lets an admin include exactly the namespaces that need it, for example the article namespace while keeping Talk pages out, which a single "all pages" switch cannot express. A boolean is just the special case of listing every content namespace.

### Behaviour

- At least one Subject: projected as today, in any namespace.
- No Subject, namespace listed: projected, with Page Properties and no Subjects.
- No Subject, namespace not listed: not projected; a slotless page issues no graph write on edit.

Default (empty list) means "only pages with Subjects".

## Change from current behaviour

Today a page that carries the subject slot but currently has zero Subjects (its Subjects were all deleted) keeps a bare Page node. Under this change that page is projected only if its namespace is listed; otherwise its node is removed. Pages that still have at least one Subject are unaffected and keep their node in any namespace. On upgrade the removals land incrementally (each affected page loses its node on its next edit), so a rebuild is recommended to reach a consistent state.

## Tasks

- [ ] `NeoWikiPageDataNamespaces` config plus a `PageProjectionPolicy` (`hasSubjects OR namespace listed`), read live from config.
- [ ] `OnRevisionCreatedHandler`: save-or-delete by the policy (covers edit, undelete, import).
- [ ] `PageMoveComplete` hook running the policy, so a page moved in or out of a listed namespace gains or loses its node.
- [ ] Refresh-without-edit path (#889): honour the policy; extend `PageRefreshOutcome` to distinguish saved / removed / skipped.
- [ ] `RebuildGraphDatabases`: broaden enumeration to "pages to project" (Subjects union listed namespaces) as a batched iterator (coordinate with #1117); change the unregistered-slot-role path added in #1115 so it still enumerates listed-namespace pages instead of returning nothing; extend the deleted-page lookup with a namespace predicate so deleted subjectless pages are cleaned.
- [ ] Graph layer: confirm a zero-Subject save retains the Page node; native RDF already emits page metadata; extend `DumpRdf` to the same enumeration.

## Design decisions

- Reconciliation stays MediaWiki-derived; the graph is never enumerated. Deleted subjectless pages are found via the archive table (extended with the namespace predicate), move-out via the move hook, and a namespace removed from the config is handled by a wipe-and-rebuild. This preserves the write-only-projection model and avoids shared-graph scoping and mid-rebuild race hazards.
- Freshness: subjectless Page nodes refresh on every edit, on rebuild, and via the #889 refresh operation. A Page Property that changes with no revision (e.g. approval toggled without an edit) is the existing #889 case, so this feature depends on the extension calling that refresh.
- Changing `NeoWikiPageDataNamespaces` in either direction requires a rebuild: widening backfills existing pages (which otherwise project only on their next edit), narrowing removes now-unqualified nodes.
- Cost: a projection is a full main-slot parse plus revision and category lookups plus a write per backend, synchronous on save. Every edit in a listed namespace pays this, and the rebuild scales with the size of the listed namespaces, hence the batched-iterator dependency on #1117. Deferred optimizations: skip unchanged writes, job-queue the projection, parallelize the rebuild.

## Open questions

- Setting name is not decided; `NeoWikiPageDataNamespaces` is a placeholder used above. Suggestions for a clearer name welcome.
- Enumeration overlaps #1117: implement the generalized batched iterator here, or land #1117 first and build on it.
- Redirects in a listed namespace project too, including the redirect left by a move; carve out `page_is_redirect`?
- Final `PageRefreshOutcome` shape.

[adr004]: https://github.com/ProfessionalWiki/NeoWiki/blob/master/docs/adr/004-use-dedicated-slot.md
[adr019]: https://github.com/ProfessionalWiki/NeoWiki/blob/master/docs/adr/019-graph-database-architecture.md
[adr023]: https://github.com/ProfessionalWiki/NeoWiki/blob/master/docs/adr/023-subject-sources.md

> <sub>AI-authored: Claude Code, `Opus 4.8 (max)`; developed with @malberts over many rounds of design steering (the namespace-list approach, MediaWiki-derived reconciliation, scope, and wording were his calls); reviewed by @malberts before posting; technical claims checked against the NeoWiki codebase on master and against the referenced issues and ADRs.</sub>

JeroenDeDauw:
Fixes https://github.com/ProfessionalWiki/NeoWiki/issues/1190

Every page now has a Page node in every configured graph database; until now projection was gated on the
page carrying the subject slot, so the Page Properties of a page without Subjects could not be queried.
Every write path projects the page (edit, move, import, undelete), the RDF surfaces expose the same set, and
`RebuildGraphDatabases` and `DumpRdf` walk the page table with a batched keyset iterator, as the
deleted-page cleanup now walks the archive.

Deviation from the issue, which proposes an opt-in namespace-list setting: this projects everything instead.
NeoWiki is in early development, and the deployments in view — BlueSpice Galaxy subwikis, ECHOLOT case-study
instances, demo wikis, all in the thousands of pages — either want every page projected or are too small to
care, so a setting every user would set the same way is an installation chore rather than configurability.
What this projects that a namespace list would typically exclude: talk, user and other non-content pages,
and redirects. A gate can be reintroduced if a deployment with a concrete cost constraint appears, but it
would reopen the every-page-is-projected contract this PR establishes across the graph and RDF surfaces — a
deliberate revert, not a toggle. The issue's reconciliation rule survives: removals derive from MediaWiki
state (the archive), never from enumerating the graph.

Notable points:

- **Unreadable subject slots are left alone.** A page whose subject slot holds content that cannot be read as
  Subjects — reachable when the content model that wrote it is no longer installed — is skipped and logged
  rather than projected: writing it as holding no Subjects would drop the Subjects it holds from the graph.
  The RDF surfaces 404 for such a page for the same reason, rather than describing it as a page without
  Subjects.
- **Undeletion projects the page's current revision, once.** It moved from the per-revision
  `RevisionUndeleted` hook to `PageUndeleteComplete`, and reads the current revision fresh rather than
  trusting the hook's last-restored one. Projecting each restored revision in turn wiped the Subjects of a
  live page when old history was restored onto it.
- No `PageMoveComplete` hook is needed. Core creates a null revision on the new title for every move and
  fires `RevisionFromEditComplete` for it, and a redirect left behind is a new page projected when it is
  created. Tests pin the write count, so a move keeps projecting each page exactly once and deleting nothing.
- RDF autodiscovery links, the page-tools RDF link and the endpoint now follow one contract: every page that
  exists has an export, and a title with no page behind it advertises nothing. 404 from the endpoint means
  the page does not exist or the caller may not read it, which stay indistinguishable.
- Page properties that cannot be built (an uninstalled content model, a throwing provider) skip the
  projection and log, rather than aborting the user's edit — matching how a failing backend is handled.
- The deleted-page lookup pages over `ar_id`, the archive primary key: MediaWiki's archive table has no index
  on `ar_page_id`, so paging over the page id would sort the whole table for every batch.
- The anonymous `neowiki-query` grant now reaches the whole page inventory rather than subject pages only.
  On an open wiki that discloses nothing MediaWiki core does not already expose; on a wiki with read
  restrictions it widens the pre-existing hole — the metadata of every page, no longer just subject pages —
  until query-time access control ([#1059](https://github.com/ProfessionalWiki/NeoWiki/issues/1059)) closes
  it.
- An existing wiki needs one `RebuildGraphDatabases` run to backfill the pages that were never projected;
  without it, each is projected on its next edit.

Cost: an edit to a page without Subjects now writes a page node where it previously wrote nothing — a parse
to build the page properties plus one upsert per backend. For magnitude: a subjectless projection does
strictly less graph work than any subject-bearing one, and #1225's benchmarks put a ten-Subject page's full
Neo4j write at ~33 ms on dev hardware. Deferred, as in the issue: skipping unchanged writes, job-queueing
the projection, and parallelizing the rebuild.

> <sub>AI-authored — Claude Code: implementation and draft `Opus 5 (max)`, design calls and final edit `Fable 5
(max)`; detailed spec from @JeroenDeDauw (issue #1190, then a session brief replacing its opt-in setting with
always-on projection), reworked twice: once after that product call, once after a review that found three
blockers; diff not yet human-reviewed; full PHPUnit and phpcs/phpstan green locally, each blocker's regression
test verified failing against the previous revision, and the edit, delete, move, undelete, refresh-without-edit,
rebuild, dump, RDF and SPARQL paths exercised end to end on a dev wiki, where the page table, Neo4j and the RDF
dump each hold the same 120 pages.</sub>

The always-on design and the rationale for deviating from the issue were the orchestrating session's calls,
not the implementing subagent's.

</details>


docs/api/graph-model.md
```diff
@@ -22,8 +22,10 @@ Two node types and two relationship categories:
 ## Page Nodes
 
 Every page is projected as a `:Page` node when it is saved and by
-[`RebuildGraphDatabases.php`](../operations/maintenance.md#rebuilding-the-graph), whether or not it holds Subjects, so
-after a rebuild every page on the wiki has one.
+[`RebuildGraphDatabases.php`](../operations/maintenance.md#rebuilding-the-graph), whether or not it holds Subjects. The
+exceptions are pages NeoWiki cannot read: one whose subject slot holds content it cannot read as Subjects, and one whose
+Page Properties cannot be built. Both are reported by the rebuild and left out of the graph rather than projected as
+holding nothing, which would drop data the page does hold.
```
JeroenDeDauw, line 28: @alistair3149 careful with these follow up PRs. I am ensuring no clutter docs are added and then via your review commits a bunch of stuff is added and merged.

This particulat addition is pure doc rot. It should not be here. It's documenting a try-catch block which itself is arguably overly defensive.

JeroenDeDauw: Btw I did not read the other changes below; would just revert or manually verify they actually make sense. And note this is just one of the additional commits you added on top of the PR before merging, others might be adding more doc rot.

alistair3149: Following up on that in a separate PR

JeroenDeDauw: Did you?

alistair3149: https://github.com/ProfessionalWiki/NeoWiki/pull/1249

PR #1227, opened 2026-08-01
Surface Constraint severity in the frontend: parse the object form and style warnings

### LINKED ISSUE #1148: Surface Constraint severity in the frontend: parse the object form and style warnings
[ADR 26](https://github.com/ProfessionalWiki/NeoWiki/blob/master/docs/adr/026-validation-severity-levels.md) gives every Constraint a severity (`error` / `warning`, default `warning`). The backend half landed in #1147. The frontend has no severity handling at all yet, which leaves two gaps.

## 1. Parse the dual Constraint form

A Constraint is now written either as a bare scalar or as an object carrying the severity — see [Constraint severity](https://github.com/ProfessionalWiki/NeoWiki/blob/master/docs/api/schema-format.md#constraint-severity). The TypeScript schema parsers read every Constraint as a scalar, so a Schema saved with the object form is mis-read today:

- `resources/ext.neowiki/src/domain/PropertyDefinition.ts` — `required: json.required ?? false` yields a truthy object.
- `resources/ext.neowiki/src/domain/propertyTypes/{Number,Text,Date,DateTime,Url,Select}.ts` — `json.maximum`, `json.minLength`, `json.uniqueItems`, `json.options` land as objects, so numeric comparisons become `NaN`.
- `resources/ext.neowiki/src/persistence/SchemaSerializer.ts` must round-trip the object form back out.

A shared `parseConstraint` helper mirroring the PHP `SeverityNormalizer` is the natural shape; the discriminator is the same (an object carrying a `severity` key). Severity applies to Constraints only — a severity on a Display Attribute such as `precision` is discarded server-side.

## 2. Carry and style violation severity

Every serialized violation now carries an always-present `severity` — see [Validation codes](https://github.com/ProfessionalWiki/NeoWiki/blob/master/docs/api/validation-codes.md#severity-blocking-and-enforcement). The frontend currently drops it:

- `resources/ext.neowiki/src/domain/SubjectViolation.ts` — add `severity`.
- `resources/ext.neowiki/src/persistence/violationParsing.ts` — carry it through (default `warning` for any legacy body).
- `useServerViolations`, `useFieldServerViolation` and `useStringValueInput` build Codex `{ error: … }` messages unconditionally; warnings should render with the `warning` status instead.
- The value inputs hardcode `:status="… ? 'error' : 'default'"` and need a three-way status.
- The aggregate banners in `SubjectEditorDialog.vue` and `SubjectCreatorDialog.vue` hardcode `CdxMessage type="error"`.
- A warning-only result must not block saving: warnings never block server-side, so the client should not treat them as fatal.

## Why it matters

Until this lands, object-form Schemas mis-render in the schema editor and display, and a user cannot tell an advisory warning from a blocking error.

alistair3149:
Fixes #1148. Frontend half of per-Constraint validation severity ([ADR 26](https://github.com/ProfessionalWiki/NeoWiki/blob/master/docs/adr/026-validation-severity-levels.md)); the backend landed in #1147.

## Parsing and round-trip

* New `domain/Severity.ts` and `domain/SeverityNormalizer.ts` mirror the PHP `SeverityNormalizer`: `extractSeverities` unwraps the object Constraint form (`{ value, severity }`; value-less booleans imply `true`; explicit `value: null` preserved; reserved `type`/`description`/`default` keys skipped; invalid severities throw), `applySeverities` re-wraps only `error`-annotated keys so unannotated Schemas round-trip unchanged.
* `PropertyDefinitionDeserializer` extracts once, drops severities on registered types' Display Attributes (matching `PropertyDefinition::fromJson`), and hands the normalized values to the type parsers — so `required` no longer parses to a truthy object and numeric bounds no longer become `NaN`. The unregistered-type fallback spreads the normalized values, keeping annotated custom keys round-trippable.
* `PropertyDefinition` gains an optional `constraintSeverities` map (set only when a Constraint is annotated), which `SchemaSerializer` re-applies on output. The schema editor preserves it through attribute edits and drops it on type change, like the other Constraint fields.

## Violation severity in the UI

* `SubjectViolation` carries the always-present `severity`; `parseViolations` defaults legacy bodies to `warning` and rejects unknown values.
* `useServerViolations` returns severity-keyed Codex message objects (`firstMessages`/`fieldLevelMessages`) plus a shared `violationStatus()` helper; `useFieldServerViolation` returns `validationMessages`/`validationStatus` (rename allowed under the public-api alpha contract); `useStringValueInput` keys per-part and summary messages by severity.
* All value inputs bind a three-way field status, so warnings render with Codex `warning` styling instead of `error`. Multi-value Text/Url fields keep their pre-existing `default` field status (per-slot statuses live in `NeoMultiTextInput`).
* The duplicated dialog banner is extracted into `SubjectViolationBanners.vue`, which splits anchorless violations into an error banner and a warning banner.
* Saving was never gated client-side; new dialog tests pin that a warning-only dry-run result does not block the save in either dialog.

## Manual Browser Check

1. On a dev wiki with demo data, open a Schema page (e.g. `Schema:Person`) with `action=edit` and give a number property an annotated bound, e.g. `"maximum": { "value": 100, "severity": "error" }`, plus a plain Constraint such as `"minLength": 2` on a text property. Save.
2. Open the schema in the Schema editor (via a Subject's edit dialog → schema link), confirm the bound shows as `100` (not blank/NaN), re-save, and check the stored JSON kept the `{ "value": 100, "severity": "error" }` object form while the plain Constraints stayed scalars.
3. Edit a Subject of that schema: enter a number above the maximum and blur — the field should show red error styling with the max-value message. Violate the `minLength` Constraint — the field should show yellow warning styling instead.
4. Rename the schema page (or edit a Subject whose schema is missing) to trigger `schema-not-found`: the dialog banner should render as a warning message, and saving the Subject should still succeed.

<sub>AI-authored — Claude Code, Fable 5; implementation of #1148 mirroring the #1147 backend semantics; verified via the tsci checks run individually (vitest 1245/1245 with a capped worker pool, vue-tsc + vite build, eslint/stylelint — a single make tsci invocation OOMs in this dev sidecar) plus an adversarial review subagent whose three minor findings are addressed.</sub>

### DIFF
--- a/resources/ext.neowiki/src/components/SubjectCreator/SubjectCreatorDialog.vue
+++ b/resources/ext.neowiki/src/components/SubjectCreator/SubjectCreatorDialog.vue
@@ -91,19 +91,7 @@
 				</template>
 			</CdxField>
 
-			<CdxMessage
-				v-if="anchorlessViolations.length > 0"
-				type="error"
-			>
-				<ul class="ext-neowiki-subject-editor__form-errors">
-					<li
-						v-for="( v, idx ) in anchorlessViolations"
-						:key="idx"
-					>
-						{{ formatViolationMessage( v ) }}
-					</li>
-				</ul>
-			</CdxMessage>
+			<SubjectViolationBanners :violations="anchorlessViolations" />
 
 			<SubjectEditor
 				v-if="statements"
@@ -171,8 +159,8 @@ import { Schema } from '@/domain/Schema.ts';
 import { StatementList } from '@/domain/StatementList.ts';
 import { withoutMissingValueViolations, type SubjectViolation } from '@/domain/SubjectViolation';
 import { ValidationFailedError } from '@/persistence/ValidationFailedError';
-import { CdxMessage } from '@wikimedia/codex';
 import SubjectEditor from '@/components/SubjectEditor/SubjectEditor.vue';
+import SubjectViolationBanners from '@/components/common/SubjectViolationBanners.vue';
 import SchemaCreator from '@/components/SchemaCreator/SchemaCreator.vue';
 import type { SchemaCreatorExposes } from '@/components/SchemaCreator/SchemaCreator.vue';
 import SummaryAction from '@/components/common/SummaryAction.vue';
@@ -317,10 +305,6 @@ const anchorlessViolations = computed<SubjectViolation[]>( () => {
 		return !renderedPropertyNames.has( v.propertyName );
 	} );
 } );
-
-function formatViolationMessage( v: SubjectViolation ): string {
-	return mw.message( `neowiki-field-${ v.code }`, ...( v.args as string[] ) ).text();
-}
 
 function handleClearViolation( payload: { propertyName: string; valuePartIndex: number | null } ): void {
 	serverViolations.value = serverViolations.value.filter(
--- a/resources/ext.neowiki/src/components/SubjectCreator/SubjectCreatorDialog.vue
+++ b/resources/ext.neowiki/src/components/SubjectCreator/SubjectCreatorDialog.vue
@@ -91,19 +91,7 @@
 				</template>
 			</CdxField>
 
-			<CdxMessage
-				v-if="anchorlessViolations.length > 0"
-				type="error"
-			>
-				<ul class="ext-neowiki-subject-editor__form-errors">
-					<li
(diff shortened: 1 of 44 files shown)


JeroenDeDauw:
Tested locally

PR #1235, opened 2026-08-01
Support Oxigraph as a second SPARQL store engine

JeroenDeDauw:
Adds Oxigraph as a second SPARQL store engine, next to QLever.

[ADR 19](https://github.com/ProfessionalWiki/NeoWiki/blob/master/docs/adr/019-graph-database-architecture.md) claims one SPARQL plugin works across stores. This PR validates that against a second engine with zero `src/` changes — for endpoint shape, auth model, and named-graph semantics. Error surfacing and result-literal serialization stay covered by fakes only, against neither live engine. The diff touches the Docker stacks, CI, docs, tests, `phpunit.xml.dist` and the Makefile.

A store entry is plain config. Oxigraph splits its endpoints, so `queryUrl` must be set explicitly, and it has no authentication (oxigraph/oxigraph#88): deployments must isolate it at the network layer, and the tools overlay and CI bind its host port to loopback only.

```php
$wgNeoWikiSparqlStores = [
	[
		'updateUrl' => 'http://oxigraph:7878/update',
		'queryUrl' => 'http://oxigraph:7878/query',
	],
];
```

What this adds:

* **Two opt-in services.** An `oxigraph` service with `--union-default-graph` (unscoped queries behave as on QLever) sits in `docker-compose.yml` next to `qlever`, behind an `oxigraph` compose profile: nothing points `$wgNeoWikiSparqlStores` at it until you repoint that setting, so neither stack starts it and `COMPOSE_PROFILES=oxigraph make dev` is what does. It lives in that file rather than the dev overlay because `docker-compose.tools.yml`, which maps its host port, works on either stack and cannot map a port for a service the base file lacks. A spec-strict `test_oxigraph` sits in the dev overlay behind the `test` profile next to `test_neo` and `test_qlever`, so violations of strict SPARQL 1.1 semantics surface in tests; `make test-backends` waits for its port the way it waits for `test_qlever`'s. The image is distroless, so neither carries a compose healthcheck. Images track `:latest` like the QLever services; behavior was verified against 0.5.9.
* **CI runs the SPARQL end-to-end tests against both engines.** An engine's test class fails rather than skips when its store is missing, so a missing store cannot silently leave the loop unverified, and a dedicated test pins each store's default-graph semantics (strict store: unscoped queries find nothing; union store: they find page data).
* **The end-to-end label query silently relied on QLever folding named graphs into the default graph**, which the spec does not guarantee, so it is now `GRAPH`-scoped and runs on either engine. The doc and demo examples are deliberately left unscoped: every configuration NeoWiki ships unions named graphs, and a blanket `GRAPH` block would constrain its patterns to a single page and so break cross-page joins — the idiom `docs/rdf/person-to-edm.md` already teaches correctly with one block per graph. [RDF Export](https://github.com/ProfessionalWiki/NeoWiki/blob/master/docs/rdf/rdf-export.md#iri-scheme) gains the named-graph query contract instead, and the Lua and parser-function pages link to it.
* Installation docs gain an Oxigraph section, and its no-authentication note matters because NeoWiki's own `accessToken` setting does nothing on this store. The federation section previously closed with "Other SPARQL stores have equivalent settings" — `oxigraph serve` has five options and none controls `SERVICE`, so that claim was false for any non-QLever store, in a security paragraph. It now states Oxigraph's gap, keeps the pointer for other stores, and names the two controls that remain.

In a live probe of 0.5.9, the write path's multi-operation update request worked as-is, and oxigraph/oxigraph#892's update-size cap did not reproduce (a 1M-triple `INSERT DATA` was accepted).

Left out deliberately: Basic auth for reverse-proxied stores (no known deployment needs it), store-choice guidance (a write-path comparison is in progress), and multi-store query addressing (the read surfaces target the first configured store, as documented).

Behavioural review findings against this branch are in a follow-up PR; the documentation findings are folded in here, since they change how the feature is documented rather than what it does.

Related: #586 (QLever support), #1053 (per-projection named graphs).

> <sub>AI-authored — Claude Code, `Fable 5 (max)` for the change and `Opus 5 (xhigh)` for the rebase onto master and a documentation revision; one-line ask from @JeroenDeDauw executing a prior decision, steered once mid-run, rebase and doc pass asked for by @alistair3149, who twice redirected the scope; diff not yet human-reviewed; both live two-engine integration suites re-run green after every rewrite, phpcs + phpstan clean, 16 compose configurations validated, teardown checked against a live stack, `oxigraph serve --help` checked before the federation claim.</sub>

Original direction, implementation and description by `Fable 5 (max)` with `Opus 5 (max)` subagents; engine facts verified against a live Oxigraph 0.5.9 container.

Rebased onto master by `Opus 5 (xhigh)` after master moved the `qlever` service from the dev overlay into `docker-compose.yml` and reworked `docker-compose.tools.yml` to serve either stack. Four files conflicted. One resolution was a judgment call rather than a merge: the `oxigraph` service moved into `docker-compose.yml` too, because a ports-only entry in the tools overlay makes `docker compose -f docker-compose.yml -f docker-compose.tools.yml` fail outright with `invalid compose project` — verified before choosing, and the profile-gated `caddy` in that same file is the existing precedent.

The documentation was then revised after two blind `Opus 5` reviews, given the artifacts and `docs/AGENTS.md` but not the authoring session. That pass reverted this branch's original propagation of `GRAPH ?g { … }` into the doc and demo examples, on the finding that it narrows the queries rather than making them portable, and moved the named-graph query contract to `docs/rdf/rdf-export.md`, which already owns the graph IRI scheme.

</details>

### DIFF
--- a/.github/workflows/ci-php.yml
+++ b/.github/workflows/ci-php.yml
@@ -53,6 +53,9 @@ jobs:
       # reaches the QLever the "Start QLever" step publishes on the runner's localhost.
       QLEVER_TEST_URL: http://localhost:7019/
       QLEVER_TEST_ACCESS_TOKEN: neowiki_test_token
+      # Live Oxigraph store for the same system test run against a second engine, published by the
+      # "Start Oxigraph" step.
+      OXIGRAPH_TEST_BASE_URL: http://localhost:7878
 
     services:
       test_neo:
@@ -143,6 +146,19 @@ jobs:
               qlever-index -i neowiki -F nt -f data.nt
               exec qlever-server -i neowiki -p 7019 -a "$QLEVER_ACCESS_TOKEN" -m 1G --service-allowed-iri-prefixes -'
 
+      # Oxigraph, the second SPARQL engine the same system test runs against. Started explicitly next
+      # to QLever rather than as a `services:` container, so the serve flags are stated here rather than
+      # inherited from the image's default. They mirror the test_oxigraph
+      # dev-stack service (see Docker/docker-compose.dev.yml) — no --union-default-graph, so the store
+      # is spec-strict; no --location, so it is in memory, which is all a fresh-per-run CI store needs.
+      # No access-token flag: Oxigraph has no authentication (oxigraph#88). No federation flag either:
+      # Oxigraph federates unconditionally and offers no option to restrict it — see
+      # docs/operations/installation.md#restricting-federation.
+      - name: Start Oxigraph
+        run: |
+          docker run -d --name test_oxigraph -p 127.0.0.1:7878:7878 \
+            ghcr.io/oxigraph/oxigraph:latest serve --bind 0.0.0.0:7878
+
       - name: Run update.php
         run: php maintenance/update.php --quick
 
@@ -159,6 +175,16 @@ jobs:
             sleep 2
           done
           echo "QLever did not become ready in time; dumping logs:"; docker logs test_qlever; exit 1
+
+      - name: Wait for Oxigraph
+        run: |
+          for i in $(seq 1 40); do
+            if curl -fsS -o /dev/null "http://localhost:7878/query?query=SELECT%20%2A%20WHERE%20%7B%20%3Fs%20%3Fp%20%3Fo%20%7D%20LIMIT%201"; then
+              echo "Oxigraph is serving"; exit 0
+            fi
+            sleep 2
+          done
+          echo "Oxigraph did not become ready in time; dumping logs:"; docker logs test_oxigraph; exit 1
 
       - name: Run PHPUnit
         run: composer phpunit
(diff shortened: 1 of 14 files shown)


JeroenDeDauw:
Heads-up on a merge-order interaction: #1238 moves the `qlever` service from the dev overlay into the base compose file, so the demo stack ships it. Simulating both merge orders: `Docker/README.md` and `docs/operations/installation.md` conflict visibly (resolve toward #1238's demo-stack framing and its renamed Docker/README anchor), but two things merge *cleanly* and would then be wrong if this lands second: the oxigraph block comment ("alongside the QLever pair above… dev-only like them" — qlever is no longer in that file) and the `test_oxigraph` comment's "the demo `oxigraph` service above" phrasing. Whichever PR merges second should sweep the other's terminology; if #1238 goes second I'll do the equivalent sweep there.

> <sub>AI-authored — Claude Code, `Fable 5 (max)`; coordination note from the session shepherding #1238.</sub>

PR #1248, opened 2026-08-04
Record sub-property bridging as a mapping-layer mechanism

JeroenDeDauw:
Adds a sub-property option to Limitations To The Upsides, written for the consortium readers weighing the local-vs-global question: a mapping can declare native predicates `rdfs:subPropertyOf` ontology terms, making the ontology term the shared anchor a global property registry would otherwise provide. States the interaction with the as-built projection honestly — predicates are flat per-name (`neo-prop:Name`), so one declaration spans all same-named properties, and per-Schema anchoring needs Schema-scoped predicates (the open predicate-scope question in NativeRdfProjection.md). Keeps the entailment caveat as a runnable property path and points ontology-shape consumers at the ontology projection.

Considered, omitted: where wiki-level declaration triples live in the per-page named-graph sync; the property path also matching co-hosted ontology projections (#1053 setups); owl:equivalentProperty for the genuinely-identical case; entailment availability on stores beyond the shipped pair (ADR 19 names more).

> <sub>AI-authored — Claude Code, `Fable 5 (max)`; Claude-proposed addition approved by @JeroenDeDauw, then rewritten for ECHOLOT-stakeholder readability on his direction (two rounds); not yet human-reviewed; predicate-shape claims verified against docs/rdf/rdf-export.md and RdfNamespaces.php, text blind-judged and blind-reviewed.</sub>

### DIFF
--- a/docs/planning/GlobalProperties.md
+++ b/docs/planning/GlobalProperties.md
@@ -70,6 +70,13 @@ terms during import/export. ECHOLOT T2.3 talks about "model-model mappings" betw
 system to adopt the same model. A mapping configuration like "Person.Name maps to `foaf:name`" and "City.Name
 maps to `dcterms:title`" achieves interoperability without changing the data model.
 
+**Sub-property declarations can keep local distinctions while answering ontology-term queries.** Instead of only
+renaming local properties to ontology terms, the projection can emit them as distinct terms declared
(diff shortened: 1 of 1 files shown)


JeroenDeDauw:
Sub-property Idea originally coming from @vrandezo, though not exactly as written here

PR #1250, opened 2026-08-05
Address the review findings on the scoped graph rebuild

alistair3149:
Addresses the findings from a review of #1233, one commit per finding. Targets `graph-rebuild-scoped` so it can be merged into that PR before it goes to master.

The review found 15 issues; one was refuted on validation and one is deferred. What is fixed here:

**Behaviour**

- **A run of page-specific failures is no longer an absorbing stop.** `storeLooksGone()` inferred store death from counts alone and rewound the cursor, so a contiguous window of pages a healthy store refuses stopped every attempt at the same place — `--resume` and a fresh run both walked back into it, and nothing behind it was ever projected. The store is now asked whether it is still there, via the `initialize()` the plugin contract already offers for exactly that. A store that answers means the pages are at fault, so they are counted, reported and walked past. This also fixes the failed page ids never reaching the log, since a rewound batch reports nothing.
- **A resumed run records who is driving it.** `resume()` copied the trigger, so `refuseWhenStartedByHand()` read an operator's resumed run as unattended and a Mapping edit cancelled it mid-walk — via the recovery route the docs endorse (cancel on the page, then `--resume`).
- **A stale queued job no longer drives a run the shell resumed.** Cancelling deliberately does not reach into the queue because a job for an ended run does nothing — which held until `--resume` reopened a run id. Only the script resumes, and it never queues, so a queued batch that finds its run being driven from a shell stops.
- **Each store sharing a projection starts in its own transaction round.** Started in one deferred update, every store after the first threw `Cannot flush pre-lock snapshot` and was left on the old vocabulary.
- **Protecting a Mapping page no longer restarts its rebuild.** Protect, unprotect and expiry changes insert a content-identical null revision; a revision now has to say something new to count as a definition change.
- **Restoring a Mapping page counts as a change.** It queued no rebuild and flipped the store straight back to In sync, so a store rebuilt while the page was gone reported as reconciled.
- **An extension's store name is held to the rules a configured one meets.** `addPlugin()` applied neither the 255-byte limit nor the case-insensitive `neo4j` reservation, so a plugin could take a name whose run records it could never find again (truncated silently on production MySQL), or stand as `Neo4j` beside the bundled backend.
- **The SPARQL query route registers only when a store will back it.** A config whose usable entries are all dropped for their names registered the route with no plugin behind it, answering a query with a 500 instead of a 404.

**Tests**

- The request-timeout half of the executor's `catch ( TimeoutException | DBError )` was untested in both phases; dropping the type left the suite green.
- Nothing linked `GraphStoreName::MAX_LENGTH` to the `nwrr_store` column width. The new check reads the width off the abstract schema, because only MySQL materialises it and CI installs SQLite.

**Docs**

- `maintenance.md` over-generalised a whole batch failing, and its offline-recovery SQL matched nothing for a run stranded `queued` — the state the trigger it names actually leaves.
- `extending.md` named two of the three paths that call `initialize()`, and its throw-timing sentence was false for every background batch after the first.

Two further doc updates follow behaviour changed here, and are folded into the commits that caused them rather than standing alone: staleness and auto-rebuild now name restoring a Mapping page alongside editing and deleting it, and the plugin-facing name rules in `extending.md` now match the ones `installation.md` states for configured stores.

## Worth a look

1. **The store probe is the one design decision here.** Classifying the failures or bounding the rewind were the alternatives; probing needs no schema change, no new state, and no guesses about what a backend throws.
2. **One commit reverses a deliberate, tested decision.** `MappingPageChangeTimeLookupTest` pinned that a restored Mapping page reads as changed when it was last edited. That holds only for a store rebuilt before the deletion; one rebuilt during it is as far from the wiki as a store rebuilt before an edit.
3. **#2 is fenced by trigger rather than by an epoch column.** Only the CLI reopens a run id and only the CLI never queues, so the trigger already distinguishes the attempts. An epoch column would be the general fix if background resumption is ever added.

## Not addressed

- **`Special:GraphStores` emits `cdx-button` but loads no Codex styles**, so the buttons render as plain submit inputs. Left out at @alistair3149's request for a separate pass.
- **A Mapping page rename** was reported and refuted: `NS_NEOWIKI_MAPPING` is declared `"movable": false`, so no core move path can rename one.
- **#1233's own description is stale** — it lists the #1236 work as deferred and names three "known gaps" that no longer match the code.

## Verification

`make cs` clean, CI green across MW 1.43–master on PHP 8.3–8.5. Every behavioural fix has a test that was confirmed to fail without it; the two test-only commits were confirmed against the mutation they exist to catch. Local full-suite runs hang on this box for environmental reasons (the unmodified branch hangs identically), so per-file runs were used locally and CI covers the whole suite.

<sub>AI-authored — Claude Code, `Opus 5 (1M context)`; review, validation and implementation in-session; each fix verified against its negative case; not yet human-reviewed.</sub>

### DIFF
--- a/docs/extending/extending.md
+++ b/docs/extending/extending.md
@@ -163,15 +163,16 @@ refused backend receives no page changes.
 `PagePropertyProvider`, and runs for every revision, so subject edits, undeletions and page moves all reach you as a
 save. `deletePage` gets only the page id.
 
-`initialize` runs on `update.php`, and at the start of a `RebuildGraphDatabases` run of your store before any page is
-projected — create the store-level structures a fresh store needs there. Make it idempotent, since both paths call it
-every time; it never runs on an individual edit.
+`initialize` runs on `update.php`, at the start of a `RebuildGraphDatabases` run of your store before any page is
+projected, and once per batch of a rebuild started from the wiki — create the store-level structures a fresh store
+needs there. Make it idempotent and cheap, since every path calls it every time; it never runs on an individual edit.
+A rebuild also calls it to ask whether your store is still there when a whole batch of pages has failed.
 
 **Signal failure by throwing.** On an edit, delete or undelete, NeoWiki logs the failure and lets the user's
 operation commit, so a backend being down never blocks the wiki or starves the other backends — your projection is
-simply out of sync until someone runs `RebuildGraphDatabases`. During a rebuild, failures reach the rebuild instead: a
-page you refuse is logged and counted and the rebuild carries on, while an `initialize` throw ends the run before a
-page is read. On `update.php` a failing `initialize` is reported and the update carries on, though the backends
+simply out of sync until the store is rebuilt. During a rebuild, failures reach the rebuild instead: a page you refuse
+is logged and counted and the rebuild carries on, while an `initialize` throw ends the run — before a page is read
+when it opens the store for the run, and at whichever batch it happens on otherwise. On `update.php` a failing `initialize` is reported and the update carries on, though the backends
 registered after yours do not initialize on that run.
 
 Make `deletePage` idempotent: the rebuild re-issues a delete for every page MediaWiki no longer has, so it will ask
--- a/docs/extending/extending.md
+++ b/docs/extending/extending.md
@@ -155,9 +155,11 @@ class MyGraphDatabasePlugin implements GraphDatabasePlugin {
 Register with `NeoWikiRegistrar::addGraphDatabasePlugin( $name, $plugin )`. Example:
 [`src/RedHerbGraphDatabasePlugin.php`](https://github.com/ProfessionalWiki/NeoWiki/blob/master/tests/RedHerb/src/RedHerbGraphDatabasePlugin.php).
 
-The name is what [`--store`](../operations/maintenance.md#rebuilding-one-store) addresses. Pick a stable one and
-namespace it to your extension: a name already taken is refused with a warning on the `NeoWiki` channel, and a
-refused backend receives no page changes.
+The name is what [`--store`](../operations/maintenance.md#rebuilding-one-store) addresses, and what a rebuild files
+its run records under. Pick a stable one and namespace it to your extension. A name is refused with a warning on the
+`NeoWiki` channel when another backend already holds it, when it is `neo4j` in any casing — reserved for the bundled
+Neo4j backend — or when it is longer than 255 bytes, which is all a run record can hold. A refused backend receives
+no page changes and cannot be rebuilt.
 
 `savePage` hands you the page with all of its Subjects and the Page Properties contributed by every
 `PagePropertyProvider`, and runs for every revision, so subject edits, undeletions and page moves all reach you as a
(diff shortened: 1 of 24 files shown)


alistair3149:
Landed in `graph-rebuild-scoped` by fast-forward rather than through this PR, so GitHub closed it as merged. All 13 commits are in #1233 with their history intact.

Folded in first so the rework onto master's new page enumeration (#1226 renamed `SubjectPageIdsLookup`/`SubjectPageRebuilder` and changed the scope to every page) happens once over the complete content, instead of resolving the same `GraphRebuildExecutor` conflicts twice. Review of these fixes continues on #1233.

alistair3149:
Correction to my earlier comment: these commits are no longer in `graph-rebuild-scoped`. #1233 has been rewound to its pre-fold state so it stands as an unmodified record of the original design and its review.

The 13 commits live on in #1254, rebased onto master along with the rest of #1233. GitHub still shows this PR as merged because it was, briefly — the diff and discussion here remain the record of the review fixes themselves.

docs/extending/extending.md
```diff
@@ -155,23 +155,26 @@ class MyGraphDatabasePlugin implements GraphDatabasePlugin {
...
 [`src/RedHerbGraphDatabasePlugin.php`](https://github.com/ProfessionalWiki/NeoWiki/blob/master/tests/RedHerb/src/RedHerbGraphDatabasePlugin.php).
 
-The name is what [`--store`](../operations/maintenance.md#rebuilding-one-store) addresses. Pick a stable one and
-namespace it to your extension: a name already taken is refused with a warning on the `NeoWiki` channel, and a
-refused backend receives no page changes.
+The name is what [`--store`](../operations/maintenance.md#rebuilding-one-store) addresses, and what a rebuild files
+its run records under. Pick a stable one and namespace it to your extension. A name is refused with a warning on the
+`NeoWiki` channel when another backend already holds it, when it is `neo4j` in any casing — reserved for the bundled
+Neo4j backend — or when it is longer than 255 bytes, which is all a run record can hold. A refused backend receives
+no page changes and cannot be rebuilt.
```
JeroenDeDauw, line 162: So this is talking about what the code logs? Did not fully verify but looks like there already was clutter before, and now there is even more.

> The name is what [`--store`](../operations/maintenance.md#rebuilding-one-store) addresses. Pick a stable one and namespace it to your extension.

Seems to be all that is needed. Still horribly worded

JeroenDeDauw: Again I did not review the other additions and expect more issues there.

I'm really tired of reviewing this AI text as I deal with it the whole day already and don't want to be constantly reviewing it on other people's PRs. This kind of rot does real damage to the product and should not get auto-merged like seems to be happening consistently here.

alistair3149: Yes it is the failure mode described in the log.
I re-checked the failure path and found that the failure reason wasn't visible to the operator in the default setup.
The doc trim and the logging setup is followed up in #1258

PR #1251, opened 2026-08-05
Resolve referenced Subjects in one lookup on the read path

### LINKED ISSUE #1243: Read path resolves referenced Subjects one at a time, though the batch lookups now exist
#1224 batched relation-target resolution for the write path, leaving the read path resolving one
Subject at a time. It listed that under "Considered and omitted", on the grounds that "Nothing there
resolves a set today". That premise does not hold: both read queries already hold a `SubjectIdList`
and loop it.

`GetSubjectQuery`, when `includeReferencedSubjects` is set:

```php
foreach ( $subject->getReferencedSubjects()->asArray() as $id ) {   // src/Application/Queries/GetSubject/GetSubjectQuery.php:50
    $referencedSubject = $this->subjectLookup->getSubject( $id );   // :51
    ...
    $referencedPageIdentifiers = $this->pageIdentifiersLookup->getPageIdOfSubject( $referencedSubject->id );  // :57
```

`GetPageSubjectsQuery` does the same at :88-:99, and additionally resolves the hosting page one
subject at a time for the page's own subjects at :47 and :54.

`Subject::getReferencedSubjects()` returns a `SubjectIdList`, which is exactly what
`SubjectLookup::getSubjects()` and `PageIdentifiersLookup::getPageIdsOfSubjects()` take, so the batch
methods #1224 added fit these call sites without new API surface.

## Cost

Every `getSubject` is a graph round trip through the subject→page index plus a revision load and a
content deserialize, and every `getPageIdOfSubject` is another round trip, since
`Neo4jPageIdentifiersLookup` answers the single-id method by running the batch query. A Subject with
M referenced targets therefore costs about 2M serial round trips on a read, and a page with N
subjects each referencing M costs N + 2NM. The write path this was measured on was 3.

This is a read path, so it runs more often than the validation path #1224 optimised, and
`MediaWikiSubjectRepository::getSubjects` already groups ids by hosting page, so batching would also
collapse the repeated page reads when several targets share a page.

I have not profiled these endpoints; the round-trip counts above are read off the loops, not
measured.

## Suggested shape

Resolve the referenced ids in one `getSubjects()` call before the loop and read them out of the
returned `SubjectMap`, then resolve their hosting pages in one `getPageIdsOfSubjects()` call, the
same shape `SubjectValidator` now uses. The per-title read authorization in the loop
(`pageIsReadableOrUnresolved`) stays per subject and is unaffected.

Worth checking whether the resulting map should be filtered before the authorization check, so that
batching does not widen what a caller can infer about pages it cannot read (#1046 territory).

alistair3149:
Fixes #1243

#1224 batched relation-target resolution for the write path and listed read-path callers under
"Considered and omitted", on the grounds that "nothing there resolves a set today". That premise
did not hold: both read queries already held a `SubjectIdList` and looped it, and
`Subject::getReferencedSubjects()` returns exactly what the batch methods take. So this needs no
new lookup API surface.

Every `getSubject` is a graph round trip through the subject → page index plus a revision load and
a content deserialize, and every `getPageIdOfSubject` is another round trip, since
`Neo4jPageIdentifiersLookup` answers the single-id method by running the batch query.

## Change

- `GetSubjectQuery` resolves the referenced ids in one `getSubjects()` and one
  `getPageIdsOfSubjects()` ahead of the loop, and reads both out of the returned maps.
- `GetPageSubjectsQuery` does the same for the referenced ids, and additionally resolves the
  hosting pages of the page's own Subjects in one call rather than one per Subject on the page.
  Collecting the referenced ids moves into `collectReferencedIds`, where `SubjectIdList`'s
  deduplication replaces the old "already in the result" check.
- `SubjectMap::getIds()` returns the ids as the `SubjectIdList` those lookups take.
- `PointInTimeSubjectLookup::getSubjects()` returns early for an empty id list, mirroring the guard
  `Neo4jPageIdentifiersLookup::getPageIdsOfSubjects()` already has. Without it, batching *added* a
  cost the old loop did not pay: a Subject with no relations now calls `getSubjects()` once instead
  of not at all, and that call deserialized the primary revision's slot in order to select nothing
  out of it — on every `?revisionId=` read of a Subject without relations, which is the common case.

Both loops iterate the requested ids rather than the returned map, so the response keeps the order
the Statements reach the targets. The map is keyed by id and promises no order — #1224 declined to
give it one precisely because every caller indexes it.

## Behaviour

Unchanged. The per-title read authorization stays per Subject, against that Subject's own hosting
page, so batching does not widen what a caller can infer about pages it cannot read (#1046).
`GetSubjectQuery` still resolves and authorizes the requested Subject before touching its targets,
so a denied request short-circuits without paying for them. The two queries keep their differing
treatment of a target with no resolvable hosting page: `GetSubjectQuery` serves it,
`GetPageSubjectsQuery` omits it. Both rules, and the short-circuit, now have tests; none did before.

The two dedup paths cannot diverge. The old loop skipped an id already present in the accumulating
result, which only held ids that had resolved *and* passed the gate; `SubjectIdList` dedups the id
itself. A repeat of an id whose first attempt failed is therefore no longer retried — but both
lookups are deterministic within a request, so the retry produced the same failure.

Verified end to end: 24 REST requests against a 5 000-Subject wiki, both endpoints across every
`expand` combination, byte-identical on `master` and on this branch (507 693 bytes each). The
`?revisionId=` path is byte-identical with and without the new guard.

## Measurements

Dev stack. Corpus from `NeoWiki:GeneratePerformanceDump`: 500 pages × 10 Subjects, 12 Statements
each of which 3 are relations to Subjects on other pages — the shape ADR 29 calls typical. 200
consecutive pages read once each, single PHP process, warm-up excluded, `master` and this branch
alternated three times each.

`GetSubjectQuery`, one Subject with 3 relation targets on 3 pages (`expand=page|relations`):

|                        | master        | this branch   |
|------------------------|---------------|---------------|
| Per read               | 24.2–27.9 ms  | 16.3–19.3 ms  |
| Graph round trips      | 8             | 4             |

`GetPageSubjectsQuery`, 10 Subjects reaching 30 distinct targets on 3 pages (`expand=relations`):

|                        | master         | this branch   |
|------------------------|----------------|---------------|
| Per read               | 202–208 ms     | 25.7–27.4 ms  |
| Graph round trips      | 70             | 3             |

The counts follow the loops exactly. `GetSubjectQuery` paid 2 + 2M and now pays a flat 4;
`GetPageSubjectsQuery` paid N + 2M and now pays a flat 3. Both were measured, not read off the
code: the harness wraps the production `PageIdentifiersLookup` and counts.

The wall-clock gap is wider than the round-trip gap on `GetPageSubjectsQuery` because
`MediaWikiSubjectRepository::getSubjects` groups the ids by hosting page, so the 30 targets cost 3
revision loads and deserializations instead of 30. Both savings grow with the number of targets;
neither depends on corpus size.

Numbers this small are dominated by a local Neo4j, so read them as ratios, not as absolutes.

## Test doubles

`InMemorySubjectLookup::getSubjects()` returned its map in the order the ids were requested, which
no implementation does — `MediaWikiSubjectRepository` unions per hosting page and
`PointInTimeSubjectLookup` puts the primary revision's Subjects first. That made the order the two
queries must preserve untestable: reading the batch out as a map instead of by the requested ids
left every test green while reordering the REST response's keys. The double now returns its own
registration order, and the tests asserting on response order register their Subjects in an order
other than the one they reference them in.

## Considered and omitted

- **Merging the two `getPageIdsOfSubjects` calls in `GetPageSubjectsQuery`.** The page's own
  Subjects and the referenced ones could be resolved in one call, saving one constant-time query
  out of three. It costs a `SubjectIdList` union and threads a shared map through
  `buildReferencedSubjectItems`, in exchange for no change in how the cost scales.
- **Avoiding the duplicate hosting-page query for referenced ids.** `SubjectLookup::getSubjects()`
  already runs `getPageIdsOfSubjects()` on the same ids internally, and the query then runs it
  again for the identifiers it puts in the response. Removing it means either returning identifiers
  from `getSubjects()` or caching inside the lookup — both real API decisions, neither an N+1 fix.
  Note this constant factor of 2 is inherited from `master`, which paid it per referenced Subject.
- **Filtering the id list down to the ids that resolved before asking for hosting pages.** It is
  one query either way, and the entries for unresolved ids are never read.
- **The remaining read-path N+1, `SubjectResolver::resolveRelationLabel`.** The Lua and
  parser-function accessors `array_map` it over a multi-value relation, one `getSubject` apiece.
  Same shape, different surface, and that path is parser-cached; it also needs a new collection-taking
  method rather than the existing batch ones. Filed as #1253.

## Surfaced, not changed here

`GetPageSubjectsQuery` authorizes the requested page and then serves each Subject's hosting-page
identifiers without checking that page. Where a stale projection or a Subject with `HasSubject`
edges from two pages makes the two differ, another page's title and namespace reach the caller
ungated. Master does exactly the same one `getPageIdOfSubject` call at a time, so this branch does
not introduce it — but `testResolvesHostingPagesOfPageSubjectsInOneLookup` is the first test to
assert the behaviour, so its fixture says so and points at the issue. Filed as #1252.

## Verification

- `make cs` green. Psalm reports nothing new; its only complaints in the touched files are
  `git blame`d to #1224.
- Green by filter: `GetSubjectQueryTest`, `GetPageSubjectsQueryTest`, `PointInTimeSubjectLookupTest`,
  `SubjectMapTest`, `SubjectValidatorTest`, `MediaWikiSubjectRepositoryTest`, `GetSubjectApiTest`,
  `GetPageSubjectsApiTest`, plus every other consumer of the `InMemorySubjectLookup` this branch
  changes (`ProposedSubjectValidatorTest`, `CreateSubjectActionTest`, `ReplaceSubjectActionTest`).
  `Domain` passes as a whole (552 tests).
- Full PHPUnit green on CI across MW 1.43, 1.44, 1.45, 1.46 and master, plus PHPCS and PHPStan.
  (Locally an unfiltered `make phpunit` stalls at ~0 % CPU before PHPUnit prints its banner, on this
  branch and on `master` alike, so CI is the signal for the whole suite here.)
- Both batching tests fail against `master`'s implementation of the queries.
- Four mutations of the new code, each failing exactly the test written for it: re-resolving per id
  inside the loop while keeping the batch call (the failure mode #1224 added its per-id counters
  for); iterating the returned map instead of the requested ids; dropping the collected-id filter;
  and treating an unresolvable hosting page as resolvable.
- REST byte-diffs above.

<sub>AI-authored — Claude Opus 5 (1M context); implementation, tests and benchmark; verified with the full PHPUnit suite on CI, phpcs, phpstan, five mutation checks, browser smoke tests of both endpoints, and byte-for-byte REST response diffs against master on a 5 000-Subject wiki and on the demo data.</sub>

### DIFF
--- a/src/Application/Queries/GetPageSubjects/GetPageSubjectsQuery.php
+++ b/src/Application/Queries/GetPageSubjects/GetPageSubjectsQuery.php
@@ -14,6 +14,7 @@ use ProfessionalWiki\NeoWiki\Domain\Page\PageId;
 use ProfessionalWiki\NeoWiki\Domain\Page\PageSubjects;
 use ProfessionalWiki\NeoWiki\Domain\Schema\SchemaName;
 use ProfessionalWiki\NeoWiki\Domain\Subject\Subject;
+use ProfessionalWiki\NeoWiki\Domain\Subject\SubjectIdList;
 use ProfessionalWiki\NeoWiki\Presentation\SchemaPresentationSerializer;
 
 readonly class GetPageSubjectsQuery {
@@ -39,25 +40,31 @@ readonly class GetPageSubjectsQuery {
 			: PageSubjects::newEmpty();
 
 		$mainSubject = $pageSubjects->getMainSubject();
+		$subjectsOnPage = $pageSubjects->getAllSubjects();
+
+		// Each Subject takes its own entry rather than the page's: these all came out of one page's
+		// content, but an unrebuilt or stale graph can still place two of them on different pages.
+		$hostingPages = $this->pageIdentifiersLookup->getPageIdsOfSubjects( $subjectsOnPage->getIds() );
+
 		$subjectItems = [];
 
 		if ( $mainSubject !== null ) {
 			$subjectItems[$mainSubject->id->text] = GetSubjectResponseItem::fromSubject(
 				$mainSubject,
-				$this->pageIdentifiersLookup->getPageIdOfSubject( $mainSubject->id )
+				$hostingPages[$mainSubject->id->text] ?? null
 			);
 		}
 
 		foreach ( $pageSubjects->getChildSubjects()->asArray() as $childSubject ) {
 			$subjectItems[$childSubject->id->text] = GetSubjectResponseItem::fromSubject(
 				$childSubject,
-				$this->pageIdentifiersLookup->getPageIdOfSubject( $childSubject->id )
+				$hostingPages[$childSubject->id->text] ?? null
 			);
 		}
 
 		$referencedSubjectItems = null;
 		if ( $includeReferencedSubjects ) {
-			$referencedSubjectItems = $this->buildReferencedSubjectItems( $pageSubjects->getAllSubjects()->asArray(), $subjectItems );
+			$referencedSubjectItems = $this->buildReferencedSubjectItems( $subjectsOnPage->asArray(), $subjectItems );
 		}
 
 		$schemas = null;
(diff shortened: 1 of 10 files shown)


src/Application/Queries/GetPageSubjects/GetPageSubjectsQuery.php
```diff
@@ -39,25 +40,33 @@ public function execute( int $pageId, bool $includeSchemas = false, bool $includ
 			: PageSubjects::newEmpty();
 
 		$mainSubject = $pageSubjects->getMainSubject();
+		$subjectsOnPage = $pageSubjects->getAllSubjects();
+
+		// One lookup for the whole page: these Subjects were read out of one page's content, so
+		// resolving them one at a time cost a round trip per Subject on the page to learn where
+		// they all live. Each item is still given its own entry out of the result, since an
+		// unrebuilt or stale graph can place two Subjects of one page on different pages.
+		$hostingPages = $this->pageIdentifiersLookup->getPageIdsOfSubjects( $subjectsOnPage->getIds() );
```
JeroenDeDauw, line 47: That comment does not seem useful. Seems to be speaking to the reviewer

PR #1257, opened 2026-08-05
Clear the error and the deprecations the CI test run reports

alistair3149:
Clears what the CI test run reports: one error, 96 PHP deprecation lines and 10 Node 20 warnings. Seven commits, one concern each.

**`composer phpunit:config`** was called unconditionally, but MediaWiki ships that script only from 1.46. Under `GITHUB_ACTIONS` Composer writes the resulting error as a `::error ::` annotation on **stdout**, which the `2>/dev/null` guarding stderr never caught. It now runs where core ships the generator.

**`$tablesUsed`** in `GeneratePerformanceDumpTest` fed nothing — core has detected changed tables by itself since 1.41 and dropped the property in 1.44 — while raising a dynamic-property deprecation from 1.44 on.

**`SpecialGraphStores`** named its right through the `SpecialPage` constructor, deprecated in 1.46 and logged 88 times per run. From 1.46 enforcement, listing and the denial page read `getRestriction()`; up to 1.45 they read the property that parameter set. All three are restated against `getRestriction()`, which holds on every supported version without a version check.

**The Neo4j waits** drop `iFaxity/wait-on-action`, which has no release that is not Node 20. The PHPUnit one becomes a bounded retry matching the QLever, Oxigraph and Fuseki waits beside it. The image build's is deleted outright: since #1284, `make up` cannot return while db or neo is still starting, which also retires the `/wait-for-it.sh` calls in `install-db` and `load-neo4j-users`. The test backends keep theirs, as nothing `depends_on` them.

The test script also disables Composer's process timeout, as core does for its own, so a full local run is no longer killed at 300 seconds.

## Why it looks like this

* **`phpunit:config` is guarded, not deleted.** NeoWiki passes its own `-c`, so the generated config looks unused — but core's bootstrap, which this script passes via `--bootstrap`, throws without it from 1.46.
* **Three overrides, not a version check.** `version_compare( MW_VERSION, '1.46', '<' )` works but exercises only one of its branches per run. Overriding `getRestriction()` alone is not enough: reduced to that, `testAUserWithoutTheAdminRightIsRefused` fails on 1.43.
* **`isRestricted()` is load-bearing.** `getUsablePages()` lists a page when `isListed() && ( !isRestricted() || userCanExecute( $user ) )`, so a page reporting itself unrestricted is listed for everyone.
* **`MediaWikiServices` rather than constructor injection.** Injecting `GroupPermissionsLookup` would make this the only special page with a constructor dependency and the only ObjectFactory `SpecialPages` entry, while the class keeps locating its other three collaborators at point of use. `SpecialPage` exposes no accessor for it on any supported version.
* **No `--health-cmd` on the Neo4j service.** It works, but the runner gates every step on service health and polls with exponential backoff — about 2, 6, 14, 30 then 62 seconds — while Neo4j answers at around 35. Measured at 26 seconds a matrix leg.
* **`actions/delete-package-versions` stays on Node 20.** No release of it is otherwise, and replacing it means a destructive loop over the packages API that only ever runs on master, so its first execution would be its first test.

## Verification

CI green on all seven jobs. Against the previous master run: `phpunit:config` errors 3 → 0, PHP deprecation lines 96 → 0, Node 20 warnings 10 → 0. Test count is baseline +2 on every version with skips unchanged; assertion totals vary between runs of the same commit, so they are not compared.

The restriction is checked on a live 1.43.8 wiki rather than only through PHPUnit: an anonymous request to Special:GraphStores is refused with *"limited to users in the group: Administrators"* — message key `badaccess-groups`, so the right reached the denial page rather than an empty string — and the page appears zero times on Special:SpecialPages for that user, while the unrestricted `Special:Schemas` appears twice.

Both new tests were mutation-checked on 1.43: overriding only `getRestriction()` fails the refusal test, and delegating `isRestricted()` to the parent fails the listing test with the page leaking into a non-admin's list.

The Makefile change was checked on a cold start with the volumes destroyed: `install-db` and `load-neo4j-users` both succeed unguarded, and `mediawiki_read` connects afterwards.

Locally, `make cs` is clean. The Scribunto/Lua subtree stalls on this stack independently of this change, so the full local suite was not run to completion.

<sub>AI-authored — Claude Code, `Opus 5 (1M context)`; root-cause analysis, fix and verification in-session; not yet human-reviewed.</sub>

### DIFF
--- a/.github/workflows/ci-php.yml
+++ b/.github/workflows/ci-php.yml
@@ -124,11 +124,12 @@ jobs:
 
       - run: composer update
 
+      # Asks Neo4j to answer a query rather than to open its port, which is what the next step needs of it.
+      # Not a `--health-cmd` on the service: that gates every step on it, measured at some 26 seconds a leg.
       - name: Wait for Neo4j
-        uses: iFaxity/wait-on-action@v1
-        with:
-          resource: http-get://localhost:7689
-          timeout: 60000
+        run: |
+          timeout 90 bash -c 'until docker exec test_neo cypher-shell -u neo4j -p password -a bolt://localhost:7689 "RETURN 1" >/dev/null 2>&1; do sleep 2; done' \
+            || { echo "Neo4j did not become ready in time; dumping logs:"; docker logs test_neo; exit 1; }
 
       # QLever for the SPARQL query system test. It cannot be a `services:` container: its bring-up needs
       # a multi-step entrypoint (build an empty index, then serve) that a service container cannot express
--- a/.github/workflows/ci-php.yml
+++ b/.github/workflows/ci-php.yml
@@ -205,33 +205,18 @@ jobs:
 
       - name: Wait for QLever
         run: |
-          for i in $(seq 1 40); do
-            if curl -fsS -o /dev/null "http://localhost:7019/?query=SELECT%20%2A%20WHERE%20%7B%20%3Fs%20%3Fp%20%3Fo%20%7D%20LIMIT%201"; then
-              echo "QLever is serving"; exit 0
-            fi
-            sleep 2
-          done
-          echo "QLever did not become ready in time; dumping logs:"; docker logs test_qlever; exit 1
+          timeout 90 bash -c 'until curl -fsS -o /dev/null "http://localhost:7019/?query=SELECT%20%2A%20WHERE%20%7B%20%3Fs%20%3Fp%20%3Fo%20%7D%20LIMIT%201"; do sleep 2; done' \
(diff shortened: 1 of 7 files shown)


.github/workflows/ci-php.yml
```diff
@@ -120,11 +120,18 @@ jobs:
 
       - run: composer update
 
+      # Waits by asking Neo4j to answer a query, which is what the next step needs of it, rather than by
+      # opening its port. A retry loop like the QLever and Oxigraph waits below rather than an action:
+      # iFaxity/wait-on-action has no release that does not run on Node 20.
```
JeroenDeDauw: 🤔

PR #1260, opened 2026-08-06
Add optional-service selection to the dev stack (services=)

alistair3149:
Lets a dev stack run only the optional services a task needs. `make dev services=<comma-list>` picks among `node`, `qlever`, `mailcatcher` and `oxigraph`; `services=none` runs core only (mediawiki, db, neo). Without the flag everything runs as it does today, so existing workflows are unchanged.

- Optional services sit behind compose profiles named after them; the Makefile turns the selection into `COMPOSE_PROFILES` (a user-set `COMPOSE_PROFILES` is unioned in) and stops newly-deselected services when `make dev` reruns.
- Wiki config follows the selection: `QLEVER_URL` / `MW_SMTP_HOST` are set empty when their service is off (empty means disabled), so the wiki never targets a dead endpoint.
- Stacks without the `node` watcher get a one-shot `dist/` build so the UI works; every `ts-*` target starts `node` on demand.
- New `make test-backends-stop` reclaims the test backends' ~0.5–0.7 GiB after a testing phase; `make phpunit` starts them again on demand.
- `make print-services` shows the selection; `Docker/tests/test-services.sh` (in `make test-scripts`) covers the selection logic.
- Docs: new [Optional services](https://github.com/ProfessionalWiki/NeoWiki/blob/dev-stack-service-selection/Docker/README.md#optional-services) section in Docker/README.md.

An idle full dev stack holds ~1.7 GiB; `services=none` drops ~1 GiB of that (node watcher, mailcatcher, qlever headroom) for tasks that need none of it.

Try it: `make dev services=none` then `make print-services` and `docker ps`; `make dev` restores the default set.

<details><summary>Design notes</summary>

qlever's profile tag lives in the dev overlay, not the base compose file, so the demo stack and raw `docker compose` invocations keep the store unconditionally while every dev-stack path is gated. The Makefile owns the `QLEVER_URL` value (endpoint when selected, empty when not) instead of compose interpolation defaults, keeping "no store running ⇒ none configured" true for non-make invocations.

</details>

> <sub>AI-authored — Claude Code, `Fable 5`; brainstormed from an open question by @alistair3149, design redirected several times (no presets, full default, per-repo split); diff not yet human-reviewed; per-task + whole-branch AI review, script suites green, live end-to-end matrix on a dev stack (core-only, escape-hatch union, restore, error paths).</sub>

### DIFF
--- a/Docker/README.md
+++ b/Docker/README.md
@@ -38,14 +38,49 @@ need to run two tools-mode stacks simultaneously.
 The MariaDB, (default) Neo4j, QLever and Oxigraph ports are not exposed to the host. Reach
 them from inside the stack via `make bash` or `docker compose exec`.
 
+## Optional services
+
+The dev stack's core is `mediawiki`, `db` and `neo`; everything else is optional. `make dev` (and
+`make dev-tools`) accept a `services=` comma-list naming the optional services to run:
+
+| Service       | Needed for                                                              |
+|---------------|-------------------------------------------------------------------------|
+| `node`        | Frontend work: watches and rebuilds the TS bundle (`dist/`)             |
+| `qlever`      | SPARQL/RDF work: the wiki projects RDF into it on every save            |
+| `mailcatcher` | Testing outgoing email (web UI on `MAILCATCHER_PORT`)                   |
+| `oxigraph`    | Second SPARQL engine; needs manual wiring via `LocalSettings.local.php` |
+
+Without the flag, `node`, `qlever` and `mailcatcher` run (the default); `oxigraph` only runs when named.
+`services=none` runs core only. Examples:
+
+```sh
+make dev                        # default set
+make dev services=node          # frontend work: watcher only
+make dev services=none          # core only
+make dev services=qlever,node   # SPARQL plus frontend
+```
+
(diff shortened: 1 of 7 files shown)


Docker/README.md
```diff
@@ -38,14 +38,50 @@ need to run two tools-mode stacks simultaneously.
...
+| `mailcatcher` | Testing outgoing email (web UI on `MAILCATCHER_PORT`)                   |
+| `oxigraph`    | Second SPARQL engine; needs manual wiring via `LocalSettings.local.php` |
+
+Without the flag, `node`, `qlever` and `mailcatcher` run by default; `oxigraph` runs only when named.
+`services=none` runs core only. Examples:
+
+```sh
+make dev                        # default set
+make dev services=node          # frontend work: watcher only
+make dev services=none          # core only
```
JeroenDeDauw, line 59: `services=none` is the same as omitting? If so, could omit it from the docs (and perhaps the code)

alistair3149: They are different.
- `none` means only the minimal services needed for NeoWiki (`mediawiki`, `db`, `neo`).
- Omitting means having all the services, that matches the existing `make dev` behavior.

PR #1265, opened 2026-08-07
Implement the Subject Sources foundation

### LINKED ISSUE #993: Implement the Subject Sources foundation (ADR 23)

### LINKED ISSUE #1341: Decide the permission check and degradation behaviour for cross-wiki subject display
Rendering a subject from another wiki goes through the REST API, not Cypher, so the query-side scoping that covers
raw query surfaces does not apply to it. ADR 27 makes MediaWiki the sole permission authority and the page the unit
of access control, but it does not say which check runs when the page owning the subject lives in a different wiki
of a farm.

To decide: the check that applies, and the degradation behaviour when the schema or the subject is not accessible.
A denied read must stay indistinguishable from absent data, so the display needs a defined fallback rather than an
error that reveals existence. Relates to [ADR 23: Subject Sources](https://neowiki.ai/docs/adr/023-subject-sources).

This decision was deferred when [ADR 27: Access Control](https://neowiki.ai/docs/adr/027-access-control) was
accepted.

> <sub>AI-authored — Claude Code, `Opus 5 (max)`, revised by `Fable 5 (max)`; detailed spec from @JeroenDeDauw, revised once on his review; not yet human-reviewed in full; text derived from the published ADR 27 wording, nothing executed.</sub>

JeroenDeDauw:
Fixes https://github.com/ProfessionalWiki/NeoWiki/issues/993
Fixes https://github.com/ProfessionalWiki/NeoWiki/issues/1341

The [ADR 23](https://github.com/ProfessionalWiki/NeoWiki/blob/master/docs/adr/023-subject-sources.md) foundation. With
only the local Source registered, every path resolves what it did before: nothing user-visible changes and nothing
stored changes.

## What lands

**T1 — `SubjectId` is a `(source, localId)` pair.** A bare `s` + 14 nanoid characters still means a local Subject and
is still stored as that bare string; anything else is `sourceKey:localId`. Every place an id arrives as text — a REST
path, a revision slot, a relation target, a Lua call — goes through `SubjectIdParser`, which is handed the local Source
key and resolves nothing. Local slot JSON, Neo4j `id` and API output are unchanged, pinned by a byte-identical slot
test.

**T2 — the Source contract, its registry, and `LocalSource`.** One interface, one registry keyed by Source key, the
local revision slot refactored into the default registered Source under the MediaWiki Wiki ID. Extensions contribute
Sources through `NeoWikiRegistrar::addSource()`, alongside the Property Type and Page Property Provider registries.
ADR 23's "Source interface contract" open question is resolved in the ADR.

**T3 — a Schema reference carries its Source.** `SchemaReference` pairs a Source with the unchanged `SchemaName`, in
the two places a reference is persisted: a Subject's `schema` and a relation property's `targetSchema`. Resolution
routes through the Source the reference names, which need not be the Subject's.

**T4 — a relation target naming an unregistered Source is an `error`-severity violation**, raised on the write and
validate paths and never on read of persisted data. It blocks like every other error (ADR 26): under
`$wgNeoWikiEnforceValidation`, and only when the edit introduces it, so a Subject already carrying such a target stays
editable. Where one is stored anyway, it degrades on read — the projections drop the edge and the triple, and
resolution yields no Subject.

## Design calls

- **Serialized form `sourceKey:localId`, split at the first colon.** Source key `[A-Za-z][A-Za-z0-9_-]{0,63}`; a
  foreign localId is RFC 3986 `pchar`, minus percent-encoding and minus `&` and `'`, capped at 256 characters. Every
  PHP grammar is `\z`-anchored, so a trailing newline cannot pass.
- **The local key is the Wiki ID, and is exempt from that grammar.** A wiki whose id is not a well-formed Source key
  still resolves its own bare ids; only the explicit `<wikiId>:<localId>` spelling is unusable there.
- **An explicitly-local reference canonicalizes to bare**, for Subject ids and Schema references alike, so one thing
  has one identity — which the id-keyed maps rely on.
- **A foreign Schema reference persists as `{source, name}`.** A bare string is always a local name, so a Schema title
  containing a colon (`ISO:9001`) stays valid.
- **No client-side source validation** (ADR 25). A server violation renders through the field-error path the editor
  already has.
- **Sources register lazily**, as factories built on first use: registration runs under an early hook where
  constructing clients would be waste, and a Source nobody references is never built.
- **A Source vouches for what it serves.** Everything a Source returns is readable by every reader of this wiki, so
  NeoWiki performs no per-user authorization on sourced Subjects; a Source over restricted data serves only its
  unrestricted part. Read paths decide on `isLocal()`, never on whether a hosting page was found — local Subjects
  keep the page-based checks. ADR 27 is amended with this as its fifth read class, *source-attested*, resolving its
  deferred cross-wiki subject display question; per-user granularity for sourced data is out of scope until demanded.
- **A local slot holds local Subjects only.** The slot now *accepts* qualified ids and object Schema references, so
  deserialization skips a non-local subject key with a warning and the index, Neo4j and RDF projections agree. Over
  REST and Lua, `schema` and `targetSchema` are references — a bare string when local, the reserved `{source, name}`
  object otherwise; consumers tolerate the object form, source-aware display is
  https://github.com/ProfessionalWiki/NeoWiki/issues/1373.
- **`isEditable()` and foreign-source `isValidLocalId()` are contract surface with no consumer** until sourced-subject
  rendering lands. They are frozen now because a Source author must answer them from the start.
- **Foreign relation edges and unresolvable Schema references degrade** — skipped with a warning — rather than
  project. The projection skip covers every foreign target, resolvable or not; T4 refuses only the unresolvable ones.
- **`RelationId` shared `SubjectId`'s `$`-anchor newline flaw**, fixed here. Pre-existing, one line.

## Considered, omitted

- A write-capability stub on `Source`. Write-back brings its own method when it is built.
- A TypeScript `SubjectIdParser`. Only canonical ids reach the frontend, so there is nothing to canonicalize there.
- Registry-aware colon splitting of Schema-name strings. Rejected: registering a Source later would retroactively
  reinterpret a Schema name that already contains a colon.
- Neo4j stub nodes for foreign relation targets. The stub would carry this wiki's `wiki_id`, and one without it scopes
  no query correctly.

## Follow-ups filed

- https://github.com/ProfessionalWiki/NeoWiki/issues/1266 — the validate endpoint is a read-authorization oracle
  (pre-existing).
- https://github.com/ProfessionalWiki/NeoWiki/issues/1267 — `ViolationDiff` identity ignores `args`.
- https://github.com/ProfessionalWiki/NeoWiki/issues/1370 — Sources have no way to materialise their Subjects.
- https://github.com/ProfessionalWiki/NeoWiki/issues/1371 — contract gaps an adapter author hits.
- https://github.com/ProfessionalWiki/NeoWiki/issues/1372 — farm-sibling registration fails on grammar-invalid Wiki IDs.
- https://github.com/ProfessionalWiki/NeoWiki/issues/1373 — frontend source-awareness.

## Reviewing

The three commits stage the foundation — identity, the Source contract, Schema references and the target guard —
with the review findings from https://github.com/ProfessionalWiki/NeoWiki/pull/1330 folded in and co-authored. Five
further commits act on three independent blind-spot reviews (authorization model, tolerant read-side boundary,
local-only slots, registry honesty, UI guards and docs), and one records the resulting ADR 27 amendment. PHP and TypeScript assert the id grammars against one shared fixture, `tests/vectors/subject-ids.json`.

> <sub>AI-authored — Claude Code, `Opus 5 (max)`; detailed in-session spec from @JeroenDeDauw; diff not yet human-reviewed; phpcs, phpstan, PHPUnit and the TypeScript gates green locally at each revision, and all CI checks green on the current head.</sub>

Design and orchestration by `Fable 5 (max)`; implementation and the post-review fix batch by `Opus 5 (max)` subagents.
Four independent AI review passes — code, security, tests, and a diff scan against a superseded branch — with every
blocking finding fixed before this description was written.

</details>


JeroenDeDauw:
I did not fully review this yet

JeroenDeDauw:
Note: before merging I'd like to ALSO finish my own review. Prior/additional review by others is useful and welcome

JeroenDeDauw:
Thanks @alistair3149. I've resumed my review and identified some structural issues (not related to your changes). Will continue on this tomorrow.

PR #1270, opened 2026-08-08
Answer 400 instead of 500-ing or deleting on a misshapen Statement write

JeroenDeDauw:
Follows-up to https://github.com/ProfessionalWiki/NeoWiki/pull/1216

Review fixes for #1216. Each is a request shape the new endpoints answer wrongly today, found by review and
reproduced over HTTP against a dev stack before and after the fix.

image

## Answered 500

* **A property type that is not a string.** `{"statement": {"propertyType": 123, …}}`. MediaWiki's body validator
  types the declared fields but nothing inside them, so it reached `?string` parameters. Also reachable through
  `PUT /subject/{subjectId}`.
* **A property name PHP reads as a decimal integer**, e.g. `.../statements/2024`. Property names travel as array
  keys, which PHP casts to `int`, and `PropertyName`, `Schema` and `StatementDeserializer` are typed. The last of
  those sits on the read-back path, so the write would have persisted a revision that could no longer be
  deserialized.

## Answered 200 after removing the Statement

Each of these reached the write path as an empty value, which the write paths read as the Statement being absent,
so a request meaning *set this* removed it and reported success:

* a JSON object where a list belongs — it unpacks into the value objects as named arguments instead of raising a
  type error, and was stored as an object no reader and no graph store can hold;
* a relation value given as bare target ids rather than relation objects, whose entries were skipped one by one;
* a statement carrying no `value` at all, which only number and boolean properties rejected.

`{"value": []}` still removes the Statement, as documented.

## Wrote to the wrong property

The Neo4j node property map is assembled with `array_merge`, which renumbers integer keys. A property named `2024`
projected its value onto a property named `0`, so the graph disagreed with the revision slot.

## Notes

* `UpdateStatementAction::setStatement` now takes `propertyType` as `mixed`. It comes from the unvalidated part of
  the body, and `StatementListBuilder` is where the other input-shape checks live, so the type reaches that check
  rather than failing at the action's own signature.
* Requiring `value` is the one judgement call here: the request shape is ambiguous between *set* and *remove*, and
  this reads it as neither. `testStatementWithoutAValueReturns400` in #1216 already asserts that reading for number
  properties; this makes it hold for the rest. Easy to drop if you meant the other reading.
* Everything except the `value` requirement is equally reachable through `PUT /subject/{subjectId}`, so the fixes
  are in the shared classes and that endpoint improves with them.

Full suite and `make cs` green.

---
🤖 Written by [Claude Code](https://claude.com/claude-code) using `claude-opus-5[1m]` (max)

### DIFF
--- a/docs/api/subject-format.md
+++ b/docs/api/subject-format.md
@@ -195,7 +195,7 @@ Subject's label and its other Statements as they are:
 
 | Field | Required | Notes |
 |-------|----------|-------|
-| `statement` | Yes | A [Statement object](#statement-object). `propertyType` may be omitted, and then takes the type the Subject's Schema currently gives the property; a property the Schema does not define is rejected with `400`. A value that is empty for its type removes the Statement. |
+| `statement` | Yes | A [Statement object](#statement-object) carrying a `value`. `propertyType` may be omitted, and then takes the type the Subject's Schema currently gives the property; a property the Schema does not define is rejected with `400`. A value that is empty for its type removes the Statement. |
 | `comment` | No | Edit summary. |
 
 `DELETE` on the same path removes the Statement and takes only `comment`. Removing a Statement the Subject does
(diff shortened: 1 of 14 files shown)


JeroenDeDauw:
LGTM

PR #1277, opened 2026-08-09
Verify Fuseki as a third SPARQL store engine

JeroenDeDauw:
Runs the existing SPARQL write+query system test against Apache Jena Fuseki in CI: a
conformance check on the SPARQL NeoWiki emits, not a new engine to ship. It passes the
shared test case unchanged — no source change and no accommodation, so the subclass
differs only in its endpoint URLs, its absent token, and the strict default-graph
semantics it shares with Oxigraph.

Fuseki stays CI-only: no dev-stack service and no phpunit.xml.dist entry, so
`make phpunit` skips this suite while CI, which starts the server and sets
FUSEKI_TEST_BASE_URL, runs it. The two shipped stores still fail rather than skip when
unset, because an unreachable store there means a broken stack. Apache publishes no
container image, so CI runs the release tarball Apache itself publishes, fetched from
the ASF archive (cached) and verified against a checksum recorded in the workflow; the
version is pinned because the reference to test against should change deliberately.

Considered, omitted: a dev-stack service for it (nothing local needs Fuseki, and the
mandatory test backends stay at three); Basic-auth support for a dataset secured with
Fuseki's own authentication (untested — the docs point at a reverse proxy, as they do
for Oxigraph).

> <sub>AI-authored — Claude Code, `Opus 5 1M (max)`; detailed spec, run unattended as a background job, no corrections; diff not yet human-reviewed; suite run locally against a live Fuseki (6 tests, 13 assertions) and with the variable unset (6 skipped, no errors), all three engine suites green together (18 tests, 39 assertions), phpcs and phpstan clean. Image-sourcing hardening (ASF tarball + checksum pin instead of a third-party image) by `Fable 5 (max)`, verified via this PR's CI.</sub>

### DIFF
--- a/.github/workflows/ci-php.yml
+++ b/.github/workflows/ci-php.yml
@@ -167,15 +167,34 @@ jobs:
       # conformance target, not as an engine NeoWiki ships: Jena is the most complete SPARQL 1.1
       # implementation to check the emitted queries and updates against. The version is pinned for that
       # reason — the reference to test against should change deliberately, not with upstream releases.
-      # Apache publishes no image of its own, so this is a third-party build of its distribution.
-      # `--mem /ds` serves an in-memory dataset unauthenticated on /ds/query and /ds/update, which is all
-      # a fresh-per-run CI store needs, and keeps named graphs out of the default graph the way Oxigraph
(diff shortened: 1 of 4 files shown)


JeroenDeDauw:
LGTM

Issue #1140, opened 2026-07-23
Make subject label default value configurable / extendable


osnard:
In BlueSpice, we show the subjects of a page in a dedicated column. Each subject has a heading, which currently the subject label. Unfortunately, as the default subject label is the page name it is being created on, this leads to some odd output. In addition, the subject label can not be changed after it has been created.

Example: 
<img width="2332" height="1228" alt="Image" src="https://github.com/user-attachments/assets/f52ff057-bad1-49c1-bcb0-666b0b97739e" />

It would be nice, if we could either completely omit the subject label from the form and calculate it. In this case we just need to change the rendering in the right column (e.g. showing the schema name rather than the subject label), or set the default value in a different way.

JeroenDeDauw:
Done: subject labels are now editeable and the defaults are better (schema name instead of page name)

Next:
* https://github.com/ProfessionalWiki/NeoWiki/issues/1283

Issue #192, opened 2024-10-17
Existing Subject should not have editable label


malberts:
![image](https://github.com/user-attachments/assets/ed069e59-77e7-43ee-8889-4b1e1835084a)

PR #1281, opened 2026-08-10
Make the subject label editable in place in the subject editor dialog

alistair3149:
For https://github.com/ProfessionalWiki/NeoWiki/issues/1140

The subject label could not be changed after creation. The edit dialog's title is now the label itself,
edited in place: a pencil button next to the title swaps it for an input; Enter or blur commits, Escape
reverts. The label stays out of the statement form, since it is the subject's identity rather than one of
its statements.

* New reusable `EditableText` component (`components/common/`): inline edit-in-place with commit/cancel
  semantics, IME-composition guards, focus restoration, and a `required` mode that keeps a blank draft
  open in error state instead of returning to display mode. Intended second consumer: the schema
  description editor.
* `SubjectEditorDialog`: title replaced by the component; `label-required` violations render at the title
  (input error status, `role="alert"` message) instead of the form-level banner; blank labels cannot be
  saved; header polish (title-group gap, subtitle size).
* The backend already supported label replacement (`ReplaceSubjectApi`); this is a frontend-only change.

Escape cancels on keyup with stopped propagation because Codex dialogs close on the Escape keyup: a
keydown-time cancel would unmount the input and let the release close the dialog.

Considered, omitted: `aria-describedby` from the input to the error message. Codex's `TextInput` binds
that attribute from its private `CdxField` injection (a Symbol), so it cannot be set from outside;
`role="alert"` plus `aria-invalid` carry the announcement instead.

## Manual Browser Check

1. Open a page whose main subject has an infobox (demo data: `Acme Anvil`) and click the infobox's edit button.
2. The dialog title is the subject label with a pencil button next to it; the body contains only statement fields.
3. Click the pencil, type a new label, press Enter: the title updates, focus returns to the pencil, and Save changes enables.
4. Save. The success toast and the infobox header show the new label; reload the page and confirm it persisted.
5. Reopen the editor, click the pencil, delete the text, press Enter: the input stays open with a red border and "Please provide a label for the subject." below the title; Save changes refuses with an error toast.
6. Press Escape while renaming: the draft is discarded, the previous label returns, and the dialog itself stays open.

> <sub>AI-authored — Claude Code, `Fable 5 (xhigh)`; #1140 worked through with @alistair3149, who picked the editable-title direction from a mock comparison; ~6 rounds of UX steering; UI checked in-browser by @alistair3149, diff not yet human-reviewed; TDD (all new specs seen red then green), two subagent review rounds acted on, full vitest + stylelint + vue-tsc + build green, browser-verified end to end incl. Neo4j persistence and the Escape/dialog interaction.</sub>

### DIFF
--- a/extension.json
+++ b/extension.json
@@ -525,6 +525,8 @@
 				"neowiki-delete-error",
 				"neowiki-subject-editor-title",
 				"neowiki-subject-editor-save",
+				"neowiki-subject-editor-label-field",
+				"neowiki-subject-editor-rename",
 				"neowiki-schema-editor-new-property",
 				"neowiki-schema-editor-delete-property",
 				"neowiki-schema-editor-optional",
--- a/i18n/en.json
+++ b/i18n/en.json
@@ -118,6 +118,8 @@
 
 	"neowiki-subject-editor-title": "Editing $1",
 	"neowiki-subject-editor-save": "Save changes",
+	"neowiki-subject-editor-label-field": "Subject label",
+	"neowiki-subject-editor-rename": "Rename subject",
 
 	"neowiki-schema-editor-new-property": "Add property definition",
 	"neowiki-schema-editor-delete-property": "Delete property definition",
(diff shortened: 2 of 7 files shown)


alistair3149:
https://github.com/user-attachments/assets/35555b12-a675-4e41-ab55-1560c8d9173f

JeroenDeDauw:
Seems like we lost this capability compared to the original prototype: https://github.com/ProfessionalWiki/NeoWiki/issues/192

PR #1285, opened 2026-08-10
Rename the native RDF term namespace from /ontology/ to /vocab/

JeroenDeDauw:
The native RDF projection now emits its `neo:` terms under `<baseUri>/vocab/` instead of `<baseUri>/ontology/`. These terms are the wiki's own local vocabulary of names, not an ontology in the sense of a shared conceptualization — a distinction raised in partner review of the mapping docs (https://github.com/ProfessionalWiki/NeoWiki/discussions/996#discussioncomment-17920073). NeoWiki is not in production, so there are no published IRIs or stored exports to migrate; the same rename after release would invalidate all of them.

> <sub>AI-authored — Claude Code, `Opus 5 (max)`; detailed spec from @JeroenDeDauw, no revisions; diff not yet human-reviewed; full PHPUnit suite and phpcs/phpstan run locally, both green.</sub>

### DIFF
--- a/docs/rdf/rdf-export.md
+++ b/docs/rdf/rdf-export.md
@@ -27,7 +27,7 @@ All NeoWiki IRIs live under `$base` (`$wgNeoWikiRdfBaseUri`). Standard vocabular
 
 | Prefix | Namespace | Used for |
 |---|---|---|
-| `neo:` | `$base/ontology/` | NeoWiki vocabulary terms (`neo:Page`, `neo:Relation`, `neo:hasSubject`, `neo:source`, …) |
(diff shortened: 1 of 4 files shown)


JeroenDeDauw:
@Habennin does this make sense to you?

Habennin:
@JeroenDeDauw , while a NeoWiki schema may not be using a formally defined ontology in the full sense of a 'share conceptual space' defined with a process, as a technical object it is a collection of what can be considered classes and properties that relate classes. This is more an ontology than a vocabulary since a vocabulary as a technical object is only classes or properties defined in a hierarchy.

So I think it would be safe and even preferable to leave it as ontology. The more a NeoWiki instance were used, the more likely their local classes and properties will be treated as and used as a formal ontology.

Issue #1259, opened 2026-08-06
Special:GraphStores reports In sync after page projections have failed


alistair3149:
A store that fell behind during a backend outage keeps reporting **In sync**, so the page an operator would check to
find out reports the opposite of the truth.

## Cause

`GraphStoreStatusLookup::stateOf()` derives the state from the last successful rebuild and whether the projection was
redefined since:

```php
if ( $lastSuccessfulRun === null || $lastSuccessfulRun->failed > 0 ) {
	return StoreSyncState::NeverBuilt;
}

return $projectionChanged === null ? StoreSyncState::InSync : StoreSyncState::Stale;
```

Neither input moves when a *save* fails. `FailureIsolatingGraphDatabasePlugin` swallows the failure so the edit still
commits, and writes no run record, so the store keeps reporting the last rebuild's verdict indefinitely.

## Reproduction

1. Rebuild a store so it reads In sync.
2. Point the backend at an unreachable host.
3. Save some pages. Each edit commits; each projection fails.
4. Check the status.

Measured on a dev stack:

```
after rebuild:  neo4j  state=in-sync  lastRebuild=20260806005636  processed=123  failed=0
after 3 saves:  neo4j  state=in-sync  lastRebuild=20260806005636  processed=123  failed=0
```

while the two sides had actually diverged:

```
MediaWiki:  126 pages, including RotProbe1, RotProbe2, RotProbe3
Neo4j:      123 pages, none of them matching RotProbe
```

## Why it matters

Swallowing the failure is deliberate and correct — the graph is rebuildable, so an outage must not block editing. The
consequence is that the log line in `FailureIsolatingGraphDatabasePlugin` is the *only* signal that the store fell
behind, and MediaWiki routes no log channel anywhere by default, so on a stock install there is no signal at all.

It also undercuts the guidance in `docs/operations/maintenance.md` — "Once Neo4j is back, rebuild the graph" — because
nothing tells the operator a rebuild is due. The store reads healthy, and the drift is silent and permanent until
somebody happens to rebuild.

## Possible directions

Not designed, listed only to frame the discussion:

- Record failed projections per store (a counter, or a marker with a timestamp) and let the state reflect them, cleared
  by the next rebuild that reconciles every page.
- Treat "a save failed for this store since the last rebuild" as a staleness input alongside projection change.
- Leave the state as it is and surface a separate "projection failures since last rebuild" figure on the page, so the
  In sync verdict keeps its current meaning.

[...]

PR #1287, opened 2026-08-10
Detect an unreachable SPARQL store by probing it in initialize()

### LINKED ISSUE #1279: Rebuilds cannot detect a dead SPARQL store because its liveness probe is a no-op
The rebuild guards against an unreachable store in two places, both by calling `GraphDatabasePlugin::initialize()` and reading a throw as "this store cannot be reached": once when a run starts (a store that will not open is meant to end the run before a page is read), and once mid-run in `GraphRebuildExecutor::whyTheStoreCannotBeReached()`, where a wholly failed full batch fails the run with the cursor rewound to the batch start instead of walking on.

`SparqlProjectionStore::initialize()` is a documented no-op — a SPARQL store has no store-level structures to create up front — so it can never throw, and both guards are dead for every SPARQL store. Only Neo4j's `initialize()` does real I/O (constraint creation through the client), so the probes work only there.

With an unreachable SPARQL store the run therefore starts normally, every page's update POST fails, `everyOfferedPageFailed()` fires the probe, the no-op answers "the store is fine", and the batch's pages are recorded as per-page failures with the cursor advancing. The run walks the entire wiki through timeout-bound HTTP attempts and completes with every page marked failed, where the design intends an early, resumable "store is gone" stop. Related: #1259 — after such a walk the store's sync state misreports too.

A fix needs a decision on where a real probe lives: `initialize()` also runs on the extension's propagating update path, so giving the SPARQL implementation a genuine round trip changes behavior there too. The alternatives are a dedicated liveness check on the plugin contract, or the rebuild probing SPARQL stores with a cheap no-op update.

> <sub>AI-authored — Claude Code, `Fable 5 (max)`; surfaced by a code-mapping subagent during the #1280 bulk-load exploration, filing approved by @JeroenDeDauw in-session; not yet human-reviewed; code path re-read and confirmed on origin/master, not reproduced against a live store.</sub>

JeroenDeDauw:
Fixes https://github.com/ProfessionalWiki/NeoWiki/issues/1279

A rebuild asks a store whether it is still there by calling `initialize()` and reading a throw as
"cannot be reached": when a run starts, when a background batch picks it back up, and whenever a
whole batch has failed — the last being what ends a run with its cursor rewound to that batch rather
than walking on. `SparqlProjectionStore::initialize()` was a no-op that could never throw, so every
guard was dead for every SPARQL store. A rebuild against a stopped store walked the entire wiki,
failed every page one at a time, and reported itself as succeeded.

`initialize()` now reaches the store: it sends the update endpoint a
`DROP SILENT GRAPH <{base}/graph/liveness-probe>`, minted in `RdfNamespaces`, where every page graph
ends in `/page/{id}` — so nothing else can name the probe graph and nothing writes to it. Dropping an
absent graph leaves the store untouched, asks no more of an engine than `deletePage()` already does,
and goes to the update endpoint under the store's own credentials, because an endpoint that will not
take the probe will not take a save either. It resolves no projection, so a broken Mapping cannot
read as a store that is gone. A page save never probes — that path does not initialize; the probe
fires once per run start, background batch, wholly failed batch, and update.php — on QLever at the
measured 591k-triple store size, ~0.2s per probe against batches costing minutes.

The plugin contract now obliges a backend with nothing to create to still reach its store here,
with the RedHerb example noting why its in-process store conforms as a no-op.
The update.php failure report no longer claims subject editing and display fail — false for a SPARQL
store, which can now trigger it: only the backend's own reads fail, and edits made meanwhile are
missing from its projection until a rebuild.

Verification with QLever stopped (rebuild scoped to its `native` store, 121 pages): master walks all
121 pages in 3m57s and reports `succeeded`; this branch ends in 2.7s with
`The SPARQL update to <http://qlever:7019/> failed with HTTP status 0` and a `--resume` hint. A
healthy rebuild is unchanged (121/121, same triples as master). Four tests pin the behaviour: the
probe update string, that `initialize()` resolves no projection, that a rebuild through a store built
from `$wgNeoWikiSparqlStores` ends before walking the wiki (seen failing on master), and that every
live engine the end-to-end suite runs against accepts the probe (QLever and Oxigraph; Fuseki checked
by hand). Unauthenticated, QLever answers `500`, so a credential failure surfaces as a store failure
too.

Considered, omitted:

* The issue's two alternatives — a dedicated liveness method on the plugin contract, or the rebuild
  probing SPARQL stores with an update of its own. Both add surface for what `initialize()` already
  promises.
* Letting `CompositeGraphDatabasePlugin::initialize()` carry on past a failing plugin: on update.php
  an unreachable SPARQL store now stops the plugins queued behind it, as an unreachable Neo4j already
  does, and Neo4j initializes first, so its constraints are never the casualty.
* Restating the obligation in docs/extending/extending.md — the contract's home is the
  `GraphDatabasePlugin` docblock, and the guide already notes the rebuild asks `initialize()`
  whether the store is still there.
* #1259, the misreported sync state visible in the master run above.

> <sub>AI-authored — Claude Code, `Opus 5 (max)` implementation and review fixes, `Fable 5 (max)` orchestration and final pass; from @JeroenDeDauw's in-session delegation of the pre-verified issue, one adversarial review round applied; diff not yet human-reviewed; new tests seen failing first (the rebuild-level one against master's production code), full PHPUnit suite plus phpcs and phpstan green locally, CI green, probe accepted by live QLever and Oxigraph (Fuseki by hand), dead-store and healthy rebuilds checked live on the dev stack.</sub>

### DIFF
--- a/src/Domain/GraphDatabase/GraphDatabasePlugin.php
+++ b/src/Domain/GraphDatabase/GraphDatabasePlugin.php
@@ -33,6 +33,8 @@ interface GraphDatabasePlugin {
 	 * run it every time. update.php calls this so an ordinary install or upgrade establishes those
 	 * structures, and a rebuild calls it on the store it is scoped to before re-projecting into it, so a
 	 * rebuilt graph carries them; the incremental per-edit path does not.
+	 *
+	 * A backend with no such structures still has to reach its store here and throw when it cannot.
 	 */
 	public function initialize(): void;
 
--- a/src/Domain/Rdf/RdfNamespaces.php
+++ b/src/Domain/Rdf/RdfNamespaces.php
@@ -110,6 +110,15 @@ readonly class RdfNamespaces {
 		return new Iri( $this->baseUri . '/graph/' . self::localName( $projection ) . '/page/' . $id->id );
 	}
 
+	/**
+	 * A named graph reserved for probing whether a store can be reached, so an update naming it is
+	 * harmless whatever the store holds: nothing writes to it. No page graph can collide with it,
+	 * whatever a projection is named, since every IRI that {@see graph()} mints ends in `/page/{id}`.
+	 */
+	public function probeGraph(): Iri {
+		return new Iri( $this->baseUri . '/graph/liveness-probe' );
+	}
+
 	public function term( string $localName ): Iri {
 		return new Iri( $this->baseUri . '/ontology/' . $localName );
 	}
--- a/src/EntryPoints/NeoWikiHooks.php
+++ b/src/EntryPoints/NeoWikiHooks.php
@@ -233,17 +233,17 @@ class NeoWikiHooks {
 		$updater->output(
 			"failed.\n"
 			. '...' . $reason . "\n"
-			. "...Re-run update.php once the cause is resolved. While a backend is unreachable, Subject\n"
-			. "...editing and display and every graph read fail, and the edits made meanwhile are missing\n"
-			. "...from the projection: run RebuildGraphDatabases.php to reconcile it.\n"
+			. "...Re-run update.php once the cause is resolved. The reads the backend serves fail while\n"
+			. "...it is unreachable, and the edits made meanwhile are missing from its projection: run\n"
+			. "...RebuildGraphDatabases.php to reconcile it.\n"
 		);
 
 		// Logged as well, because update.php --quiet discards everything written to the updater, which
 		// would otherwise leave a failed initialization with no trace anywhere. The redacted reason
 		// rather than the exception, so that what is kept out of the terminal stays out of the log too.
 		LoggerFactory::getInstance( 'NeoWiki' )->error(
-			'NeoWiki failed to initialize its graph databases during update.php. The wiki is updated, but '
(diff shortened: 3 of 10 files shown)


docs/extending/extending.md
```diff
@@ -170,7 +170,9 @@ save. `deletePage` gets only the page id.
 `initialize` runs on `update.php`, at the start of a `RebuildGraphDatabases` run of your store before any page is
 projected, and once per batch of a rebuild started from the wiki — create the store-level structures a fresh store
 needs there. Make it idempotent and cheap, since every path calls it every time; it never runs on an individual edit.
-A rebuild also calls it to ask whether your store is still there when a whole batch of pages has failed.
+A rebuild also calls it to ask whether your store is still there when a whole batch of pages has failed. If your
+backend has no such structures, still reach the store here and throw when you cannot: that throw is the only way a
+rebuild can distinguish a store that is gone from pages it will not take.
```
JeroenDeDauw: oh dear lord

JeroenDeDauw:
LGTM; did not do deep dive

PR #1289, opened 2026-08-13
Show edit notices in the Subject editing interfaces

### LINKED ISSUE #1262: Allow for message injection into edit interfaces
We make use of approval extensions (Extension:ContentStabilization). In cases, where there is an approved (staibilized) version of a page, when the user performs an edit, the editor interface will show a hint about the implications

Example:

Image

We found users to be confused when editing a NeoWiki subject but then not seeing their changes right away. Therefore we'd like to be able to inject a message into the various places a user can edit a subject.

alistair3149:
Fixes https://github.com/ProfessionalWiki/NeoWiki/issues/1262

Shows a message before a user edits a Subject, so an approval extension can warn that a saved change stays hidden until it is reviewed.

Two producers feed one ordered list. Wiki admins write notices as interface messages, following the convention core uses for its own edit notices:

```
MediaWiki:Neowiki-editnotice-{ns}
MediaWiki:Neowiki-editnotice-{ns}-{dbkey}
MediaWiki:Neowiki-editnotice-schema-{Schema}
```

Extensions register a provider through `NeoWikiRegistrar::addSubjectEditNoticeProvider()`. Exact key rules and the provider contract are in [`docs/authoring/edit-notices.md`](docs/authoring/edit-notices.md) and [`docs/extending/extending.md`](docs/extending/extending.md#edit-notices); the endpoint is in [`docs/api/rest-api.md`](docs/api/rest-api.md).

`GET /neowiki/v0/page/{pageId}/editNotices?schema={Name}` serves both. The editors fetch when they open rather than receiving notices with the page, because a notice can depend on the viewer and on state that changes without an edit. A denied read returns a body byte-identical to a page with no notices, so the endpoint cannot be used to probe page readability.

Core's `Title::getEditNotices()` is not reused. Those notices are authored for wikitext editing and would be noise in a Subject dialog, and the hook behind them receives only a title, so it cannot express Schema scope. An extension already implementing `TitleGetEditNotices` therefore needs a second handler.

A notice brings its own presentation; NeoWiki adds none.

Scope is the Subject editor, the Subject creator, and Manage subjects. The Schema, Layout and Mapping editors are not covered.

Known limitation, documented in `docs/authoring/edit-notices.md`: notices follow the page being viewed, so a Subject rendered elsewhere with `{{#view: <subjectId>}}` shows the viewing page's notices rather than its own.

Tests: unit tests cover key resolution across both of core's namespace branches, ordering, disabling with `-`, a message that parses to empty, and that a stale fetch cannot overwrite a newer one. Integration tests cover the three guards against real interface messages, and the endpoint's read gate by asserting a denied page returns exactly what a page with no notices returns.

## Manual Browser Check

1. Create `MediaWiki:Neowiki-editnotice-0` with any text, and `MediaWiki:Neowiki-editnotice-schema-<Schema>` for a Schema used on a page with a Subject.
2. Open that page and click the edit button on the infobox. Both notices appear above the fields, namespace first.
3. Open "Create subject" on the same page. Only the namespace notice appears, until a Schema is chosen.
4. Set `MediaWiki:Neowiki-editnotice-0` to `-`. It disappears from both.
5. Put `{{PAGENAME}}` in the page-scoped notice `MediaWiki:Neowiki-editnotice-0-<PageDbKey>`. It resolves to the page being edited.

To check that an admin's wikitext is sanitized, put `<script>alert(1)</script>`, `` and `<div onclick="alert(3)">x</div>` in a notice. All three render as inert text or lose the handler; none execute. Provider HTML is deliberately not sanitized, and is documented as the extension's responsibility.

## Considered, omitted

- A severity per notice: no producer in scope needs more than one level, and `Domain\Validation\Severity` already exists if that changes.
- Title-keyed rather than pageId-keyed routing: only needed once the Schema, Layout and Mapping editors are covered, and those save by title.

<sub>AI-authored — Claude Code, `Opus 5 1M (xhigh)`; design steered by @alistair3149 across several rounds, including two reversals of the approach; full diff not yet human-reviewed; PHP and TypeScript suites, phpcs, phpstan and lint green, endpoint and rendering verified in a browser.</sub>

### DIFF
--- a/docs/api/rest-api.md
+++ b/docs/api/rest-api.md
@@ -15,7 +15,8 @@ Every endpoint is also published as a complete [OpenAPI 3.0 description](#full-s
 
 ## Permissions
 
-The Subject, page-subjects, subject-labels, Schema, Layout, Mapping, RDF export, and entity-dereference read endpoints
+The Subject, page-subjects, edit-notices, subject-labels, Schema, Layout, Mapping, RDF export, and entity-dereference
+read endpoints
 enforce the caller's per-page `read` permission; page protection and `$wgNamespaceProtection` do not restrict them,
 because MediaWiki's `read` action ignores both. When you may not read a page they respond as if the data were absent — a
 `null` value, an empty list, or a `404` — never a `403`. `GET /subject-labels` omits the labels of Subjects whose page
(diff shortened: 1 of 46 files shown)


alistair3149:
image

JeroenDeDauw:
@alistair3149 did you already review the docs and ensure there is nothing out of place in there?

alistair3149:
Yes just did another pass on it.

PR #1290, opened 2026-08-13
Give StringValue a canonical form without empty parts

### LINKED ISSUE #1209: subject-format.md promises empty values are dropped without error; string types disagree for ""
Found while reviewing #1199, independent of that PR and of the severity work.

## The documented contract

[`docs/api/subject-format.md`](../blob/master/docs/api/subject-format.md) states:

> A statement entry without `propertyType`, or whose value is empty for its type, is dropped without error.

## What actually happens

Only the *no parts* case is dropped. `StatementListBuilder` drops a Statement when `NeoValue::isEmpty()`, and `StringValue::isEmpty()` is `$this->strings === []`. Deserialization does `new StringValue( ...(array)$value )` with no filtering, so a JSON `""` arrives as `[ '' ]` — not empty by that definition — and reaches validation.

What validation then does with it varies by type:

- `text`, `url`, `date`, `dateTime` — treat it as absent and raise `required`/warning if the property is required, matching the documented spirit though not the letter (a violation is reported, not "dropped without error").
- RedHerb `color` — treats it as malformed and raises a blocking `invalid-color`. Tracked separately in #1208.

So a caller following the documentation and sending `""` to clear a field gets a violation rather than a silent drop.

## Why it has not bitten

No shipped client sends `""`: `newStringValue` (`resources/ext.neowiki/src/domain/Value.ts`) trims and filters empty parts, so a cleared field serialises to `null` and the statement is dropped client-side. This only affects hand-written REST clients and import paths — which are exactly the readers the API docs are written for.

## Decision needed

Either:

1. **Make the code match the doc** — normalize empty and whitespace-only string parts at the deserialization boundary, so `""` is dropped server-side the way the client already drops it. This also resolves #1208 for every string type at once, rather than per type. Needs a check that no type legitimately treats `""` as a meaningful value.
2. **Make the doc match the code** — state precisely which shapes count as "empty for its type" (`[]` / `null` / omitted, but not `""`), and note that an empty string part is validated rather than dropped.

Option 1 is the more useful contract and removes a class of per-type inconsistency; option 2 is safer if any consumer relies on `""` round-tripping.

JeroenDeDauw:
Fixes https://github.com/ProfessionalWiki/NeoWiki/issues/1209

`docs/api/subject-format.md` promises that a Statement whose value is empty for its type is dropped without error.
A JSON `""` (or `[""]`, `[" "]`) instead became `StringValue([''])` and reached validation and persistence, storing
a junk empty part in the revision slot and the graph projections.

`StringValue` now has a canonical form: its constructor drops parts that trim to empty, and parts with content are
stored as written — an "empty but has parts" state can no longer exist. A value built from `""` has no parts, so
the builder's existing emptiness gate drops the Statement on every endpoint that takes `statements`.

The canonicalization applies wherever the value is constructed — the request path and the stored-slot path alike —
so the two sides of the violation diff see the same part positions: a pre-existing violation on stored data cannot
newly block a write under enforcement (regression test: a stored `['', 'not a url']` echoed back with enforcement
on). Validators never see empty parts, from requests or from storage.

Two further behavior changes:

* `SelectStatementResolver` treated `""` as an unknown option and answered 400 while the validate endpoints left it
  in place, so the dry run said clean and the save failed. Clearing a select with `""` now behaves like every other
  string type; writes that previously answered 400 succeed.
* Values stored with empty or whitespace-only parts read back without them: they stop rendering in the
  parser-function and Lua read paths, REST reads return the canonical parts, and the next write persists them.

Out of scope: empty number, boolean and relation values still answer 500
(https://github.com/ProfessionalWiki/NeoWiki/issues/1291), and parts with content are never trimmed.
https://github.com/ProfessionalWiki/NeoWiki/issues/1208 narrows to its padded-value point — the empty-part state
it describes is no longer representable. Docs are unchanged: for string values the documented sentence now holds
as written.

> <sub>AI-authored — Claude Code, `Opus 5 (max)`; open question + go-ahead from @JeroenDeDauw, design redirected once by him mid-review; diff not yet human-reviewed; new tests mutation-tested, full PHPUnit and phpcs/phpstan green locally, CI green.</sub>

Design by `Fable 5 (max)` and @JeroenDeDauw; implementation by `Opus 5 (max)` subagents. An independent AI review
pass verified behavior live over REST mid-development.

</details>

### DIFF
--- a/src/Application/SelectStatementResolver.php
+++ b/src/Application/SelectStatementResolver.php
@@ -96,6 +96,13 @@ readonly class SelectStatementResolver {
 	}
 
 	private function resolveSingle( string $propertyName, SelectProperty $property, mixed $value ): string {
+		// An empty value is a cleared field, not a reference to an option. Left in place for
+		// StatementListBuilder to drop, the way the validate paths already treat it, rather than
+		// answering 400 to a caller clearing a select.
+		if ( is_string( $value ) && trim( $value ) === '' ) {
+			return $value;
+		}
+
 		if ( !is_string( $value ) && !is_array( $value ) ) {
 			throw new InvalidArgumentException(
 				"Select value for \"{$propertyName}\" must be a string or object"
--- a/src/Domain/Value/StringValue.php
+++ b/src/Domain/Value/StringValue.php
@@ -11,10 +11,6 @@ readonly class StringValue implements NeoValue {
 	 */
 	public array $strings;
 
-	/**
-	 * Canonical form: a part without content does not exist, and a part with content is stored as
-	 * it was written.
-	 */
 	public function __construct( string ...$strings ) {
 		$this->strings = array_values(
 			array_filter( $strings, static fn ( string $part ): bool => trim( $part ) !== '' )
--- a/src/Domain/Value/StringValue.php
+++ b/src/Domain/Value/StringValue.php
@@ -11,8 +11,14 @@ readonly class StringValue implements NeoValue {
 	 */
 	public array $strings;
 
+	/**
+	 * Canonical form: a part without content does not exist, and a part with content is stored as
+	 * it was written.
+	 */
 	public function __construct( string ...$strings ) {
-		$this->strings = $strings;
+		$this->strings = array_values(
+			array_filter( $strings, static fn ( string $part ): bool => trim( $part ) !== '' )
+		);
 	}
 
 	public function getType(): ValueType {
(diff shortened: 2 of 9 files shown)


JeroenDeDauw:
Solution looked wrong to me. Fable when asked said it's the right thing/goal. Had to spell out my alternative. So still not fully reliable for small design decisions (with big impacts in this case)

PR #1299, opened 2026-08-15
Stop silently discarding unparseable text in number fields

JeroenDeDauw:
Typing `5foo` into a number field in Firefox (or `5e` in Chromium) leaves
the characters visible on screen while the DOM reports `value === ''` —
a native `<input type="number">` never exposes text it cannot parse.
NumberInput therefore emitted `undefined`, the statement was dropped from
the model, and saving persisted the subject without it: the text the user
was looking at vanished with no error shown.

While the widget reports `validity.badInput`, the field now shows the
existing `neowiki-field-invalid-number` message with error status, and the
subject editor and creator hold the save — the handler returns early with
an error notification and the dialog stays open with the field flagged.
This is widget-state handling, not frontend constraint validation: the
backend never receives this text, and constraint violations, warning and
error alike, still never block a save client-side.

The text is not stored: a statement on a number property whose value is
not a number is rejected with a 400 by design, so [ADR 21][adr21]'s
"invalid parts should not disappear on save" is honoured at the widget
boundary instead, by refusing to save what cannot be represented.

`ValueInputExposes` gains an optional `hasUnparseableInput()`, which
`SubjectEditor` aggregates across its fields; only NumberInput implements
it. Every subject save in the UI goes through one of the four gated
dialogs (subject editor, subject creator, and RedHerb's two examples).
Known gaps of the same class, left as follow-ups: RelationInput discards
unmatched text by a different route, and the schema editor's initial-value
field shows the new message but still saves with the default dropped.

Considered, omitted: a reactive disabled save button (new emit plumbing
for no added clarity), deferring the message to blur (it flashes on a
leading minus while typing a negative number; a touched ref is more
structure than the papercut warrants), and badInput handling for the date
and date-time inputs, whose segmented pickers make the state hard to reach
by typing.

[adr21]: https://github.com/ProfessionalWiki/NeoWiki/blob/master/docs/adr/021-add-backend-validation.md

> <sub>AI-authored — Claude Code, `Opus 5 (max)`; detailed spec from @JeroenDeDauw, design agreed up front; diff not yet human-reviewed; vitest suite (badInput stubbed — jsdom cannot produce it), TS build and lint local plus CI green, behaviour verified in a dev wiki, independent AI review pass applied.</sub>

Spec, steering and final review by `Fable 5 (max)`; implementation and review-fix application by `Opus 5 (max)` subagents. An independent AI review of the first version produced the flag-lifecycle guard, the save-time notification, the exported `SubjectEditorExposes` interface, and three of the tests; all applied.

</details>


JeroenDeDauw:
image
Tested locally
image

JeroenDeDauw:
LGTM

PR #1301, opened 2026-08-17
Add a per-Constraint severity control to the Schema editor

### LINKED ISSUE #1149: Add a per-Constraint severity control to the Schema editor
[ADR 26](https://github.com/ProfessionalWiki/NeoWiki/blob/master/docs/adr/026-validation-severity-levels.md) lets a schema author mark each Constraint `error` or `warning` (default `warning`), which is what decides whether violating it can block a write under `$wgNeoWikiEnforceValidation`. The backend landed in #1147, but there is no authoring UI: setting `error` currently requires hand-editing the Schema JSON.

Add a severity control beside each authorable Constraint. A shared `SeveritySelect` component, wired through the existing `emit( 'update:property', … )` path, keeps every site consistent — the `Partial<T>` emit contract already carries it unchanged.

## Sites

| Constraint | Component |
|---|---|
| `required` | `components/SchemaEditor/PropertyDefinitionEditor.vue` (the only one in the container rather than an attributes editor) |
| `minimum` / `maximum` | `Property/NumberAttributesEditor.vue`, `Property/DateAttributesEditor.vue`, `Property/DateTimeAttributesEditor.vue` |
| `minLength` / `maxLength`, `uniqueItems` | `Property/TextAttributesEditor.vue` |
| `uniqueItems` | `Property/UrlAttributesEditor.vue` |
| `options` | `Property/SelectAttributesEditor.vue` |

Deliberately **not** severity-bearing, so they get no control: Display Attributes (`precision`) and the shape-declaring fields (`multiple`, `relation`, `targetSchema`). The backend discards a severity on the former and rejects it on the latter, so offering a control would be misleading.

## Open design question

ADR 26 treats `min-value` and `max-value` (and `min-length` / `max-length`) as separate Constraints with independently configurable severities, but the editor groups each pair in one `NeoNestedField`. Either give each input its own control (faithful to the model, more controls), or one control per range group (fewer controls, but needs a rule for serializing back to two separate JSON keys). Worth deciding before building.

## Dependency

Depends on #1148: the editor must parse and re-serialize the object Constraint form before it can round-trip an authored severity.

alistair3149:
Fixes https://github.com/ProfessionalWiki/NeoWiki/issues/1149. Follows-up to https://github.com/ProfessionalWiki/NeoWiki/pull/1227, which made the editor round-trip the object Constraint form.

Adds a per-Constraint severity control to the Schema editor, so marking a Constraint `error` ([ADR 26](https://github.com/ProfessionalWiki/NeoWiki/blob/master/docs/adr/026-validation-severity-levels.md)) no longer requires hand-editing the JSON.

## The control

* `SeverityInput.vue`: a `CdxMenuButton` whose face is the current level's icon (grey `cdxIconAlert` for warning, red `cdxIconError` for error) at `size="small"` plus an x-small chevron. The menu offers Warning and Error, each with its icon in the Codex warning/error colour and a one-line meaning; the Error line depends on `$wgNeoWikiEnforceValidation` ("Flags the value as an error and blocks saving" / "Flags the value as an error"), which `FrontendModuleLoader` now exposes as `wgNeoWikiEnforceValidation`. Accessible name and tooltip name the Constraint too, since several controls can sit on one property: "Severity of "Minimum": Warning".
* Placement: the end of the Constraint's label row (label at the start, input untouched below), or the end of the checkbox row for boolean Constraints. The trigger keeps Codex's 32px hit area but takes 20px in the flow through negative block margins, so label rows and checkbox rows keep their height. The two placement helper classes (`ext-neowiki-severity-field`, `ext-neowiki-severity-row`) live with the component.
* Shown only once the Constraint is set: `required` checked, a bound with a value, `uniqueItems` checked, at least one option. Unticking `required` or `uniqueItems`, or removing the last option, drops the severity with it; a cleared bound keeps its severity until save (a number input reports interim text such as `1.` as empty, so pruning there would lose the choice mid-typing), and serialization drops the severity of any Constraint still absent. Changing the Property Type keeps `required` with its severity and drops the type-specific ones.
* Sites: `required` (`PropertyDefinitionEditor`), `minimum` / `maximum` (Number, Date, DateTime), `minLength` / `maxLength` / `uniqueItems` (Text), `uniqueItems` (Url), `options` (Select). Display Attributes (`precision`) and shape fields (`multiple`, `relation`, `targetSchema`) get none, matching what the backend accepts.

## Wiring

* `withConstraintSeverity( property, name, severity )` returns the `Partial` an attributes editor emits: it sets the annotation, removes it for warning (the default), and drops an emptied map, so unannotated definitions stay identical to factory-built ones. `withoutSeveritiesOfClearedConstraints` prunes the annotations of Constraints a change unticks or empties; `PropertyDefinitionEditor` applies it to every attribute change.
* `NeoNestedField` gap 8px → 16px so a bound's control does not crowd the next bound's label.
* New messages, `cdxIconError` in the icon allowlist, and `SeverityInput` exported through the public API for extension attributes editors.
* Tests: `SeverityInput`, `withConstraintSeverity` and `withoutSeveritiesOfClearedConstraints` get their own specs; each site's spec covers reveal-once-set and the emitted `constraintSeverities`; `PropertyDefinitionEditor` covers unsetting and the type change; `FrontendModuleLoaderTest` and `NeoWikiExtension.spec.ts` cover the new config var.

## Considered, omitted

* A boxed `CdxSelect` per Constraint: four boxed fields on a fully constrained text property.
* A collapsed per-property "Severity" section: the author would pick a severity without the rule's value in view.
* Toggle buttons: the glyph could not change with the level, and Codex's pressed state reads as blue rather than error.
* A help sentence on the form: the menu descriptions carry the meaning at the moment of choice.

## Manual Browser Check

1. On a dev wiki with demo data, open `Schema:Validation Demo` and click the edit pencil.
2. Select *Score*: each bound's label row ends with a small grey ⚠▾ and the inputs keep full width. Open the *Maximum* one: a menu with Warning (yellow) and Error (red), each with a one-line meaning. Pick Error — the trigger turns red and its tooltip reads "Severity: Error".
3. Select *Title*: the trigger ends the "Require a value" row. Untick the box and it disappears; tick it and it is back.
4. Select *Status*: the trigger ends the "Options" label row; the chip input keeps full width.
5. Save. The stored JSON has `"maximum": { "value": 100, "severity": "error" }` while the untouched Constraints stay bare. Reopen the editor: the trigger shows Error again.
6. With `$wgNeoWikiEnforceValidation = true` in `LocalSettings.php`, the Error item reads "Flags the value as an error and blocks saving"; without it, "Flags the value as an error".

<sub>AI-authored — Claude Code, Fable 5; design settled with Alistair, then implemented; verified with vitest (124 files, 1351 tests), PHPUnit for the loader, vue-tsc/vite build, eslint/stylelint/phpcs/phpstan, mutation-tested new specs, an adversarial code review whose confirmed findings are addressed, and a browser round trip on a dev wiki in both enforcement settings.</sub>


alistair3149:
image

^ The hover state background is clipping a bit. I left it as it is intentionally to preserve the 32px clickable height, the hover affordance, and the compact height of the label without introducing more spacing.

JeroenDeDauw:
"Require a value" only shows the severity option when its checked, which makes some sense since violations are not possible if it isn't checked.

Allow multiple values OTOH seems to never show the severity selector. While you can violate it (but only when it is NOT checked). Perhaps difficult to create the violation via the UIs, but possible via APIs, and possible when the schema changes (from allowing multiple to not allowing it).

image

Edit: this PR is in line with ADR 26, though the ADR seems to not be making the optimal choice wrt the single-value constaraint
* https://github.com/ProfessionalWiki/NeoWiki/pull/1315

JeroenDeDauw:
This text seems poor. "Flags the value as a warning". Should presumably be something like "Flag constraint violations as warnings". Both the grammar change ("Flags" reads claudish and is just weird) and the sentences subject: constaint violations, not value. And remember: we need to be consistent with the ubiquitous language, probably worth doing a dedicated pass for that over these new UIs/messages.

image

JeroenDeDauw:
There is something to only showing the severity controls when violations are actually possible (e.g., hiding when "require a value" is false). That said there are also downsides, and I wonder if it's net-better to always show.

If we do hide them, we should avoid the current behavior of resetting by cycling between the hidden and visible states of the UI as seen here:
[Screencast_20260823_004613.webm](https://github.com/user-attachments/assets/d1c7896a-937f-4e8b-b1de-f010f08da7b5)

JeroenDeDauw:
Unsure if added by this pr, but there is some bad behavior when switching the property type. Here you can see the text has min and max lengths set. When I change to URL type and then back to text type, these min and max values vanish.
[Screencast_20260823_004940.webm](https://github.com/user-attachments/assets/baf4d58b-7e7d-4887-ab89-a883bd0b4632)

PR #1305, opened 2026-08-18
Move the Schema description into the schema editor dialog header

alistair3149:
The description was edited through a text area in a band pinned above the
property list, costing 140px of a dialog capped at 859px whether the
description was 409 characters or empty, and leaving the property list and
the property editor 515px each. It also could not show what it held: the
tallest description on the demo wiki needs 88px of text in a 62px window.

It is now the Schema's subtitle in the dialog header, clamped to two lines,
which frees the whole band. The header grows by the line or two the
description takes, so the two panes that carry the work gain between 97px
and 117px. The band is deleted rather than slimmed, so the body opens
directly onto the property list.

EditableText grows three options to carry this. `multiline` edits in an
autosizing text area, committing on Ctrl or Cmd with Enter since Enter
inserts a newline; five of the twenty-one demo Schemas exceed one line.
`clampLines` limits the display and moves the edit button onto the last line
rather than the far right of a wrapped block, joined by a reveal button over
a fade once the text overflows. `addLabel` replaces both with a link while
the value is empty.

The clamping lives in a useClampedText composable, next to the existing
useOverflowDetection. Two mechanisms in it are load-bearing. The clamp is a
measured max-height rather than `-webkit-line-clamp`, because that needs
`display: -webkit-box`, which lays element children out as their own
vertical boxes and so cannot hold a control on the last line. Overflow is
measured on the text node through a Range rather than on the element around
it, because the element's scroll height also counts the controls and the
invisible area padding their pointer target, which reports a description
that fits as overflowing. The text box clips only while it is cutting text
off, since clipping otherwise sliced the focus ring of the buttons it holds.

SchemaEditor no longer renders the description, so its hosts present it.
The prop that carries it is optional: SchemaEditor is exported from
public-api, and a caller that does not present a description at all must
keep the one the Schema arrived with rather than clear it. The create dialog
keeps a plain single-line field beside the name, since creating is filling in
a form; that also drops its own text area, taking the chrome above the first
property from 248px to 180px.

SchemaDeserializer now reads an absent description as an empty one. It was
passing `undefined` into a parameter typed `string`, which the old text area
tolerated and the new empty state does not.

## Manual Browser Check

1. Open a Schema whose description runs past two lines (`Schema:Validation Demo`
   on a dev wiki with demo data) and click the edit pencil on the Schema box.
   The description sits under the dialog title, cut off at two lines, fading
   into an ellipsis and a pencil at the end of the second line.
2. Click the ellipsis: the description expands in full and the button becomes a
   collapse. Click it again to re-collapse.
3. Click the pencil: a text area opens with the whole description. Enter must
   insert a newline rather than save; Ctrl+Enter commits and enables Save.
   Reopen it and press Escape, which must discard the draft and leave the
   dialog open.
4. Narrow the window while the dialog is open: the clamp and the fade follow the
   new width.
5. Open a Schema whose description fits in two lines (`Schema:Person`). The
   pencil trails the last word instead, with no ellipsis to reveal, and tabbing
   to it shows a focus ring on all four sides.
6. Open a Schema with no description at all (`Schema:Company`). The header
   offers an "Add description" link in place of the text and the pencil, and
   it opens the same editor.
7. On `Special:Schemas`, press Create. Name and Description sit in one band as
   single-line fields, with no text area.

<sub>AI-authored — Claude Code, Opus 5; designed against measurements taken from the running dialog, then implemented and revised against a code review; verified with the full vitest suite (1341 tests), vue-tsc/vite build, eslint and stylelint, and by walking all seven steps above on a dev wiki, including the empty, two-line and overflowing descriptions, a resize, and the create dialog.</sub>


alistair3149:
image

PR #1309, opened 2026-08-19
Make the Subject label optional

### LINKED ISSUE #1283: Make the subject label optional, with a computed display-name fallback
Follows up on #1140: after #1281 (labels editable in place) and #1282 (schema-name defaults for child subjects), the remaining design question is whether a Subject should require a label at all.

## Problem

The label is required, so every subject stores one even when nobody chose it. #1140 already describes the first consequence: subjects that exist as page metadata have no use for a label, yet must store one. The planned import capabilities make this acute: imported subjects (event and qualifier subjects especially) often have no natural name, so a required label means the importer synthesizes one, and the synthesized value persists as if a curator wrote it.

Stored defaults also cannot be told apart from chosen labels, which blocks the obvious fixes. Verified on a dev wiki: moving a page updates the Page node's `name` in the graph, while the main subject keeps its stored page-name label — and auto-syncing it is unsafe precisely because we cannot tell whether the label was customized.

## Proposal

1. **Label becomes optional.** Drop the `label-required` violation and the non-empty guards in the create and replace endpoints; the slot JSON stores no label when none was given. An absent label means "nobody named this subject".
2. **Display falls back at the point of use.** Surfaces that need a name apply the existing shared rule (`defaultSubjectLabel`: Main Subject → page name, Child Subject → Schema name) with context they already hold. The subject editor's title and the creator's field show the computed name as a placeholder, not as a stored value — the `EditableText` component from #1281 already supports this.
3. **The graph materializes the fallback only for Main Subjects.** The projector holds the page when it writes, and freshness is free: page moves already reproject through the null-revision → `RevisionFromEditComplete` path (verified round-trip on a dev wiki). Label-less Child Subjects get no `name` property — their fallback value is already on the node as its Schema label, so queries can derive it.
4. **RDF emits `rdfs:label` from the effective name for every subject.** RDF consumers key on `rdfs:label`, and there the Schema appears as `rdf:type`, not as a label, so the property-graph argument for omission does not carry over. The Child Subject value is the Schema name, which is immutable (ADR 17), so this materialization cannot go stale.
5. **Label search sees only stored labels.** Relation-target autocomplete and label search will not find label-less Child Subjects: a subject nobody named cannot be found by name, and it stays reachable through its page and through queries by Schema or Statement values. If this bites in practice, the escape hatch is one line — materialize the Child fallback into `name` too — and costs nothing to take later.
6. **Existing stored defaults are nulled out once,** followed by a per-store rebuild. This is load-bearing, not optional polish: the stale-label fix only applies to subjects without a stored label, so pre-existing subjects keep the bug until their unedited page-name defaults are removed. Cheap while we are pre-production; the heuristic (stored label exactly equals the computed default) can only misfire on a curator deliberately typing the default, which loses nothing.
7. **Read APIs return both** the stored label (nullable) and the effective display name, so the editor can distinguish placeholder from value and API consumers need not reimplement the rule.
8. **No synchronization framework.** If a consumer someday needs guaranteed-fresh Child Subject names in the graph, that is the moment to pay for it.

## What this does not fix

Several label-less Child Subjects of the same Schema display identically. The fallback computation can add discriminators later without touching stored data; stored defaults could not.

## Open questions

- API field naming for the effective display name.
- Lua: should `subject.label` return the effective name (leaning yes) or the stored, nullable one?
- Consumer coordination: relaxing the write endpoints is backward-compatible for writers that keep sending a label; the actually breaking surfaces are the read APIs (`label` becomes nullable, a new field appears) and the graph shape (Child Subjects lose `name`).

A possible future relation between subjects on the same page does not gate this: the fallback rule works under any resolution of that question, and such a relation could later refine the Child Subject fallback.

> <sub>AI-authored — Claude Code, `Fable 5 (xhigh)`; design worked out in discussion with @alistair3149, who narrowed the scope once from a materialize-everywhere draft; artifact version reviewed by @alistair3149, this generalized public revision not re-reviewed; page-move reprojection and label staleness verified empirically on a dev wiki.</sub>

alistair3149:
Fixes https://github.com/ProfessionalWiki/NeoWiki/issues/1283

A Subject may now have no label. The slot omits the key, and a blank label is read as absence, so the two cannot drift
apart. Display falls back at the point of use — a Main Subject to its page name, every other Subject to its Schema
name — held in one rule, `SubjectDisplayName`, mirrored for the creator by `placeholderSubjectLabel.ts`.

The requirement was enforced in three independent layers, all dropped together: the `label-required` violation, hard
guards in the create and replace Actions, and `PARAM_REQUIRED` on four endpoints. Relaxing fewer would have been
invisible from the API.

Decisions worth a look:

- **The graph materializes the fallback for Main Subjects only.** The Schema name there would make every unnamed
  Subject of a Schema indistinguishable in query results, and the Schema is already the node's second label. An absent
  `name` is then no longer the mark of a stub, so the orphan sweep now identifies one by its node labels — the old
  oracle would have deleted real Subjects.
- **`rdfs:label` always falls back; a Mapping's `labelPredicate` does not.** That term states what the thing is
  called in the target ontology, and a Schema name under `foaf:name` would assert a type as a name, which
  reconciliation keys on.
- **Lua diverges from REST on purpose.** `subject.label` is the display name and `subject.storedLabel` the nullable
  stored one, because an absent key is `nil` in Lua and concatenating it replaces the rendered page with a script
  error. REST splits them the other way, with a nullable `label` and a non-null `displayName`.
- **PUT is a full replacement**, so an omitted `label` clears a stored one.

`NeoWiki:ClearDefaultSubjectLabels` clears pre-existing stored defaults once. A stored default cannot be told from a
chosen label, so a moved page kept Subjects named after its old title. The fix only reaches Subjects that have no
stored label; without the script, existing Subjects keep the bug.

Considered, omitted:

- **Making the display name configurable.** [#1140](https://github.com/ProfessionalWiki/NeoWiki/issues/1140) is
  titled "configurable / extendable", but its body asks for an outcome — stop showing the page name, show the Schema
  name — which this delivers. The consumer with the strongest claim renders its own subject column from the REST API
  and now receives `label`, `displayName` and `schema`, so it needs no server-side hook. If configurability does
  arrive it is most likely Schema-authored data, for the identical-siblings case, which wants more arguments to a
  pure function rather than dependency injection.
- **Discriminating several label-less Child Subjects of one Schema.** They still display identically. The fallback
  can gain discriminators later without touching stored data, which stored defaults could not.

Known gaps, not fixed here: `?oldid=` views resolve main-ness from the live page, so a historical label-less Subject
can be named from today's state; and `GET /subject-labels` returns a computed page name in its `label` field for a
label-less Main Subject without saying so.

## Manual Browser Check

1. On a page with no Subjects, open "Create subject" and pick a schema: the label field is marked "(optional)" and
   shows the page name greyed as a **placeholder**, with the field itself empty. On a page that already has a Main
   Subject, it shows the schema name instead.
2. Save without typing a label. The new Subject appears under that name, and `action=raw&slot=neo` shows its entry
   with no `label` key.
3. Open that Subject in the editor: the title shows the fallback greyed, and the rename field shows it as its
   placeholder. Edit only a statement and save: the slot still has no `label` key. Then set a label and save, and
   clear it and save again: the key appears and disappears, and the infobox title follows.
4. `GET /rest.php/neowiki/v0/subject/<id>` returns `"label": null` beside a non-null `"displayName"`.
5. Relation autocomplete still finds Subjects by name; a label-less Child Subject is absent from it by design.

> <sub>AI-authored — Claude Code, `Opus 5 (xhigh)` and `Fable 5`; design authored by @alistair3149 in #1283 and followed as written, with their redirections on ADR non-modification, doc trimming, and the YAGNI call above; implementation and review fanned out across subagents, then a code review, a design review and blind text reviews, of which 18 of 34 findings survived adversarial verification. Verified: phpcs, phpstan, 1954 unit tests, the affected `@group Database` classes individually, vitest, lint and build locally, and the full suites in CI; browser-verified on a dev wiki: the creator placeholders, a blank save storing no label, the editor rename and clear round trip, a statement-only save, the relation picker, the graph name policy, EDM `foaf:name` skipped without a stored label, the migration clearing 77 real labels, and a page move renaming a label-less Subject.</sub>


alistair3149:
https://github.com/user-attachments/assets/7074ce75-53d5-4a9c-94f4-8df63c490538

Issue #971, opened 2026-06-29
Create and edit relation targets while editing a relation


JeroenDeDauw:
## What we want

When you fill a relation, and the target Subject you want doesn't exist yet, you should be able to create it and carry on.

You should also be able to edit an existing target's own fields in place, without changing which Subject the relation points to.

This can continue along a chain: each target is itself a normal Subject that can reference further Subjects through its own relations. A view/layout governs which fields of the referenced Subject are shown while editing.

The goal is low-friction, low-error creation and editing of relation targets during relation editing. The optimal UI is for the implementer to investigate.

The motivation: typed, in-context creation and editing is considered the lowest-error way to hand-edit a knowledge graph: targets are type-correct by construction, and you avoid the orphaned/dangling references that come from breaking off to create them elsewhere.

## References to draw on

- **Freebase / Metaweb** (now dead) is the canonical example: editing a relation ran a real-time search scoped to the target type; if nothing matched the typed label — or always, via an option at the bottom of the list — you could create a new item of the required type right there; and views decided which properties of the related items were shown.
- **[Arches](https://www.archesproject.org/)** is a live, open-source cultural-heritage platform whose form-based editing creates and links graph-structured related resources — worth studying as a current, maintained take on the same idea.

## Data-model note (scope)

NeoWiki has no nested/contained Subjects: only flat Subjects that reference one another via Relation Statements. So this is **editing-layer work** over the existing model; the "depth" above is a UI drill-in along a chain of relations, not nesting in the data.

## Open questions

- **Where do the newly created Subjects live? (the main one.)** Every Subject is stored in some page's subject list, so creating one as a relation target means choosing its page, and none of the options is clearly right:
  - On the **same page** as the referencing Subject (i.e. as a Child Subject — interacts with #959). Easy, but often wrong: the target is frequently its own thing that belongs elsewhere.
  - Let the user **pick a page** — more UI complexity, and the page that makes sense **may not exist yet**, compounding it (create the page in-flow too?).
- Editing an existing target means editing data that lives on another page. How do we make that clear to the user, and how do permissions and save semantics work?
- Where/when a "create new" affordance appears.
- Should there be a limit on how far down the chain a user can go before the UI gets overwhelming?
- Interaction with relation-property schemas (#630).

## Related, not in scope

[...]

PR #1323, opened 2026-08-27
Edit a Subject's related Subjects without leaving the editor

alistair3149:
For https://github.com/ProfessionalWiki/NeoWiki/issues/971

The subject editor gains a tree of the Subjects the edited one references through its relations — outgoing only,
whichever page stores the target, three levels deep — and edits any of them in place. Selecting a row swaps the form,
the form you left stays mounted with what you typed, and one Save writes every Subject you touched. The MVP shape and
what it leaves out are recorded in [the design state on the issue](https://github.com/ProfessionalWiki/NeoWiki/issues/971#issuecomment-5429073355).

Decisions worth a look:

- **The tree is fully expanded, with no collapse control.** The design state says one level with expand on demand;
  this ships without it to keep the MVP small. Rows are per path, so a Subject reached two ways appears twice, by
  design, and a Subject with wide fan-out lists many rows. Collapse is a follow-up.
- **Save writes one Subject per request, root first, and stops at the first refusal.** The refused pane comes on
  screen, what was written stays written, and the footer says how many were.
- **The editor is inert while writing.** The loop harvests every dirty pane up front, so a keystroke during it would
  otherwise be reverted and lost without a word.
- **A target fetch belongs to its opening.** Both hosts keep the dialog mounted after it closes, so a slow fetch
  abandoned by closing would have opened its pane in the next session.
- **`NeoTree` is a generic widget** — one tab stop, arrow keys, no domain knowledge. `SubjectTree` adapts the walk to
  it, and the walk is a pure function that returns what it could not resolve rather than fetching.

Considered, omitted:

- Creating a target Subject mid-edit, the issue's other half — deferred per the design state.
- Tabs beside the tree — the tree already covers switching between open Subjects.
- A responsive layout below roughly 700px — follow-up.
- Renaming a nested Subject: the header's rename control edits the root; the pane has no label control yet.
- A lighter read for the tree: each target fetch returns a full Subject and the tree keeps four fields of it, one
  request per node per level. A relations endpoint is proposed in #1324.

Known gaps, not fixed here: edit notices are loaded for the root's page only; a missing or unreadable target shows its
raw id; a Subject reached by two routes carries `aria-selected` on both rows.

## Manual Browser Check

1. Open a page whose Main Subject has relation targets (demo data: `ACME Inc`) and click the infobox's Edit. The
   dialog has two columns: a "Structure" tree on the left, the form on the right, each scrolling on its own. The
   tree's first row is the root; captions name the relation properties.
2. Click a child row. The form switches to that Subject, a "Stored on <page>" line above it links to its page, and
   the row is highlighted. Click the root row to return: what you typed is still there.
3. In a relation field, the pencil beside a picked target opens that target the same way.
4. Type into the root, then into a child. Both rows carry a dot and the footer reads "Saving updates 2 subjects on
   2 pages". Save: both pages gain a revision and the toast says "Updated 2 subjects".
5. Make a child's save fail (a value that violates an `error`-severity constraint, or its page protected) and save
   with both dirty: the root is written, the failing pane comes on screen, and the footer says "Saved 1 of 2
   subjects".
6. Keyboard: Tab into the tree (one stop), ArrowDown and ArrowUp move between rows, Enter opens a row; Shift+Tab back
   in lands on the row you last chose.
7. Open a Subject with no relations: no tree, the single-column dialog as before.

> <sub>AI-authored — Claude Code, `Fable 5 (xhigh)`, earlier sessions `Opus 5 (1M context)`; design by @alistair3149 over several sessions with frequent redirections (tree only, no collapse, styling passes), then a same-session `/pr-review` whose 4 blocker fixes and cleanups are not yet human-reviewed; earlier UI passes browser-checked by @alistair3149; verified: 1668 vitest tests, lint and build locally, tree, pane and padding behaviour measured on a dev wiki; CI not yet run.</sub>


alistair3149:
image

This is a MVP implementation, there are still quite some follow-ups that needed to be done.

JeroenDeDauw:
image

PR #1325, opened 2026-08-31
Revalidate the Subject editor when its Schema changes

JeroenDeDauw:
Editing a Subject's Schema through the subject editor's header link left
the violations on screen answering the old Schema: with Score capped at
100, a value of 200 kept showing "Maximum value is 100." after the Schema
was saved with a cap of 1000. A Schema edit that introduced a violation
showed nothing at all.

Everything else the Schema decides is derived at render time and followed
the save already. The violations do not: they are a snapshot of a server
response, and a Schema save was not among the things that trigger a fresh
one. The nested schema editor is a sibling dialog, so nothing remounted
the subject editor either. The backend needs no change; it already
validates against the saved Schema on the next call.

Each edit pane's validate-on-mount watcher now also watches the Schema
it validates against, which covers the nested schema editor's save
reaching the root pane and the dialog replacing a pane's Schema.
Non-obvious in it: `flush: 'post'`, since the validator reads the fields
off the editor, which before the re-render still holds the ones the save
replaced.

> <sub>AI-authored — Claude Code, `Opus 5 (max)` implementation, `Fable 5 (max)` diagnosis, spec, review, and the port onto the multi-pane editor; delegated by @JeroenDeDauw; not yet human-reviewed. The new tests fail without the watcher and each of its elements is mutation-tested; full TS suite, build, and lint green locally; the symptom walked in the browser before the fix (red) and on this branch (green).</sub>

### DIFF
--- a/resources/ext.neowiki/src/components/SubjectEditor/SubjectEditPane.vue
+++ b/resources/ext.neowiki/src/components/SubjectEditor/SubjectEditPane.vue
@@ -197,14 +197,17 @@ function handleClearViolation( payload: { propertyName: string; valuePartIndex: 
 	);
 }
 
-// Existing subjects are expected to be complete, so validate as soon as the
-// editor mounts: pre-existing violations (e.g. a now-empty required field)
-// surface immediately, without the user having to touch a field first.
-watch( subjectEditorRef, ( editor ) => {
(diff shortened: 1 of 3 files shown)


JeroenDeDauw:
Also verified manually

Issue #1267, opened 2026-08-07
ViolationDiff treats a changed violation at the same index as not new


JeroenDeDauw:
`ViolationDiff` identifies violations by (property, code, valuePartIndex) and ignores `args`, so replacing one offending value with a *different* offending value at the same position is classified as pre-existing rather than new.

Consequence: guards that rely on "only new violations block" do not fire on an in-place swap. This includes the always-blocking `relation-target-unresolvable-source` violation from #1265: an existing unresolvable relation target can be edited into a different unresolvable target without the write being blocked, even though writing that target fresh would be refused.

Fix direction: include `args` (or the offending value) in the violation identity, while keeping genuinely unchanged statements non-blocking.

> <sub>AI-authored — Claude Code, `Fable 5 (max)`; found during the #1265 code review, filed on Jeroen's behalf; issue text not human-reviewed.</sub>

PR #1330, opened 2026-08-31
Act on the review of the Subject Sources foundation

alistair3149:
For https://github.com/ProfessionalWiki/NeoWiki/issues/993

Follows-up to https://github.com/ProfessionalWiki/NeoWiki/pull/1265, whose head this targets. Merge it into
that branch before #1265 lands.

Acts on a review of #1265. **The design finding supersedes two of the bug findings**, so the first commit
resolves three of them at once — worth knowing before reading the diff.

## The design finding, and why it swallows two bugs

#1265 added `Violation::alwaysBlocksWrites()`, a hardcoded exemption letting one code block a write whatever
`$wgNeoWikiEnforceValidation` says. [ADR 26](https://github.com/ProfessionalWiki/NeoWiki/blob/master/docs/adr/026-validation-severity-levels.md)
decides the opposite: the setting "remains the master switch", "severity never blocks a write on its own",
"with enforcement off, no violation blocks", and its Consequences claim the carve-out is generalized "instead
of growing a parallel mechanism". That is the mechanism. No ADR authorises it, and #1265 amends ADR 23, not 26.

The exemption also did not work:

* It reached two of the three write paths. `UpdateStatementAction` never consulted it, so the statement
  endpoint accepted what create and replace refused — and enforcement is off by default, so on a stock wiki
  that path had no guard at all.
* On the replace path it was bypassable regardless: `ViolationDiff` identity is
  `(propertyName, code, valuePartIndex)` and ignores `args`, so swapping one unreachable target for a
  different one is not a new violation.

Nothing needs it. Such a relation already degrades wherever it is read: the Neo4j projection drops the edge,
the RDF projection drops the triple, resolution yields no Subject. So the violation keeps its `error` severity
and its reporting, and blocks under enforcement like every other error.

Removing it fixes both bugs above, since there is no longer an exemption to miss or to bypass.

`ViolationDiff` ignoring `args` is left alone: it is
[#1267](https://github.com/ProfessionalWiki/NeoWiki/issues/1267), and changing it would alter blocking
semantics for every violation type — editing one bad URL into a different bad URL would start blocking.

## The rest

* **A foreign relation target is named under its own Source in every projection.** The native projection
  learned this in #1265; the ontology projection kept minting every target under this wiki's base, so an EDM
  or CIDOC-CRM export asserted this wiki owns entities that live elsewhere. The rule now has one home used by
  both projectors.
* **The frontend can read a Schema reference carrying a Source.** `targetSchema` became `string | {source, name}`
  server-side, and the schema format documents the object form — but the frontend still declared a string, so
  following that documentation produced `targetSchema.trim is not a function` on blur and `[object Object]` in
  the picker. The type now admits what the server sends, and nothing is rewritten in transit, so a reference
  the server sent is the reference that goes back.
* **The subject→page index keys only the local Source's ids.** Widening `SubjectId::isValid()` widened this
  filter with it, so a slot key naming another Source became a candidate row — in a `binary(32)` column a
  qualified id can overrun tenfold, truncating the key. ADR 32 says ids are keyed bare and that no row means
  "no local page holds this", which is already the right answer for a sourced Subject.
* **The two statement endpoints parse their Subject id.** They were the last constructing `SubjectId`
  directly, so `<wikiId>:<localId>` was read as foreign and 404'd for a Subject that is right there.

## One review finding withdrawn

The review suspected `schemaContentSchema.json` disagreed with the domain by anchoring a Source key with `$`
where `SubjectId` uses `\z`. It does not: the validator compiles patterns with PCRE's `DOLLAR_ENDONLY`, so
`$` already matches only at the very end. Verified in the container — `"otherwiki\n"` is refused. No production
change; a test pins the agreement, since it rests on a compilation flag no test named.

## Considered, omitted

* Extracting the now-identical blocking predicate from the three write actions. After the revert they carry
  master's own shapes and what repeats is a one-line closure; extracting it would touch three actions, three
  factories and three test factories for no behaviour change, riding on a revert whose value is that its hunks
  vanish from the diff against master.
* A frontend `SchemaReference` value object. A type alias and two pure functions cover every call site without
  a `toJSON` round-trip that only works by accident, and without normalising a value that must survive
  untouched.
* Widening `nwsp_subject_id` instead of narrowing the filter. That is the right change when a Source's Subject
  is genuinely stored in a local slot; nothing does that today, and ADR 32 says otherwise.
* Changing the Subject-level `schema` field's frontend handling beyond widening its type. The lossy collapse
  considered would have gone past the finding.

## Documentation

Two published statements became false and were corrected; nothing else under `docs/` is touched. The
translator note for the `error` severity returns to master's wording, which is true again.

## Reviewing

One commit per concern. Every behavioural commit was seen failing first, and the three tests that could have
passed for the wrong reason were checked by mutating the fix and watching them flip. One test written during
this work was deleted rather than kept: it asserted the right thing but passed under mutation, because Vue
swallows the render error and the wrapper keeps stale props.

> <sub>AI-authored — Claude Code, `Opus 5 (1M context)` (ultracode); review findings and this fix batch requested by @alistair3149, who chose to take the design finding alongside the bug fixes; diff not yet human-reviewed; phpcs, phpstan, 2126 non-Database PHPUnit tests, 1522 TypeScript tests, build and lint green locally, and all 20 CI checks green including the PHPUnit matrix across MW 1.43-1.46 and master.</sub>

The review that produced these findings ran eight parallel passes (`code-review`, `security-review`,
`review-tests`, three design lenses, two second opinions), each blocking finding then put to three
refute-first verifiers; 56 raw findings became seven distinct ones. Two of the four fixed here were proven
with throwaway failing tests before being reported. The fix plans were written by six parallel planners and
cross-checked against each other; two were scoped down and one was withdrawn after its premise was tested in
the container.

</details>


JeroenDeDauw:
Fast-forwarded into `subject-sources-foundation` in full — all seven commits taken, including the enforcement-switch revert, whose ADR 26 argument holds. The #1265 description now reflects the reverted semantics.

Issue #1338, opened 2026-09-01
Move a Subject to another page


alistair3149:
A Subject created in flow, from a relation field in the Subject editor, is stored on the page holding the Subject
whose relation was being filled in ([#971](https://github.com/ProfessionalWiki/NeoWiki/issues/971)). That is the
right default — panes routinely span pages, and asking for a page mid-edit drags page creation into the flow — but
it is often not where the Subject belongs. A person created beside a museum usually deserves a page of their own.

Nothing today can move a Subject from one page to another. The only route is deleting it and recreating it
elsewhere, which mints a new id and breaks every relation pointing at the old one.

## What we want

Move an existing Subject to another page, keeping its id, so the relations targeting it keep resolving.

## Open questions

- Where the affordance lives: the Subject editor's pane (which already names the page a Subject is stored on),
  Special:ManageSubjects, or both.
- Whether the target page may be created in the flow, and whether a Subject can be moved to a page that has no
  Subjects yet.
- What happens to the page's Main Subject marker when the moved Subject is one.
- Whether this is one write or two, and what a half-completed move leaves behind.

Related: [#959](https://github.com/ProfessionalWiki/NeoWiki/issues/959) on dropping the Child Subject term.

> <sub>AI-authored — Claude Code, `Opus 5 (1M context) (xhigh)`; filed as a follow-up the user chose while directing the in-flow creation work; not yet human-reviewed; no code involved, claims about current behaviour checked against the implementation in this session.</sub>

PR #1339, opened 2026-09-01
Create a relation's target without leaving the Subject editor

alistair3149:
For https://github.com/ProfessionalWiki/NeoWiki/issues/971

Filling in a relation whose target does not exist yet meant leaving the editor to create that Subject somewhere
else and coming back — the flow that produces dangling references, and the one Freebase's inline typed creation
avoided. The picker's menu now ends with a create option, and choosing it opens the new Subject as a pane of its
own with the relation already pointing at it.

## What it does

Focusing a relation field opens its menu on the create option, before anything is typed. Typing narrows the
results and names the option after what was typed, so the new Subject arrives with a label. Choosing it mints an
id, adds the Subject to the tree, and brings its editor on screen. A Subject created in one field is offered by
every other field of the same target Schema, so a second reference points at it rather than minting a duplicate.

Nothing is written until the dialog is saved. Minting reserves nothing, so abandoning the edit leaves no
half-made Subject behind — and a draft the user points the relation away from is dropped from the save set
rather than written as an orphan. Drafts are written before the Subjects that refer to them, in dependency
order, so no write names a target the server does not have yet.

**Where the new Subject is stored.** On the page holding the Subject whose relation is being filled in. Panes
routinely span pages, so the dialog's own page would be wrong as soon as the user has drilled in. Asking for a
page mid-edit was the alternative, and #971 notes the page you want often does not exist yet, which drags page
creation into the flow too. Moving a Subject afterwards is filed as #1338.

**Nested panes can now be renamed.** They could not be renamed at all before — only the dialog header renames,
and only the root. That was survivable while every pane held a Subject named elsewhere, but a Subject created
from an empty field would have been unnameable, so this closes it. Separable from the rest if you disagree.

**The create option is a menu item, not Codex's footer.** `CdxMenu` has a `footer` prop for exactly this, but
`CdxLookup` passes none of it through, and all three of its menu-expansion guards read `menuItems` — so a
footer-only menu could never open on focus, and reaching the prop at all needs a cast through `menuConfig`. As a
last item it gets Codex's own separator styling instead.

**An unpickable "No subjects found" item.** With a create option always present the menu is never empty, and
Codex only shows its no-results slot for an empty menu. Pickers without creation keep the slot.

## How it hangs together

`POST /neowiki/v0/subject-ids` and the create route's optional `id` already existed server-side, built so an
importer could wire relations before creating anything; neither was reachable from the frontend.

The editor hands the picker one injected capability (`SubjectCreation`): it creates a Subject of a Schema, and
answers for the drafts of a Schema the session already holds. Naming a draft is therefore the editor's job, not
a per-picker snapshot, which is what lets a rename in a draft's own pane reach the fields pointing at it.
`SubjectEditorDialog` provides it only when its host passes the new optional `onCreate` prop, so a host that
passes none — RedHerb's card, for instance — keeps its relation fields exactly as they were. `RelationInput` is
untouched.

## Review

Reviewed at max effort after the first cut, which found six defects that only appeared once a relation field was
anything but empty — none of them covered by the tests, all now fixed and regression-tested. The worst: a field
already holding a target offered to create a duplicate of it, a draft was invisible to every field but the one
that made it, and a draft the user changed their mind about was written as an orphan anyway.

## Testing

1773 tests pass, up from 1673 on master. The new ones were each mutation-verified: every assertion inverted and
seen to fail by name, with the pre-existing tests staying green. One mounts the real Codex `CdxLookup` and proves
the menu opens on focus with an empty field — the part that depends on Codex internals, since `initialMenuItems`
is snapshotted at setup and the create option has to be in `menuItems` from the first render.

## Manual Browser Check

On a wiki with the demo data, as a user who can edit:

1. Open the `ACME Inc` page and click the infobox's pencil.
2. Click into **Main product** (empty) without typing. The menu opens with one option: **Create a new Product**.
3. Click into **Headquarters**, which already holds *ACME Amsterdam HQ*. The option must read **Create a new
   Office** — not `Create "ACME Amsterdam HQ" as a new Office`.
4. In **Main product**, type `Acme`. The matching products appear, with **Create "Acme" as a new Product** last,
   under a separator. Type something with no matches: a greyed, unpickable **No subjects found**, then the
   create option — and not while the search is still in flight.
5. Create a product from the trailing empty row of **Products**. A pane opens, its header shows the name and
   `Stored on ACME Inc`, the navigator shows it, and the footer says the save covers 2 subjects on 1 page.
6. Open **Main product**: the draft you just made is listed there too. Point that field at it.
7. Rename the draft from its pane header — the navigator and both referring fields follow.
8. Clear the fields pointing at it: the footer drops back to 1 subject, and the draft is no longer marked
   unsaved. Point one back at it and it returns.
9. Save, then reload. The new Product is on the page and the relations resolve.

> <sub>AI-authored — Claude Code, `Opus 5 (1M context) (xhigh)`; feature described in one message by @alistair3149, who chose the storage page and the draft-until-save model from options put to them; diff not yet human-reviewed, though it has been through an AI review pass whose findings are folded in; 1773 vitest tests, build and lint green locally, CI green, and every fixed defect re-checked end to end in a browser against the dev wiki.</sub>

### DIFF
--- a/docs/extending/extending.md
+++ b/docs/extending/extending.md
@@ -379,6 +379,10 @@ editing when `canEditSubject` is true. Editing reads go through the repositories
 props. Saving updates the stores on its own: a Subject write answers with the Subject as persisted and the Schema
 it instantiates, and `nw.useSubjectStore()` records both.
 
+Relation fields inside the dialog offer creating the target Subject on the spot, but only when the dialog is given
+an `onCreate` handler alongside `onSave`. Creating a target writes a new Subject to a page, and which page that is
+is the host's to answer, so a dialog without the handler lets relations point only at Subjects that already exist.
+
 Full example: [`resources/init.js`](https://github.com/ProfessionalWiki/NeoWiki/blob/master/tests/RedHerb/resources/init.js)
 with [`RedHerbCard.vue`](https://github.com/ProfessionalWiki/NeoWiki/blob/master/tests/RedHerb/resources/RedHerbCard.vue).
 
--- a/docs/planning/Relations.md
+++ b/docs/planning/Relations.md
@@ -110,8 +110,8 @@ Unconstrained ("any Subject") targets; cardinality beyond single/multiple; no-va
 
 ### Editing
 
-- In-flow creation and editing of relation targets, and where in-flow-created Subjects live
-  ([#971](https://github.com/ProfessionalWiki/NeoWiki/issues/971)).
+- Moving a Subject to another page ([#1338](https://github.com/ProfessionalWiki/NeoWiki/issues/1338)). In-flow
+  creation stores a target on the page holding the Subject that refers to it, which is often not where it belongs.
 - Red-link create affordance for missing targets ([#1120](https://github.com/ProfessionalWiki/NeoWiki/issues/1120)).
 - Main-Subject prefill as editing sugar (model decision 5).
 
--- a/extension.json
+++ b/extension.json
@@ -603,7 +603,7 @@
 				"neowiki-subject-editor-stored-on",
 				"neowiki-subject-editor-unsaved",
 				"neowiki-subject-editor-target-load-error",
-
+				"neowiki-subject-editor-create-target-error",
 				"neowiki-subject-editor-partial-save",
 				"neowiki-subject-editor-save-scope",
 				"neowiki-subject-tree-label",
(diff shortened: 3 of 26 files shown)


alistair3149:
https://github.com/user-attachments/assets/f1c77e16-3ac6-4f94-b903-2fa034128a06

I made the decision to only store the created subject on the current page to simplify the PR.
Being able to create the subject as another page or on another page is deferred.

Issue #1328, opened 2026-08-31
Subjects editor UX follow-ups: schema placement and badge, deep-tree layout, relation edit-button ambiguity


JeroenDeDauw:
Collected follow-ups from the 2026-08-31 review of https://github.com/ProfessionalWiki/NeoWiki/pull/1323; none blocks its merge. One issue on purpose — split anything out that grows.

* **Schema placement and display.** Move the schema up beside the label so the header is one compact row (the editor is wider now), and unify how a subject's schema is shown — currently a link in the header and subtle text next to tree nodes — into one form, e.g. a badge that always means "schema". That also lets the header schema tag go, and should reduce the misclick where the label pencil gets hit when the user means to edit the schema (the one edit icon near the top attracts the click; a one-row layout probably fixes it, not certain).
* **Deep trees.** Allow horizontal scrolling within the tree pane and/or make the divider between tree and editor pane adjustable.
* **Relation edit-button ambiguity.** The per-relation edit button edits the target subject; someone wanting to change *which* subject is linked may click it and edit the wrong subject's data instead of changing the link. Accepted for now as a fail-once cost — a two-button picker was judged worse — recorded here in case testing shows otherwise.

> <sub>AI-authored — Claude Code, `Fable 5 (max)`; filed on @JeroenDeDauw's go-ahead from the 2026-08-31 sync follow-up list, bundling chosen by him; text not human-reviewed; content checked against PR 1323 and the meeting record.</sub>

JeroenDeDauw:
The edit pencil next to the label also keeps confusing me, as in I keep clicking it automatically when I want to edit the schema

<img width="301" height="96" alt="Image" src="https://github.com/user-attachments/assets/8738b63e-70b4-4b59-8edf-63ab7fbc7256" />

I just did it again after mentioning it in the meeting earlier. And I know what each UI element does. Still incorrectly clicking the pencil.

PR #1340, opened 2026-09-02
Let the reader set how wide each editor's first column is

alistair3149:

The subject editor and the schema editor each drew two columns and had drifted
apart: one was missing the rules its column divider ran into, and both schema
dialogs cut off their own content. This converges them and lets the reader set
how wide the first column is, which is what #1328 asks for so a deep tree can be
read without truncation. The issue's other bullets are untouched.

## What changed

**The schema dialogs stopped clipping.** Both set `overflow: hidden` on the
dialog body while every `overflow-y: auto` inside the editor sat behind the
desktop breakpoint, so below 1120px nothing scrolled and the body swallowed the
overflow: 214px of the property editor was unreachable at a 1000px viewport,
with no scrollbar and the footer still in place. The creator lost content at
every width. Keeping Codex's own `overflow-y: auto` makes the body the scroller
of last resort.

**The header and footer rules are asked for by name.** They were derived from
whether either column overflowed, which is false in exactly the states where the
columns are not scrollers — so they vanished precisely when content was being
cut off, and a schema too short to scroll drew the rule between the columns with
nothing to terminate it. That retires `useOverflowDetection`, its ResizeObserver
and an `overflow` event relayed through four files.

**A divider between the columns.** `usePaneSize` keeps the reader's request and
the width that currently fits as two separate quantities, only the first of them
state, so a container that narrows changes nothing that is remembered and the
pane returns to the chosen width when the room comes back. `PaneDivider` is the
separator: a hairline in a track of its own, dragged with a pointer or moved with
the arrow, paging and Home/End keys.

Deliberately absent: Enter-to-collapse, because a pane sized to zero leaves its
tree and its listbox focusable but invisible, and the APG models the collapsed
position as the minimum of the range where ours is a width still worth having.
Escape stays unbound because cancelling a drag on keydown would also have to
swallow the keyup that closes the dialog.

**Removed from the public API.** `useOverflowDetection` is gone from
`public-api.ts`. Nothing in the tree or in RedHerb used it, and ADR 024's 0.x
contract covers the removal, but it is a published export and worth naming.

## Manual Browser Check

Two editors, in a window at least 1120px wide.

1. Open a Schema page and press its edit button. Check the rules under the
   header and above the footer are present, and that the vertical rule between
   the columns meets them at both ends. Do this on a schema with three or four
   properties, so nothing scrolls — that is the case that used to be missing.
2. Drag the divider left and right. The property list should follow the pointer,
   the cursor should stay a resize cursor over the whole page during the drag,
   and no text should be selected by the gesture.
3. Drag it to the far right. The property editor beside it should not clip its
   labels and should not scroll sideways.
4. Focus the divider with Tab and try the arrow keys, PageUp/PageDown, Home and
   End. Then reload: the width you left it at should come back.
5. Make the window narrower than 1120px. The columns should stack, the divider
   should disappear entirely, and everything below the fold should be reachable
   by scrolling.
6. Repeat 2-4 on a page with a subject that has relations, using the subject
   editor's structure panel.
7. With a wide window, drag the panel wide, then reopen the same dialog in a
   much narrower window. The panel should shrink to fit; widen the window again
   and it should return to the width you chose rather than staying shrunk.
8. Open a property's type dropdown near the bottom of the property editor and
   confirm the menu is not clipped by the dialog.

## Considered, omitted

- Extracting a shared two-column layout component. The two grids differ in row
  count, responsive behaviour, which column is conditional, and the surrounding
  scroll containers; the divider is the part that is genuinely common. Carbon
  and Atlassian both shipped the fused version and then reversed it.
- `aria-valuetext`. Pixels announce as a bare number, but the widget takes all
  its strings from callers by house rule, so this needs a message and a
  screen-reader pass rather than a guess.
- A responsive design for either editor. The schema editor's stacked layout is
  still the acknowledged stopgap its TODO describes, and the subject editor has
  no breakpoint at all.

> <sub>AI-authored — Claude Code, `Opus 5 (xhigh)`; open question from @alistair3149, who reversed my initial recommendation against extracting anything, then steered across ~6 rounds; diff not yet human-reviewed; 1761 unit tests, lint, build and `vue-tsc` green locally, both editors exercised in a browser, and the branch reviewed in-session before this PR.</sub>

The review that preceded this PR ran in the same session that wrote the code, which is the weak case for reviewing — a fresh session would not have inherited the author's framing. It was run as six independent read-only passes plus two refuters rather than a self-read, and one blocking claim was dropped on refutation, but the caveat stands.

An earlier agent in the session wrote an implementation directly into the working tree; it was reverted and rewritten deliberately, so none of it is in these commits.

Two of the design decisions above are contested by the review and left as they are: `paneStorage()` guards a `mw.storage` that `extension.json` already declares as a dependency, and "pane" now names both a layout column and, in the same file, one open subject's edit form.

</details>


alistair3149:
https://github.com/user-attachments/assets/56811076-656e-4ead-b164-e1bae785e6a2

JeroenDeDauw:
Product LGTM

Found a way to remove some lines:
* https://github.com/ProfessionalWiki/NeoWiki/pull/1348

PR #1345, opened 2026-09-02
Show a Subject's Schema as one badge wherever it is named

alistair3149:

A Subject's Schema was drawn four ways: a permission-gated link in the editor header that opened the
Schema editor rather than navigating, subtle text beside a navigator row, a subtitle link in the
infobox, and another beside a statement count in the Subjects manager. Nothing about any of them said
"this string is a Schema" — the register they share is also the register of statement counts, subject
ids and the tree's group captions.

`SchemaNameDisplay` renders the bare name as a badge and owns the rules each surface used to decide
for itself: where the link goes, whether it opens a tab of its own, and when the name is withheld
because it would only repeat the name beside it.

**A badge, deliberately not a `CdxInfoChip`.** A chip is a bordered pill at `@font-size-small`, which
is the size of a navigator row's own text; thirteen of them down the demo wiki's tree read as a wall.
The badge is a step smaller, filled rather than outlined, and on `@background-color-neutral` rather
than `-subtle`, which would all but vanish on a hovered row and clash on a selected one. Chip and
badge stay distinct shapes for distinct ideas — the chip means Property Type on the Schema page, and
this PR leaves it alone.

**The tree gets a slot rather than a stylesheet.** `NeoTree` gains a `secondary` slot, rendered
inside the element that held the plain text, so the tree learns nothing about Schemas and one
component still owns the badge. The row's DOM is unchanged, which matters:
`SubjectTree.spec.ts:809` asserts a row's children and text exactly, to catch a re-introduced glyph,
and it still does.

**Always a link, and the withholding rule moves in.** Reading a Schema is not an edit right, so the
badge no longer hides behind `neowiki-schema-edit`. `schemaNameToShow` takes the Subject and answers for every surface, so a Subject whose own name is its
Schema name (ADR 31) shows it once — the infobox and the Subjects manager printed it twice before. It reads
the Subject rather than the name a surface shows, since `(unnamed Company)` names the Schema too.

**Not here: the editor's own header.** It keeps `Schema: Company` for now. The follow-up removes that
subtitle wholesale, and changing its click target here would quietly retire the only in-dialog route
to the Schema editor — a call that belongs with the PR redesigning that header. So this PR leaves the
editor header inconsistent with the tree below it, on purpose and briefly.

Also here: `neowiki-infobox-type` is deleted, being declared and defined but referenced by no code;
`DataExportButton`'s copy of the visually-hidden rules moves to a shared mixin.

## Testing

All tests pass. The new ones were mutation-verified: disabling the withholding rule, making `new-tab` stop
opening one, linking the tree's badge, removing the badge from the tree, dropping the guard at a manager row,
handing the tree the bare Schema name, and ignoring the label typed in a pane each turn a test red by name.
Several of those found real gaps — the first cut of this PR had no test that would have noticed a surface
forgetting to withhold the badge.

## Manual Browser Check

On a wiki with the demo data, logged in:

1. Open `ACME Inc`. Under the infobox title, `Company` is a small filled badge, not blue link text.
   It links to `Schema:Company` in the same tab. Hovering underlines it.
2. Open a page whose Main Subject has no label of its own — its title already is its Schema name.
   The badge is absent rather than repeating the title. (`Special:Subjects` on such a page shows the
   same: the row named after its Schema carries no badge, one named after its page does.)
3. Click the infobox pencil. Every navigator row carries the same badge. Rows stay 32px; the badge
   does not push them taller. The badge is not a link here — the row is the click target.
4. Select a row. The badge still separates from the selected row's blue-tinted background.
5. Switch the skin to dark. The badge follows: dark fill, light text, same contrast relationship.
6. With a screen reader, the badge announces "Schema: Company" — the noun is present once, not twice.
7. The dialog header still reads `Schema: Company` as a link that opens the Schema editor. That is
   deliberate for this PR; the follow-up removes it.

> <sub>AI-authored — Claude Code, `Opus 5 (1M context)`; design chosen by @alistair3149 from options put to them in a design artifact, including the badge-not-chip call and the decision to unify behaviour as well as appearance; diff not yet human-reviewed; verified: 1689 vitest tests, lint and build green locally, six mutations confirmed red, and the badge checked in a browser against a dev wiki in light and dark, in the infobox, the Subjects manager and the editor's navigator.</sub>

### DIFF
--- a/extension.json
+++ b/extension.json
@@ -490,7 +490,6 @@
 				"neowiki-property-type-unregistered-notification",
 				"neowiki-boolean-true",
 				"neowiki-boolean-false",
-				"neowiki-infobox-type",
 				"neowiki-infobox-edit-link",
 				"neowiki-field-required",
 				"neowiki-field-unique",
--- a/i18n/en.json
+++ b/i18n/en.json
@@ -78,7 +78,6 @@
 
 	"neowiki-schema-label": "Schema: $1",
 
-	"neowiki-infobox-type": "Type",
 	"neowiki-infobox-edit-link": "Edit",
 	"neowiki-field-required": "Please provide a value.",
 	"neowiki-field-unique": "All values must be unique.",
--- a/i18n/en.json
+++ b/i18n/en.json
@@ -123,7 +123,7 @@
 
 	"neowiki-subject-generated-name": "(unnamed $1)",
 
-	"neowiki-subject-editor-title": "Editing $1",
+	"neowiki-subject-editor-title": "Edit subject",
 	"neowiki-subject-editor-save": "Save changes",
 	"neowiki-subject-editor-label-field": "Subject label",
 	"neowiki-subject-editor-rename": "Rename subject",
--- a/i18n/qqq.json
+++ b/i18n/qqq.json
@@ -17,7 +17,7 @@
 	"action-neowiki-admin": "{{doc-action|neowiki-admin}}",
 	"specialpages-group-neowiki": "Group heading for NeoWiki special pages on [[Special:SpecialPages]].",
 	"neowiki-special-schemas": "Title and description of the [[Special:Schemas]] page.",
-	"neowiki-schema-label": "Label used to identify the schema. $1 is the schema name or link.",
+	"neowiki-schema-label": "Names a subject's schema. $1 is the schema name or a link to it. Also used as the accessible name of the schema badge, whose visible text is the bare name.",
 	"neowiki-cypher-raw-error-json-encode": "Error message shown when JSON encoding of query results fails.",
 
 	"neowiki-cypher-error-empty-query": "Error shown on the <code>{{#cypher_raw}}</code> parser function and the <code>nw.query()</code> Lua function when the supplied Cypher query is empty or only whitespace.",
--- a/i18n/qqq.json
+++ b/i18n/qqq.json
@@ -68,6 +68,7 @@
 
 	"neowiki-subject-generated-name": "The name shown for a subject that has no label of its own, so its name was derived from its schema rather than chosen by anyone. $1 is the schema name: a noun the wiki's own users invented, in an unknown language and grammatical gender, so do not inflect it, do not change its case, and make no word agree with it. The round brackets mark the whole string as a stand-in the system supplied rather than a name someone typed, as in {{msg-mw|Blanknamespace}}; use whichever brackets your language uses for that.",
 
+	"neowiki-subject-editor-title": "Title of the subject editor dialog, and its accessible name. Names the task rather than a subject: the dialog can edit several at once, and each one is named by the pane that edits it.",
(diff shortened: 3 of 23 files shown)


alistair3149:
image
image

JeroenDeDauw:
Merged master into this branch in https://github.com/ProfessionalWiki/NeoWiki/commit/6c9c9af22cc0a1103c14661a6a464a04a62583f9. The conflicts were the overlap with https://github.com/ProfessionalWiki/NeoWiki/pull/1359: it shows a Subject nobody named as `(unnamed Company)` and decides the tree's schema label by its new `nameIsGenerated` flag, while this PR decides by comparing the displayed name. A plain textual merge would have compared the *marked* name in the pane and the tree and shown the badge beside `(unnamed Company)`, and the pane had no test that would have noticed.

What the merge commit does instead:

- `schemaNameToShow` takes the Subject and withholds the badge where the Subject's own name (`getDisplayName()`, unmarked) is its Schema name. That covers the generated name and one typed after the Schema, and nobody can hand it the marked string.
- `SchemaNameDisplay` loses its `displayName` prop and always renders; each surface guards its own badge with `v-if`. The walk carries the outcome as `WalkNode.schemaLabel`, replacing master's `nameIsGenerated`, whose only consumer was that decision.
- Tests: the helper and badge specs follow; the manager's wiring test became an outcome test (main row badged, the `(unnamed Person)` row not); three pane tests added. Four mutations (never withholding, the pane ignoring the typed label, a manager row losing its guard, the walk handing over the bare name) each turn exactly the expected tests red.

Two commits on top:

- https://github.com/ProfessionalWiki/NeoWiki/commit/990ecd6b2e330aef59771fa332b82969b4e2ecd0: master is red since https://github.com/ProfessionalWiki/NeoWiki/pull/1357 merged (a five-argument `Subject` call). The one-line fix is https://github.com/ProfessionalWiki/NeoWiki/pull/1360; the same line is carried here so this PR's CI means something meanwhile, and whichever lands first makes the other a no-op.
- https://github.com/ProfessionalWiki/NeoWiki/commit/04a8923b7a406213925185017011437df76ee2ff: the bullet before a row's statement count assumed a badge before it and dangled once the badge was withheld. It is now drawn only after one.

Checked in a browser on a worktree wiki: the infobox, and `?action=subjects` on Rijksmuseum with an unnamed `Attendance` child added. Labelled rows carry the badge; `(unnamed Attendance)` carries neither badge nor bullet. The PR body's withholding-rule paragraph and the test-count sentence are edited to match.

> <sub>AI-authored — Claude Code, `Fable 5.1 (max)`; one-line ask from @JeroenDeDauw to resolve the conflicts, no revisions; diff not yet human-reviewed; 1889 vitest tests, vue-tsc build and lint green locally in a scratch container, four mutations checked, two surfaces checked in a browser.</sub>

JeroenDeDauw (approved):
LGTM, checked stuff locally. Fable saying you should have another look due to semantic merge conflict resolution

PR #1350, opened 2026-09-03
Let every pane name itself, and shrink the dialog's header to suit

alistair3149:

Stacked on #1345 — its base is `1328-schema-badge`, because this uses the badge that PR adds.
Review that one first; this diff is only the commit on top.

The dialog's header named the root Subject and carried the field that renamed it, so a nested
pane could not be renamed at all and the header described a Subject the user might not be looking
at. It named the root's Schema too, with the same problem. Neither could follow the active pane:
`title` is a `CdxDialog` prop, so binding it re-renders the dialog, and under the spec suite's
teleport stub that remounts every pane.

So the header stops naming anything. Its title is the task — "Edit subject" — and the Subject's
name, its rename control and its Schema move into the pane that edits them, where they are correct
by construction.

**The Schema is a link that usually does not navigate.** It is the badge from #1345, so it carries
that component's interactive styling honestly rather than being a styled span with a click handler.
A plain left click opens the Schema editor, as the old header link did; a modified or middle click
is left to the browser, so the Schema page stays reachable and unsaved work is never at risk from
the primary gesture. Only the root pane offers the editor, since that is the Schema
`SchemaEditorDialog` is bound to — a nested pane's badge is a plain link. The Schema editor itself,
and the `onSaveSchema` prop three call sites pass, are untouched.

**The header keeps a title rather than losing one.** "Edit subject" names the task, which is the
only thing about this dialog that does not change as panes are switched. `neowiki-subject-editor-title`
therefore takes no parameter now, and the dialog's accessible name and its visible heading say the
same thing. A header that named the active pane would have to re-render `CdxDialog`.

**The root pane keeps no "Stored on" line.** The root's Subject is on the page the editor was
opened from, so the line would say what the reader already knows. `SubjectEditPane.spec` asserted
that before this change and still does.

## What the header costs now

49px, against the 81px a title over a subtitle needed — `align-items: center` and
`padding-block: @spacing-50`, since the row is one line of static text beside a 32px close button.
The two rules that existed to space the old title/subtitle pair go with it, as does the
`&-schema` block.

In the pane, the name takes the free space so the Schema and the storage line sit flush to the end,
and the row wraps rather than crushing either side once the pane is narrow enough — measured, it
holds one line down to about 260px and wraps by 200px.

## Testing

1858 tests pass. Five existing tests asserted the old arrangement — the root pane being unnamed and
without a rename control, the schema link living in the dialog header, and the dialog's accessible
name carrying a Subject's display name. Each was rewritten to pin the new contract rather than
deleted, and two were added for the click behaviour: that a user who cannot edit the Schema is left
to navigate, and that a modified click is too.

## Manual Browser Check

On a wiki with the demo data, logged in as a user who may edit Schemas:

1. Open `ACME Inc` and click the infobox pencil. The dialog header is one shallow row reading
   **Edit subject**, with no Subject name and no Schema.
2. The form's own header reads **ACME Inc** with a pencil, and a **Company** badge flush right.
   The badge is blue — it is a link.
3. Click the badge. The Schema editor opens; the page does not navigate.
4. Ctrl-click or middle-click it instead. `Schema:Company` opens in a new tab and the Schema editor
   does not open.
5. Click a child row in the navigator. That pane names itself, shows *its* Schema, and adds
   "Stored on <page>". Rename it from its own header — this was impossible before.
6. Return to the root. Its name is still there, and renaming it still works.
7. As a user without `neowiki-schema-edit`, the badge is still a link and still reaches the Schema
   page; clicking it navigates rather than opening the editor.
8. Narrow the window until the pane is around 200px. The pane header wraps; neither the name nor
   the badge is crushed.

> <sub>AI-authored — Claude Code, `Opus 5 (1M context)`; design chosen by @alistair3149 from options put to them in a design artifact, with the header treatment, the badge styling and the pane layout directed by them during the work; diff not yet human-reviewed; verified: 1858 vitest tests, lint and build green locally, and the header, badge states and wrap behaviour measured in a browser against a dev wiki.</sub>

### DIFF
--- a/i18n/en.json
+++ b/i18n/en.json
@@ -121,7 +121,7 @@
 	"neowiki-delete-success": "Deleted $1",
 	"neowiki-delete-error": "Failed to delete $1",
 
-	"neowiki-subject-editor-title": "Editing $1",
+	"neowiki-subject-editor-title": "Edit subject",
 	"neowiki-subject-editor-save": "Save changes",
 	"neowiki-subject-editor-label-field": "Subject label",
 	"neowiki-subject-editor-rename": "Rename subject",
--- a/i18n/qqq.json
+++ b/i18n/qqq.json
@@ -66,6 +66,7 @@
 	"neowiki-schema-editor-description-collapse": "Accessibility label and tooltip for the button that collapses an expanded schema description back to two lines.",
 	"neowiki-schema-editor-description-add": "Label of the button that starts writing a description for a schema that has none, shown in the schema editor dialog header.",
 
+	"neowiki-subject-editor-title": "Title of the subject editor dialog, and its accessible name. Names the task rather than a subject: the dialog can edit several at once, and each one is named by the pane that edits it.",
 	"neowiki-subject-editor-label-field": "Accessibility label for the input shown in place of the dialog title while renaming a subject in the subject editor dialog.",
 	"neowiki-subject-editor-rename": "Accessibility label for the button next to the subject editor dialog title that starts renaming the subject.",
 	"neowiki-subject-editor-edit-target": "Aria-label of the per-target edit button shown beside a selected relation target in the subject editor. Clicking it opens that target subject for editing in a pane.",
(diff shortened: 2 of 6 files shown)


alistair3149:
image
image

PR #1361, opened 2026-09-06
Refactor the Subject move: read the source page once, drop dead paths

JeroenDeDauw:
Follows-up to https://github.com/ProfessionalWiki/NeoWiki/pull/1356, rebased onto master after it merged. One commit per item, so any of them can be dropped on its own. User-visible behaviour is unchanged; the one API-visible difference is under the second item.

- **One read of the source page.** `PageSubjects::without()` answers a copy without the Subject. The move writes the copy, and the rollback writes back the object it read, which nothing has mutated. The second read of the source page goes, and with it the reliance on repository reads handing out distinct objects, which `SubjectRepository` never promised.
- **One not-found outcome for the source.** A source page that vanishes between the read check and its write is presented as the Subject not found: it is written first, so nothing has been written yet. `presentSourcePageNotFound` goes. The response is still 404, with the message `Subject not found` instead of `Page not found`.
- **One error path in the REST client.** `moveSubject` no longer reads a server message off a non-ok Response. The production client rejects on every non-2xx but 422, so that branch never carried one. The rejection path still delivers the server's message to the dialog, unchanged.
- **Three `@group Database` tests dropped**: source-page 403, source untouched on an unreadable target, edit summary on both pages. `MoveSubjectActionTest` covers each; the target-page 403 test keeps proving the Authority wiring.

Considered, omitted:

- Moving page creation out of `MoveSubjectDialog` into a `PageCreator` service behind the registry: it takes the `mw.Api` call and error-code parsing out of the Vue component, but costs an interface, a class and three registration points for one caller. Worth doing only together with `MappingCreatorDialog`, which has the same inline `mw.Api().create`, as a second caller.
- Replacing `dropFromRegistryOnceUnlisted` with a plain listing reload: the moved Subject's registry copy carries its old page, so dropping it does real work.
- Sharing the Codex lookup handling between `PagePicker` and `SubjectPicker`: the largest duplication in the original PR, but it means reworking SubjectPicker; a PR of its own.

> <sub>AI-authored — Claude Code, `Fable 5.1 (max)`; @JeroenDeDauw asked for a follow-up PR that reduces complexity after an in-session assessment of the original, scope chosen by the assistant, then cut by one item at his direction; diff not yet human-reviewed; `make cs`, the move-related PHPUnit classes and the full `tsci` green locally on the final branch, the full PHPUnit suite green on the branch before the cut, each new test mutation-checked.</sub>

Design, diff review and this text by `Fable 5.1 (max)`; implementation, tests and the rebase onto master by `Opus 5 (max)` subagents from written specs.

</details>

### DIFF
--- a/resources/ext.neowiki/src/persistence/RestSubjectRepository.ts
+++ b/resources/ext.neowiki/src/persistence/RestSubjectRepository.ts
@@ -368,7 +368,7 @@ export class RestSubjectRepository implements SubjectRepository {
 		// A move can fail for reasons the user can act on - the target page is protected, the Subject
 		// is already there - so the server's own message is carried through rather than collapsed into
 		// a status code. The production client rejects on every non-2xx but 422, so that message
-		// arrives on the rejection rather than on a response.
+		// arrives on the rejection.
 		try {
 			response = await this.httpClient.post(
 				`${ this.mediaWikiRestApiUrl }/neowiki/v0/subject/${ id.text }/move`,
@@ -388,7 +388,7 @@ export class RestSubjectRepository implements SubjectRepository {
 		}
 
 		if ( !response.ok ) {
-			throw new Error( await this.errorMessageOf( response ) ?? 'Error moving subject' );
+			throw new Error( 'Error moving subject' );
 		}
 	}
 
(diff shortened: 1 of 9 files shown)


JeroenDeDauw:
Ugh. Try to get claude to remove lines. :thinking:

alistair3149:
> Moving page creation out of MoveSubjectDialog into a PageCreator service behind the registry

Might be a good direction to take because of #1374. @malberts what do you think?
This is not a blocker and can be done as a follow-up.

PR #1364, opened 2026-09-07
Give the move-subject action an icon that is not the drag cursor

JeroenDeDauw:
Follows-up to https://github.com/ProfessionalWiki/NeoWiki/pull/1356

The Move action on the Data tab used `cdxIconMove`, the four-way arrow that is also the
browser's drag cursor, one button away from the row's drag handle in a list that is dragged to
reorder. A glyph that says "drag me" sat on the button that relocates a Subject to another page.

It now uses `cdxIconArticleRedirect`, a page with an arrow leading away from it: the
container-plus-arrow shape mail and file tools use for "move to", with a page as the container.
The dialog's confirm button and the overflow-menu entry carry the same icon, and the
ResourceLoader icon list follows.

Considered, omitted: a custom page-with-arrow glyph without the redirect association, and moving
the action into a labelled overflow menu on desktop.

> <sub>AI-authored — Claude Code, `Fable 5.1 (max)`; one-line ask from @JeroenDeDauw after an in-session icon discussion, no revisions; diff not yet human-reviewed; SubjectsManager vitest specs and eslint run locally, icon checked in the dev wiki, CI pending.</sub>

### DIFF
--- a/extension.json
+++ b/extension.json
@@ -449,7 +449,7 @@
 						"cdxIconEdit",
 						"cdxIconDownload",
 						"cdxIconLink",
-						"cdxIconMove",
+						"cdxIconArticleRedirect",
 						"cdxIconCheck",
 						"cdxIconSearchCaseSensitive",
 						"cdxIconArticles",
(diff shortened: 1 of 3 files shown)


JeroenDeDauw:
image

Issue #1120, opened 2026-07-21
Render missing relation targets as red links with a create affordance


JeroenDeDauw:
## Problem

A missing relation target is now legitimate, wiki-native state, not an error:

- Referenced Subjects are kept as stub nodes (https://github.com/ProfessionalWiki/NeoWiki/pull/1080).
- `relation-target-not-found` is a non-blocking **warning**, not a blocking error
  (https://github.com/ProfessionalWiki/NeoWiki/pull/1082).

But the display layer still treats a missing target as an error.
`resources/ext.neowiki/src/components/Value/RelationDisplay.vue` renders an unresolved target as a `<span class="error">`
with a "Subject not found" title.

## Wanted

MediaWiki red-link semantics for relation targets:

- Render an unresolved target as a visibly not-yet-created link (red link), not an error.
- Clicking it starts a creation flow with the relation property's target schema pre-selected.

## References

- In-flow target creation: https://github.com/ProfessionalWiki/NeoWiki/issues/971.
- Precedent from the previous codebase: https://github.com/ProfessionalWiki/LegacyNeoWiki/issues/735 (a missing target
  must not break rendering; show a placeholder).
- Tracked home for the missing-target-UI part of the dangling-references row of
  https://github.com/ProfessionalWiki/NeoWiki/issues/630.

---

> <sub>AI-authored — Claude Code, `Opus 4.8 (max)`; detailed spec from @JeroenDeDauw via an orchestrating agent; not yet human-reviewed; code/doc citations and cross-referenced issues/PRs verified against origin/master and the GitHub API.</sub>

PR #1374, opened 2026-09-08
Create a page and its Main Subject in one step

### LINKED ISSUE #1355: Investigate creating a page and its Main Subject in one step
## Problem

A Subject can only be created from the page tools of a page that already exists. Someone who wants "a new Person" creates a page, names it, opens **Create subject**, picks the Schema, and names the Subject again. The UI never says the page is a prerequisite, so users who think in entities rather than pages, as anyone with a database or Wikibase background does, expect **Create subject** to create the thing itself.

Subjects stay stored on pages ([ADR 4](https://github.com/ProfessionalWiki/NeoWiki/blob/master/docs/adr/004-use-dedicated-slot.md), [ADR 32](https://github.com/ProfessionalWiki/NeoWiki/blob/master/docs/adr/032-subject-page-index.md)); this is about the entry point.

ECHOLOT requirement FR5.4, manual creation of records through forms ([echolot-WP2#42](https://github.com/ECHOLOT-ECCCH/echolot-WP2/issues/42)), currently maps only to [#677](https://github.com/ProfessionalWiki/NeoWiki/issues/677).

## What we want

A time-boxed design investigation, not implementation, into schema-first creation: pick a Schema first, and the page comes into existence with its Main Subject. Candidates:

1. **"Create new <Schema>"** on the Schema page or Special:Schemas: a title plus the Schema's fields, saved as a new page with its Main Subject. The 2023 codebase shipped this as `Special:NeoSubject/<Schema>` and dropped it when the form moved to TypeScript.
2. **Page Schemas with a "create page of type" flow**, sketched as Phase 2 in [#593](https://github.com/ProfessionalWiki/NeoWiki/issues/593) and never broken out.
3. **A page-title step in the Subject creator** when it is opened from outside a page.
4. **A fresh page for an in-flow relation target** ([#971](https://github.com/ProfessionalWiki/NeoWiki/issues/971), [#1120](https://github.com/ProfessionalWiki/NeoWiki/issues/1120)) instead of the host page ([#1338](https://github.com/ProfessionalWiki/NeoWiki/issues/1338)).

Outcome: a recommendation on which entry points to build, or a decision that page-first stays the model, plus how the UI should say so.

## Open questions

- Whether the Main Subject label and the page title are one value at creation ([#750](https://github.com/ProfessionalWiki/NeoWiki/issues/750), [#959](https://github.com/ProfessionalWiki/NeoWiki/issues/959)).
- What happens when a page with that title exists, with or without a Main Subject.
- Where the entry lives for wikis used as a database rather than as pages ([#765](https://github.com/ProfessionalWiki/NeoWiki/issues/765)).
- Whether this belongs in NeoWiki core or a forms layer ([#677](https://github.com/ProfessionalWiki/NeoWiki/issues/677)).

## Prior discussion

- [#427](https://github.com/ProfessionalWiki/NeoWiki/issues/427) and [#750](https://github.com/ProfessionalWiki/NeoWiki/issues/750): the current page-first creation flow and its page-tools entry point.
- [#985](https://github.com/ProfessionalWiki/NeoWiki/issues/985): why Subjects need pages at all.
- [#959](https://github.com/ProfessionalWiki/NeoWiki/issues/959): Main Subject semantics, and whether ECHOLOT-style use cases need page links.

> <sub>AI-authored — Claude Code, `Fable 5.1 (xhigh)`; one-line ask from @malberts after a research thread on prior creation-flow work, redirected once; body reviewed by @malberts in chat before posting; issue numbers, ADR paths and cited decisions checked against the GitHub API and origin/master, plus a fresh-context judge pass.</sub>

malberts:
Fixes https://github.com/ProfessionalWiki/NeoWiki/issues/1355

A Subject could only be created from a page that already existed: create the page, name it, open Create subject,
pick the Schema, name the Subject again. Nothing said the page was a prerequisite, and users who think in entities
expected Create subject to create the thing itself.

## What it does

Schema-first creation, with the page created on the way:

- **Special:CreateSubject** opens the Subject creator with a Page field: matching existing pages plus "Use X as a new
  page". A subpage pins the Schema (`Special:CreateSubject/Person`).
- A **Create Person** call to action below the description of `Schema:Person` (named after each Schema), leading
  there with the Schema pinned.
- **Create subject** in the sidebar links there from every page. The page tool that adds a Subject to the page being
  viewed is now called **Create subject here**, and the Data tab's button **Create subject**, so every entry uses
  one verb.

Picking a page that exists adds the Subject to it: as its Main Subject when it has none, otherwise alongside the one
it has, which the dialog names. A new page is created empty on save and the Subject becomes its Main Subject; the
browser then lands on that page.

**The label is not copied from the page title.** A Main Subject without a label already displays as its page name
(ADR 31), and a stored copy would be exactly the default that ADR removed: it cannot be told from a chosen label, so
a page move would stop renaming the Subject. The page title is the required input; the label stays optional.

**Two writes, not one endpoint.** The page is created through MediaWiki's own API and the Subject through the
existing endpoint, as the move dialog does. Core enforces createpage, nothing is written until save, and the page id
is kept so a retry after a failed Subject write does not create the page twice. The cost is an empty page when that
retry never comes, and two recent-changes entries. A NeoWiki endpoint writing both slots in one revision is the
follow-up if that bites.

**The page picker rather than a title field.** A title collision becomes a visible choice between the existing page
and a new one, instead of a silent merge or an error.

**A link rather than an in-place dialog on Schema pages.** One mount point, and a URL that later entry points (red
links, in-flow creation) can reuse.

**A special page rather than a sidebar link that opens the dialog in place.** The frontend bundle (Vue, Codex,
Pinia and NeoWiki's own code) is loaded only on content, Schema, Layout and Mapping pages and on NeoWiki's special
pages. Anywhere else the link would first have to load all of that before a dialog could open, so a page that loads
it is the plain, reliable path. The URL is also what makes the entry point placeable: `[[Special:CreateSubject/Person]]`
works in any page, template or MediaWiki:Sidebar without JavaScript, and it is what the red-link affordance for
missing relation targets (https://github.com/ProfessionalWiki/NeoWiki/issues/1120) can point at. Opening the dialog in
place where the bundle is already loaded is a possible follow-up, not a change of approach.

**The special page is restricted to edit and refuses users without createpage.** A special page names one
restriction; edit is the one the Subject write needs and the one anonymous users lack on hosted wikis, so the page is
hidden from them. The sidebar link and the Schema page button require the same pair of rights.

## Considered, omitted

- Prefilling the page title from a URL parameter.
- Edit notices for the chosen page.
- Locking a pinned Schema; the back button still reaches the picker.

## Screenshots (UI, Playwright)

| | |
|---|---|
| **Sidebar on a content page.** The page tool now reads "Create subject here"; "Create subject" is the new link to the special page. | sidebar-create-subject |
| **Schema page.** "Create Person" below the description, with the pencil alone in the corner. | [SCREENSHOT: schema-page-create-button.png] |
| **`Special:CreateSubject/Person`.** The dialog on the Person step, with a new page chosen in the Page field. | special-createsubject-dialog |
| **The created page.** Its Main Subject rendered as the infobox under the page name. | created-page-infobox |
| **An existing page chosen.** The notice naming the page's Main Subject; the new Subject is added alongside it. | existing-page-notice |

## Testing

Vitest covers the creator's page-first mode (page required, new page created then written as Main Subject, existing
page with and without a Main Subject, retry without a second creation, the taken-title error), the special page's
frontend, the Schema page button and its permission hint, and the shared page creation. PHPUnit covers the special
page (mount point, subpage normalization, both refusals) and the sidebar link and its gate. The save path's guards
(the page and tier decided when the save started, no second submit while one is in flight, stale reads of a page the
user moved off) each have a test that fails without them.

> <sub>AI-authored — Claude Code, `Fable 5.1 (high)`; one-line ask from @malberts, design proposals approved without changes, one mid-task instruction on process; diff reviewed by two Opus subagent passes (code review, mutation-checked test review) with their findings applied, not yet human-reviewed; vitest, phpcs, phpstan and the full PHPUnit suite pass locally, Playwright check on the dev stack with screenshots, CI pending.</sub>


JeroenDeDauw:
> A NeoWiki endpoint writing both slots in one revision is the
follow-up if that bites.

Makes sense to log as a low priority follow up. Having two entries in recentchanges and similar isn't nice.

i18n/_Aliases.php
```diff
@@ -8,4 +8,5 @@
 	'Layouts' => [ 'Layouts' ],
 	'Mappings' => [ 'Mappings' ],
 	'GraphStores' => [ 'GraphStores' ],
+	'CreateSubject' => [ 'CreateSubject' ],
```
JeroenDeDauw, line 11: Adding NewSubject as alias seems useful. Can see people coming from Wikibase typing that

JeroenDeDauw:
> A special page rather than a sidebar link that opens the dialog in place. 

The direct dialog seems like a useful follow-up. Same with `"Create new " on the Schema page or Special:Schema"` from the linked issue.

JeroenDeDauw:
Ideally we also end up with consistent behavior with the "Create new $SchemaName" in the SubjectEditor. 

image

Though there might be reason to diverge as well. Warrants a look as follow-up.

JeroenDeDauw:
The UI on Schema pages needs work

image

That button looks weirdly placed. I also don't like it goes to the special page and then opens the dialog. It could just open the dialog on the Schema page itself. Right now if you click the button and then you dismiss the dialog, you end up on `Special:CreateSubject/Artist` which is a blank page with a button rather than the Schema page you came from.

Additionally, and this could also be a follow-up, `Special:Schemas` could have a + Icon (or other sensible icon) for creating a new Subject of that type per Schema row.

JeroenDeDauw:
I like the text of the "Create subject here" toolbar link got its current behavior made more obvious

image

However I think we ought to have a new field/section in the dialog where you choose if you add it to the current page or to another page (with the option to create a new page for the latter case).  

When opened via the sidebar link on a standard Page, the default would be that page, so the current behavior. With ideally there being no extra steps compared to the current flow. So you'd be able to choose to put the Subject on another Page, but if you want to put it on the current page, no more work is needed than currently.

When opened from Special:CreateSubject or another location that is not a standard Page, e.g. Special:Schemas or Schema:Example, you'd just have the option to create it on a new page or an existing page (so dialog hides the "current page" option) and the default would be "create new page". That's the sensible default since adding to an existing page is typically done when you look at that page, and we need to make it easy to add subjects for people that don't care about wiki pages.

JeroenDeDauw:
@malberts I'm taking over this task

PR #1386, opened 2026-09-09
Take a withdrawn username out of the graph projection

### LINKED ISSUE #1246: A suppressed username stays in the graph and is readable anonymously
## Problem

The `lastEditor` Page Property is written into the graph from the editing user at save time. Nothing re-projects a page when a revision's visibility changes afterwards, so suppressing a username with RevisionDelete hides it in MediaWiki while the graph keeps serving it.

`extension.json` registers no revision-visibility hook — the projection entry points are `RevisionFromEditComplete`, `PageDeleteComplete`, `PageUndeleteComplete` and `AfterImportPage`, none of which fire when only visibility changes.

## Reproduction

On a dev wiki, suppressing the username on a page's current revision (`DELETED_USER`):

1. MediaWiki's API reports the revision as `"userhidden": ""` — the name is gone from the normal surfaces.
2. The graph still holds `lastEditor` for that page.
3. An anonymous, logged-out call to the Cypher query endpoint returns it:

```
POST /rest.php/neowiki/v0/query/cypher
{"cypher":"MATCH (p:Page) WHERE p.id = 127 RETURN p.lastEditor AS name"}

{"columns":["name"],"rows":[{"name":"NeoWiki"}],"truncated":false,"resultCount":1}
```

The RDF export path is not affected: `RdfPageLoader` passes `$revision->getUser()`, whose default `FOR_PUBLIC` audience nulls a suppressed name. Only the graph write path keeps the raw value, so the two read surfaces disagree.

## Scope

Pre-existing for pages carrying Subjects. [#1190](https://github.com/ProfessionalWiki/NeoWiki/issues/1190) widened it to every page on the wiki, since every page now has a node with a `lastEditor`.

The same question applies to any other Page Property derived from revision data that can later be hidden — `categories` comes from the revision's content, so suppressing a revision does not remove categories the graph already recorded.

Related but distinct from [#1059](https://github.com/ProfessionalWiki/NeoWiki/issues/1059): that covers permission filtering of query results at read time, whereas this is stale data in the projection itself, which read-time trimming by page readability would not catch — the page stays readable, only the username is suppressed.

## Possible directions

1. **Re-project on visibility change.** Hook the relevant core hook (`ArticleRevisionVisibilitySet` or its current equivalent) and refresh the page node, reusing the refresh-without-an-edit path from [#889](https://github.com/ProfessionalWiki/NeoWiki/issues/889). Keeps the graph consistent with MediaWiki, at the cost of one more write path to keep correct.
2. **Project through the public audience.** Write `lastEditor` using the same `FOR_PUBLIC` audience the RDF path already uses, so a suppressed name never reaches the graph in the first place. Simpler and closes the write path, but does not fix pages already projected — those need a rebuild — and it means the graph cannot answer questions an administrator legitimately could.

(2) plus a rebuild may be enough on its own; (1) is what makes the graph track later changes.

## Note

The description of [#1226](https://github.com/ProfessionalWiki/NeoWiki/pull/1226) says the widened anonymous `neowiki-query` grant "discloses nothing MediaWiki core does not already expose" on an open wiki. That is not accurate for this case, and the claim is worth correcting wherever it is repeated.

> <sub>AI-authored — Claude Code, `Opus 5 (1M context)`; found while reviewing #1226 and reproduced on a dev wiki against merged master. Not human-reviewed before posting; the directions above are unvalidated suggestions.</sub>

alistair3149:
Fixes https://github.com/ProfessionalWiki/NeoWiki/issues/1246

The graph and RDF projections stored `lastEditor` as the user who saved the revision. MediaWiki can
later withdraw that name from public view — with RevisionDelete on the revision, or a hideuser block
on the account — and the projection kept serving it. `neowiki-query` is granted to `*` by default, so
an anonymous Cypher query returned a name the wiki no longer shows anywhere.

`PagePropertiesBuilder` now takes the name from `$revision->getUser()`, whose default audience answers
with nobody once the name is hidden. That is what the RDF export path already did; the graph write
path was the outlier. The `?UserIdentity $user` parameter, which every caller derived from the
revision anyway, is gone, so no future call site can reintroduce a raw name.

Withdrawing a name creates no revision, so nothing reprojected the pages holding it. Two paths now do:

- `ArticleRevisionVisibilitySet` reprojects the page, reading its current revision from the primary
  since the change has only just committed.
- `BlockIpComplete` and `UnblockUserComplete` queue a job that reprojects every page the user is the
  last editor of. The queueing is itself deferred: the unblock hook runs before MediaWiki clears the
  flags, and `lazyPush` is not lazy under the command line.

Pages projected before this lands keep the name they were projected with until they are edited or the
graph is rebuilt.

## Notes for review

- Categories are not at stake. MediaWiki refuses to hide the *content* of a page's current revision,
  and only the current revision is projected.
- The job takes no cursor. `$wgHideUserContribLimit` (1000 by default) bounds the page set, though it
  can be disabled; the job is idempotent, so the queue's own retry covers failure.
- The lookup joins on `rev_page` as well as `page_latest`, because `page` carries no index on
  `page_latest` and pairing them alone means a full scan of it.
- ADR 27's consequence "Dumps and projections contain restricted content" now has an exception. That
  may warrant an Update note in the ADR; I have not written one.

## Manual check

Core grants neither right by default, so first:

```php
$wgGroupPermissions['sysop']['deleterevision'] = true;
$wgGroupPermissions['sysop']['hideuser'] = true;
```

Edit a page as some user, then read the graph anonymously:

```
POST /rest.php/neowiki/v0/query/cypher
{"cypher":"MATCH (p:Page) WHERE p.id = <page id> RETURN p.lastEditor AS name"}
```

Hide that revision's author on `Special:RevisionDelete` — the name goes empty at once, and comes back
when you show it again. Then block the author with "hide username" and run the job queue: the name
goes empty on every page they last edited.

## Known gaps

- The work list comes from `page_latest` in the database, but the leak lives in the graph. A page
  whose projection write was lost — the edit path is failure-isolated — is never revisited.
- `Special:MergeHistory` rewrites `page_latest` with no hook at all, so `lastEditor` goes stale there.
  Pre-existing, worth its own issue.
- A rename between queueing and running the job silently reprojects nothing. Core's own
  `RevisionDeleteUser` resolves the actor by name too, but runs synchronously, so it has no window.

> <sub>AI-authored — Claude Code, `Opus 5 (1M context; xhigh)`; one-line ask from @alistair3149, paused before opening for review, then rescoped on request to cover the block route; diff not yet human-reviewed; tests written and run locally, both routes verified end to end on a dev wiki, findings from an automated review round fixed and mutation-tested.</sub>

Two verification caveats a reviewer should know. The automated `code-review` pass did not complete — it died on a session rate limit — so this change has had test, security, design and correctness review but not that one. And the full `@group Database` suite will not run to completion in my local environment (it hangs identically with this change reverted), so the affected classes were run individually and the whole-suite result rests on CI.

</details>

### DIFF
--- a/docs/api/graph-model.md
+++ b/docs/api/graph-model.md
@@ -37,7 +37,7 @@ A single-wiki install has just one `wiki_id`.
 | `namespaceId` | integer | MediaWiki namespace ID of the page (e.g. `0` for the main namespace, `12` for Help) |
 | `creationTime` | datetime | When the page was created |
 | `lastUpdated` | datetime | When the page was last modified |
-| `lastEditor` | string | Username of the last editor |
+| `lastEditor` | string | Username of the last editor, empty where the wiki hides it |
 | `categories` | string[] | MediaWiki categories the page belongs to |
 
 Built-in namespaces have the same ID on every wiki, so `namespaceId` filters consistently across a shared graph.
--- a/extension.json
+++ b/extension.json
@@ -50,6 +50,9 @@
 		"PageDeleteComplete": "ProfessionalWiki\\NeoWiki\\EntryPoints\\NeoWikiHooks::onPageDeleteComplete",
 		"PageUndeleteComplete": "ProfessionalWiki\\NeoWiki\\EntryPoints\\NeoWikiHooks::onPageUndeleteComplete",
 		"AfterImportPage": "ProfessionalWiki\\NeoWiki\\EntryPoints\\NeoWikiHooks::onAfterImportPage",
+		"ArticleRevisionVisibilitySet": "ProfessionalWiki\\NeoWiki\\EntryPoints\\NeoWikiHooks::onArticleRevisionVisibilitySet",
+		"BlockIpComplete": "ProfessionalWiki\\NeoWiki\\EntryPoints\\NeoWikiHooks::onBlockIpComplete",
+		"UnblockUserComplete": "ProfessionalWiki\\NeoWiki\\EntryPoints\\NeoWikiHooks::onUnblockUserComplete",
 
 		"CodeEditorGetPageLanguage": "ProfessionalWiki\\NeoWiki\\EntryPoints\\NeoWikiHooks::onCodeEditorGetPageLanguage",
 
@@ -75,6 +78,10 @@
 	"JobClasses": {
 		"neowikiGraphRebuild": {
 			"class": "ProfessionalWiki\\NeoWiki\\EntryPoints\\Jobs\\GraphRebuildJob",
+			"needsPage": false
+		},
+		"neowikiRebuildLastEditorPages": {
+			"class": "ProfessionalWiki\\NeoWiki\\EntryPoints\\Jobs\\RebuildLastEditorPagesJob",
 			"needsPage": false
 		}
 	},
--- a/src/Application/LastEditorPagesRebuilder.php
+++ b/src/Application/LastEditorPagesRebuilder.php
@@ -1,0 +1,44 @@
+<?php
+
+declare( strict_types = 1 );
+
+namespace ProfessionalWiki\NeoWiki\Application;
+
+use MediaWiki\Title\TitleFactory;
+use MediaWiki\User\UserIdentity;
+use ProfessionalWiki\NeoWiki\Persistence\LastEditorPageIdsLookup;
+use Wikimedia\Rdbms\IDBAccessObject;
+
+/**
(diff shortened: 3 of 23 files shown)


JeroenDeDauw:
@alistair3149 FYI: Doing this kind of stuff, which no one might care about, at this point might be counterproductive as the product is still evolving. We add complexity that in effect is not needed. In part mitigated by AI and not a big deal, but mentioning as I think its quite non-obvious this is possibly net-negative.

PR #1387, opened 2026-09-09
Fold a relation caption heading one row onto that row

alistair3149:
**This narrows the tree half of #1328, which #1345 unified two days ago.** The schema badge leaves the Subject editor's tree navigator entirely, taking `NeoTree`'s `secondaryLabel` and its `secondary` slot, which have no other consumer. Every other surface keeps it, including the pane a row opens.

The tree also printed a caption line above every group of relation targets. In the demo wiki 126 of 135 such groups hold one child, and captions were 37% of all printed lines. A caption heading one row now prints on that row; two or more targets keep the caption line. Where caption and name cannot share a line the row wraps rather than shortening either.

Accepted rather than solved: under ADR 31 a label-less target reads `(unnamed City)` while a labelled sibling under the same relation shows no Schema at all. A row also changes shape as its data changes, since a second target lifts the caption onto a line above and removing one puts it back.

## Manual Browser Check

1. Open `Birth of Johann Sebastian Bach` on the demo wiki and click **Edit** on its infobox. The navigator should print one row per Subject, no caption lines, no schema badges: nine lines become five. The pane on the right still shows the Schema badge.
2. Drag the navigator divider narrow. A row whose caption and name cannot share a line should wrap rather than shorten either; `Took place at Eisenach` stays on one line throughout.
3. Edit a nested Subject so its row shows the unsaved dot, then narrow again; the dot stays beside the name.
4. Open `ACME Inc`, which has two targets under one property: the caption line stays above rows that carry none.
5. Tab in and arrow through the tree: the focus ring sits on the row, which announces its caption, then its name.

> <sub>AI-authored — Claude Code, `Opus 5 (1M context, xhigh)`; opened as a design question from @alistair3149 about which of two labels to drop, settled over four rounds that redirected it three times; then a one-line ask to build it and a further round dropping the badge from the root as well; reviewed by a directed multi-pass review whose confirmed findings were applied, diff not yet human-reviewed; 1956 TS tests, lint and build green locally, browser-verified on a dev wiki, CI green.</sub>

### DIFF
--- a/i18n/qqq.json
+++ b/i18n/qqq.json
@@ -87,7 +87,7 @@
 	"neowiki-subject-editor-save-scope": "Note shown beside the subject editor's save button, naming what the save writes. Parameters:\n* $1 - how many subjects have unsaved changes\n* $2 - how many wiki pages hold them\nShown only when the user cannot see the whole scope already, so $1 is at least two or the one changed subject is not the one on screen.",
 	"neowiki-subject-tree-label": "Accessible name of the structure tree in the subject editor, listing the subjects reachable from the one being edited.",
 	"neowiki-subject-editor-resize-navigator": "Accessible name of the divider between the structure tree and the edit form in the subject editor. Dragging it, or moving it with the arrow keys, changes how wide the structure panel is. The panel itself is named by {{msg-mw|neowiki-subject-tree-label}}.",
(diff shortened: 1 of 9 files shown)


alistair3149:
Only single relations:
image

Both single and multi relations:
image

Wrapping behavior:
image

PR #1389, opened 2026-09-09
Let approval extensions choose the revision NeoWiki publishes

### LINKED ISSUE #1379: Let approval extensions choose the canonical revision NeoWiki reads and projects
NeoWiki reads and projects the latest revision of a page everywhere except the page view. An approval extension
such as [ContentStabilization](https://github.com/wikimedia/mediawiki-extensions-ContentStabilization) shows
readers a stable revision instead, and reaches each data consumer through a seam that consumer exposes —
Semantic MediaWiki's `RevisionGuard`, whose `ChangeRevision` and `IsApprovedRevision` callbacks
ContentStabilization implements behind its own `StabilizeSMWProperties` setting. NeoWiki exposes no equivalent,
so it is the consumer such an extension cannot stabilize.

ContentStabilization is already a NeoWiki consumer: its `RegisterNeoWikiProperties` hook handler registers a
`PagePropertyProvider` through `NeoWikiRegistrar`, and calls `newPageRebuilder()->rebuild( $title )` when a stable
point is added or removed. So the registration mechanism and the re-projection trigger both exist; what is missing
is any way for it to say which revision that rebuild should read.

## What already works, and must not be duplicated

The page view. ContentStabilization's `StabilizeContent::onArticleViewHeader` sets the OutputPage revision id to
the stable revision; `NeoWikiHooks::handleContentPage` compares it against `getLatestRevID()` and passes it to the
main-subject placeholder, and the frontend does the same with `wgRevisionId` / `wgCurRevisionId`, so the Subject
read resolves through `PointInTimeSubjectLookup`. Views hydrate client-side through that same read, so `{{#view}}`
and the auto-rendered main subject are stabilized too.

One limitation of that path is worth recording rather than assuming away: `PointInTimeSubjectLookup` resolves
Subjects hosted on *other* pages by timestamp, taking each one's newest revision at or before the requested
revision's timestamp. Under stabilization that is not the target page's own approved revision, so a relation
target on a stabilized page renders whatever that page happened to hold on the date the viewed revision was
saved.

## What does not

Six surfaces, each needing one of three answers.

**Resolve a revision.** These ask "which revision of this page do I read?" and today always answer "the latest":

- `PageRebuilder::rebuild`, which takes `$wikiPage->getRevisionRecord()`. This is the path ContentStabilization
  already calls, so on stable-point removal it re-projects an unapproved revision instead of falling back to the
  previous stable point or withdrawing the data;
- RDF export: `RdfPageLoader::loadByTitle` does the same;
- the by-id REST read without a `revisionId`: `MediaWikiSubjectRepository::getContentByPageId` takes
  `getRevisionByPageId`;
- Schemas, Layouts, Mappings and wiki config: `PageContentFetcher` resolves by title. Whether these should
  stabilize at all is a real question — a Schema governs validation and rendering wiki-wide — and this issue is
  where it gets answered.

**Skip a revision.** The graph projection and the subject-page index run from `RevisionFromEditComplete` into
`OnRevisionCreatedHandler::onRevisionCreated`, which is *handed* the new revision. There is nothing to resolve; the
question is whether a draft save should touch shared state at all. `RevisionGuard` splits `ChangeRevision` from
`IsApprovedRevision` for exactly this reason.

**Check a viewer.** `GET /neowiki/v0/subject/{id}?revisionId=N` verifies only that the revision's page is readable
(`GetSubjectApi::revisionPageIsReadable`), never that this viewer may see *that revision*. ContentStabilization
refuses a non-reviewer both the page view and the diff of a draft, and its `getUserPermissionsErrors` handler only
fires once a view has been built from an `Article`, which never happens on a REST request. Resolution cannot help
here: naming a revision bypasses it by definition.

## The parse-time accessors need no new seam

`{{#neowiki_value}}` and the `nw` data functions are the sixth surface, and they are the one case MediaWiki
already has a mechanism for. ContentStabilization renders a stabilized page by calling
`RevisionRenderer::getRenderedRevision()` on the stable revision, having first set
`ParserOptions::setCurrentRevisionRecordCallback` so everything in that parse resolves the page to it.
`SubjectResolver` ignores the callback and reads `$wikiPage->getRevisionRecord()`, so a stabilized page renders
stable wikitext carrying draft Subject values.

Reading through `Parser::fetchCurrentRevisionRecordOfTitle()`, which routes to that callback and memoizes per
parse, fixes this for any extension that stabilizes, with no NeoWiki-specific registration. The `subject=` form
needs its hosting page resolved first, which `SubjectResolver` already does.

## Proposal

One registered policy, three methods, one per question above:

```php
interface RevisionPolicy {
	/** The revision shared state is built from. Null when the page has none. */
	public function canonicalRevisionOf( PageIdentity $page ): ?RevisionRecord;

	/** Whether a revision may update that shared state. */
	public function revisionIsCanonical( RevisionRecord $revision ): bool;

	/** Whether a viewer may read a revision they asked for by id. */
	public function revisionIsReadableBy( RevisionRecord $revision, Authority $viewer ): bool;
}
```

The default resolves to the latest, calls every revision canonical, and makes every revision readable, so with
nothing registered nothing changes.

**Only the third method takes an `Authority`, and that asymmetry is the point.** The first two serve shared state —
the graph, the subject-page index, the RDF export, the parser cache — which cannot vary per viewer; letting a
registered policy vary them by user would poison the parser cache per user, the class of problem #1347 just
closed. The third serves one request and nothing else. ContentStabilization's SMW integration takes no user at all,
because SMW's store has no per-request surface to protect.

**Registration**: a single slot on `NeoWikiRegistrar` — two extensions cannot both choose the revision. This is a
new idiom: all six existing registrar methods are `add*` multi-registries, `addGraphDatabasePlugin` included, so
the collision policy needs deciding rather than copying.

**Naming**: "canonical revision" is not in the glossary and `naming.md` reserves `<X>Lookup` for data access, which
this is not. Both need settling; the shape above is a starting point, not a proposed final name.

## Decision this issue must make: what the subject-page index follows

`OnRevisionCreatedHandler` writes `neowiki_subject_page` alongside the projection, and ADR 32 makes that index
authoritative for subject-to-page resolution. So the policy governs it too, and the choice is not free.

Recommendation: **the index follows the canonical revision, like the graph.** One canonical state everywhere is
easier to reason about than an index that tracks drafts while the graph does not, and the alternative produces its
own inconsistency — a Subject removed in a draft would be absent from the index while still present in the graph,
so Cypher would return it and the by-id REST read would not find its page.

The cost, which should be recorded rather than discovered later: a Subject that exists only in a draft is not
indexed, so `getPageIdOfSubject` returns nothing for it. `PointInTimeSubjectLookup` still reads it out of the
requested revision's slot, so the reviewer's draft view keeps working, but `GetSubjectQuery` gets no page
identifiers and `SubjectDisplayName::labelOrPageName` falls back to an empty page name. That is a presentation
follow-up — the point-in-time lookup knows the hosting page it read from and can supply it — not a reason to
reject the recommendation.

## Smaller things to fold in

- `newGetSubjectQueryForRevision` passes the latest-revision `PageSubjectsLookup`, so "is this the main subject" on
  a revision-keyed read is answered from the current revision rather than the requested one.
- Denying a revision read should take the not-found path the endpoint already uses for unreadable pages, keeping
  the existence oracle closed (#1046), rather than a 403.

## Done when

An extension can register a policy; the resolve, skip and viewer-check surfaces above use it; the parse-time
accessors read through the parser's current-revision callback; with nothing registered nothing changes; the
page-view path is untouched; the index decision is recorded in ADR 32 and revision choice classified in ADR 27;
and `extending.md` documents registration.

> <sub>AI-authored — Claude Code, `Opus 5 (1M context)`; rewrite of this issue's original text, requested by @alistair3149 after a review of its claims against the code, then narrowed by him from a three-issue split back to one; text not yet human-reviewed; every claim re-read on master and in the public ContentStabilization source, nothing executed.</sub>

alistair3149:
Fixes https://github.com/ProfessionalWiki/NeoWiki/issues/1379

NeoWiki published each page's latest revision everywhere except the page view, so an approval extension such as
ContentStabilization could stabilize what a reader sees on a page but not what the graph stores or the RDF export
served. It already registers a Page Property Provider with us and calls `newPageRebuilder()->rebuild()` when a
stable point changes; it had no way to say which revision that rebuild should read.

Extensions can now register a `RevisionPolicy` through `NeoWikiRegistrar::setRevisionPolicy()`. It answers three
questions, each asked where it is natural:

- `publishesRevision()` — asked as a revision is written. A save already has its revision and only needs to know
  whether to publish it, so this costs one lookup. Return false and the graph keeps whatever it published before.
- `publishedRevision()` — asked when nothing but the page is known: by `PageRebuilder::rebuild()`, the reprojection
  path an approval extension calls when its answer changes, and on every Schema, Layout, Mapping and configuration
  read and every RDF export. Returns the revision to publish, or `null` when the page has nothing publishable.
  `rebuildFromPrimary()` deliberately does not ask: import and undelete write a revision, so they take the save
  path's semantics.
- `revisionIsReadableBy()` — asked when a caller names a revision by id, which neither publishing method can
  intercept. A refused revision answers exactly like an absent one, so the sequential revision ids stay unsweepable
  (#1046).

Registration is a single slot rather than the `add*` list every other extension point uses, because two extensions
cannot both decide this. A second policy is refused with a warning and the first keeps deciding. With nothing registered every surface reads the latest revision, exactly as before.

The registered policy is wrapped in `FailureIsolatingRevisionPolicy`, as every other extension-contributed plugin is.
A policy that throws publishes nothing and hides everything — the edit hook fires inside `PageUpdater`'s atomic
section, so an uncaught throw there would roll the contributor's save back. It also refuses a revision from another
page, and a revision whose text is suppressed: only the current revision was ever projected before, and core forbids
suppressing that one, so a policy naming an older revision must not start publishing suppressed Subjects.

## The subject-to-page index is deliberately not governed by the policy

It records where a Subject lives, not whether it is published, and it is written from every revision. Every id-keyed
read and write addresses its page through it — `ReplaceSubjectAction`, `UpdateStatementAction`, `DeleteSubjectAction`,
`MoveSubjectAction`, and the check in `CreateSubjectAction` that an id is not already taken — so a Subject left out of
it could not be edited, moved or deleted, and its id would read as free and be re-mintable onto another page. Adding a
Subject in an unapproved revision is the normal case on an approval wiki, so that would fire constantly.

Keeping the index out of it also makes the maintenance rebuilder correct rather than contradictory: it writes from
`page_latest` and needs no knowledge of the policy, which is why it can keep running under `update.php`.

## Which surfaces this covers

The graph projection and the RDF export publish through the policy. Schemas, Layouts, Mappings and the on-wiki
configuration page read through it too, so an unapproved edit to one does not take effect until it is approved; the
Schema editor is unaffected because it loads page source through core's own `/v1/page/` endpoint.

Reads by Subject id stay on the latest revision. Routing them through the policy is what an earlier revision of this
branch did, and it broke the subject editor: the same endpoint is the editor's load path, so the editor loaded
approved values and saved them back over the draft.

## Known gaps, all tracked

- Three Subject read surfaces still serve the latest revision — `action=subjects`, the page-keyed REST read, and
  referenced Subjects on other pages via `PointInTimeSubjectLookup`
  (https://github.com/ProfessionalWiki/NeoWiki/issues/1390). The parse-time accessors are unchanged for a different
  reason: they need the parser's own current-revision callback rather than this interface.
- The RDF export stops answering for a page once approval is withdrawn, but what was published stays in the graph
  until the page is deleted; a rebuild does not remove it (https://github.com/ProfessionalWiki/NeoWiki/issues/1391).
- Schemas and Mappings cache on the page's latest revision id, so an approval change with no page edit does not take
  effect until that page is edited or the entry expires; and a Subject is validated against the published Schema
  while the Schema editor shows the latest one (https://github.com/ProfessionalWiki/NeoWiki/issues/1392).
- Undelete and import take the save path, so a restored page whose latest revision is unpublished comes back
  addressable but absent from the graph until the approval extension reprojects it.

## Considered, omitted

- Collapsing the two publishing methods into one nullable substitution. It reads as a reduction, but it forces every
  draft save to resolve and re-project a revision the graph already holds, and it was what dragged the index and the
  editor into the design.
- Making the index follow the published revision, or hold the union of both. The first breaks editing as described
  above; the second leaves the maintenance rebuilder unable to remove a row, which is the one thing it exists to do.

## Not in this change

The decision itself is not recorded in an ADR. `docs/AGENTS.md` says an ADR is a dated record, not retro-edited
apart from status links, so an earlier revision of this branch that appended to ADR 27 and ADR 32 has been reverted.
Revision choice wants its own ADR.

<sub>AI-authored — Claude Code, `Fable 5.1` after a mid-session model switch from `Opus 5 (1M context)`, which wrote the first design; implemented from #1379 on @alistair3149's go-ahead, redesigned after a five-pass review found the first design broke editing, then hardened after a second six-pass review; diff not yet human-reviewed; phpcs and phpstan clean, nine touched test classes pass locally including one that drives the registered policy through NeoWikiExtension's real wiring into Neo4j, CI on the previous push green across MW 1.43–master.</sub>

### DIFF
--- a/docs/extending/extending.md
+++ b/docs/extending/extending.md
@@ -168,9 +168,70 @@ the Page node. No new revision is created. `rebuild()` returns a `PageRefreshOut
 - `SkippedUnreadableSubjects` — the page's subject slot holds content NeoWiki cannot read as Subjects.
 - `SkippedUnreadablePageProperties` — the page's properties could not be built, for instance because a provider or the
   page's own parse threw.
+- `SkippedUnpublishableRevision` — a registered revision policy publishes no revision of the page.
 
 A graph store that fails is logged and skipped, exactly as on a normal page save; only request timeouts and
 wiki-database errors throw.
+
+### Choosing which revision NeoWiki publishes
+
+By default NeoWiki publishes each page's latest revision: that is the revision it projects to the graph stores and
+exports as RDF. An approval extension shows readers an approved revision instead, and registers a `RevisionPolicy` so
+NeoWiki publishes the same one.
+
+```php
+public function onNeoWikiRegistration( NeoWikiRegistrar $registrar ): void {
+	$registrar->setRevisionPolicy( new MyApprovalPolicy( $this->approvalLookup ) );
+}
+```
+
+```php
+class MyApprovalPolicy implements RevisionPolicy {
+
+	public function publishesRevision( RevisionRecord $revision ): bool {
+		return $this->approvalLookup->isApproved( $revision );
+	}
+
+	public function publishedRevision( RevisionRecord $revision ): ?RevisionRecord {
+		return $this->approvalLookup->lastApprovedRevisionOf( $revision->getPage() );
+	}
+
+	public function revisionIsReadableBy( RevisionRecord $revision, Authority $viewer ): bool {
+		return $this->approvalLookup->isApproved( $revision )
+			|| $viewer->isAllowed( 'my-extension-see-drafts' );
+	}
+
+}
+```
+
+`publishesRevision()` is asked as a revision is written; return false and the graph keeps whatever it published
+before. `publishedRevision()` is asked when a page is reprojected; return the argument to leave the projection alone,
+or `null` when the page has nothing publishable. `revisionIsReadableBy()` is asked only when a caller names a revision
+itself, which neither publishing method can intercept — a revision it refuses answers exactly like one that does not
+exist.
+
+- **A policy answers for the wiki, not for a viewer.** The graph and the RDF export are one state every reader sees.
+- **Only one extension can decide this.** A second policy is refused with a warning in the `NeoWiki` log channel; the
+  first one keeps deciding.
+- **Subject writes read the latest revision**, so a contributor still edits what they last saved. Schemas are the
+  exception: a Subject is validated against the Schema revision the policy publishes, while the Schema editor shows
+  the latest one ([#1392](https://github.com/ProfessionalWiki/NeoWiki/issues/1392)).
+
+Call [`newPageRebuilder()->rebuild( $title )`](#refreshing-a-pages-data-without-an-edit) whenever your extension
+changes which revision it approves; nothing else tells NeoWiki the answer has changed. `publishedRevision()` is also
+asked on every Schema, Layout and Mapping read and every RDF export, so keep it cheap.
+
+Two gaps:
+
+- Schemas and Mappings are read through a cache keyed on the page's latest revision id, so an approval change with
+  no accompanying page edit does not take effect until that page is edited or the cache entry expires
+  ([#1392](https://github.com/ProfessionalWiki/NeoWiki/issues/1392)).
+- The RDF export stops answering for a page once approval is withdrawn, but what was published stays in the graph
+  stores until the page is deleted; a rebuild does not remove it
+  ([#1391](https://github.com/ProfessionalWiki/NeoWiki/issues/1391)). Revocation is not a retraction mechanism.
+
+Subject reads over REST, and the parse-time accessors, are not yet covered
+([#1390](https://github.com/ProfessionalWiki/NeoWiki/issues/1390)).
 
 ### Graph Database Backends
 
--- a/maintenance/DumpRdf.php
+++ b/maintenance/DumpRdf.php
@@ -102,7 +102,7 @@ class DumpRdf extends Maintenance {
 		}
 
 		if ( $page === null ) {
-			$this->error( "Skipped page $pageId: it no longer exists or its subject slot does not hold Subject data" );
+			$this->error( "Skipped page $pageId: it no longer exists, its subject slot does not hold Subject data, or the registered revision policy publishes no revision of it" );
 			return false;
 		}
 
--- a/src/Application/FailureIsolatingRevisionPolicy.php
+++ b/src/Application/FailureIsolatingRevisionPolicy.php
@@ -1,0 +1,104 @@
+<?php
+
+declare( strict_types = 1 );
+
+namespace ProfessionalWiki\NeoWiki\Application;
+
+use MediaWiki\Permissions\Authority;
+use MediaWiki\Revision\RevisionRecord;
+use Psr\Log\LoggerInterface;
+use Throwable;
+
+/**
+ * Holds a registered policy to what NeoWiki can accept from one, and fails closed when it cannot.
+ *
+ * A policy that throws is treated as publishing nothing and hiding everything: the edit hook it runs
+ * on sits inside PageUpdater's atomic section, so a throw there would roll the contributor's save
+ * back, and a throw on a read would take every Schema, Layout and Mapping read down with it. The other
+ * extension-contributed plugins are wrapped the same way; see FailureIsolatingGraphDatabasePlugin.
+ *
+ * Two answers are refused as well as caught. A revision from another page: every caller keys its write
+ * or its export on the returned revision's page id, so accepting one would publish page B under page
+ * A's name, behind A's read gate. And a revision whose text is suppressed: only the current revision
+ * was ever published before this policy existed, and core refuses to suppress that one, so nothing
+ * read a suppressed revision's Subjects — a policy naming an older one must not start to.
+ */
+class FailureIsolatingRevisionPolicy implements RevisionPolicy {
+
+	public function __construct(
+		private readonly RevisionPolicy $policy,
+		private readonly LoggerInterface $logger,
+	) {
+	}
+
+	public function publishesRevision( RevisionRecord $revision ): bool {
+		if ( $revision->isDeleted( RevisionRecord::DELETED_TEXT ) ) {
+			return false;
+		}
+
+		try {
+			return $this->policy->publishesRevision( $revision );
+		} catch ( Throwable $exception ) {
+			$this->logFailure( 'publishesRevision', $revision, $exception );
+			return false;
+		}
+	}
+
+	public function publishedRevision( RevisionRecord $revision ): ?RevisionRecord {
+		try {
+			$published = $this->policy->publishedRevision( $revision );
+		} catch ( Throwable $exception ) {
+			$this->logFailure( 'publishedRevision', $revision, $exception );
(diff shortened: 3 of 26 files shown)


JeroenDeDauw:
Rebased onto master to resolve the conflicts with https://github.com/ProfessionalWiki/NeoWiki/pull/1386, which dropped the `$user` argument from `OnRevisionCreatedHandler::onRevisionCreated()` and from the page-properties source (the revision now supplies its own author). Besides the two textual conflicts, that change reached three spots of this branch that git had merged silently:

- `PageRebuilder` now hands the handler the revision alone, after the policy substitution.
- The two new handler tests call `onRevisionCreated( $revision )` without the user argument.
- The three new `PageRebuilderTest` cases use master's spy, which records a plain list of revisions. The assertion that the published revision brings its own author went with the parameter it tested: the handler now reads the author from whatever revision it is given, so `[ $published ]` being what the handler received covers it.

Committed separately: two `fwrite( STDERR, … )` debug dumps left in `PageRefreshWithoutEditTest`, which printed on every run.

One interaction for the author's eye rather than a fix here. Master added two reprojection callers for https://github.com/ProfessionalWiki/NeoWiki/issues/1246, `onArticleRevisionVisibilitySet` and `LastEditorPagesRebuilder`, and both go through `rebuildFromPrimary()`, which this branch defines as the non-substituting path. With a policy registered, a page whose latest revision is unpublished is handed to the handler as that latest revision, `publishesRevision()` declines it, and the hidden name stays in the graph until the approval extension reprojects the page. No draft content gets published by it, and it matches the undelete/import gap already listed in the description. Flagging it so that coupling "reads primary" to "does not substitute" stays a deliberate choice rather than an accident of the merge.

Verification: `make cs` clean; full local PHPUnit suite green (3233 tests, 10 skipped) before the debug cleanup, with the affected class re-run after it.

> <sub>AI-authored — Claude Code, `Fable 5.1 (max)`; one-line ask from @JeroenDeDauw to resolve the conflicts, no revisions; rebase and adaptations not yet human-reviewed; phpcs, phpstan and the full local PHPUnit suite run locally, CI pending on the new push.</sub>

Issue #1377, opened 2026-09-08
Add server-side filter injection for raw query surfaces


alistair3149:
[ADR 27](https://neowiki.ai/docs/adr/027-access-control) decides that raw query surfaces have whole-store read
semantics, and that "raw queries will support server-side filter injection": a deployment registers scoping
predicates that core applies to every caller-supplied query, for wiki farms sharing one graph such as BlueSpice
Galaxy. Nothing implements it.

In a shared graph every node carries `wiki_id` ([ADR 22](https://neowiki.ai/docs/adr/022-multi-wiki-node-identity)),
but only NeoWiki's own queries use it: the relation-target lookup scopes itself (#1084). A caller-supplied query does
not, so a user holding `neowiki-query` on one wiki, which the default grant gives everyone (#1342), reads every wiki's
projection through the query endpoint.

Surfaces the predicate must reach: `POST /query/cypher`, `POST /query/sparql`, `{{#cypher_raw}}`, `{{#sparql_raw}}`,
`nw.query` and `nw.sparqlQuery`. Since #1346 they all run through `Neo4jQueryService` and `SparqlQueryService`, so
the two services are the place.

To design:

- **The contract.** Either a list of allowed Wiki IDs that core turns into predicates itself, or a callback that
  rewrites the query. The list keeps the rewriting in core, testable once; the Source registry
  ([ADR 23](https://neowiki.ai/docs/adr/023-subject-sources)) already knows the local key and any registered farm
  siblings. Registration through `NeoWikiRegistrar`, like the other registries.
- **The default.** With nothing registered, whether a wiki keeps today's unscoped behaviour or is scoped to its own
  Wiki ID. Scoping by default is secure for farms and changes nothing on a single wiki, whose nodes all carry its id.
- **Application to Cypher.** NeoWiki has no Cypher parser; the validator is a keyword check plus `EXPLAIN`. Options:
  a parser that adds `WHERE n.wiki_id IN $scope` for every node variable; post-filtering rows, which only works when
  rows return nodes rather than projected scalars; or a restricted query shape.
- **Application to SPARQL.** The path forwards queries unvalidated (#1264). Every store receives the same per-page
  named graphs under the wiki's base IRI, which may be a cheaper handle than rewriting, except on stores that fold
  named graphs into the default graph.
- **Failure mode.** A query the injector cannot scope is rejected, never run unscoped.

[...]

PR #1401, opened 2026-09-10
Document configuring NeoWiki for wikis with restricted content

### LINKED ISSUE #1342: Decide the default grant of the neowiki-query right
The `neowiki-query` right is granted to `*` by default. It gates the REST query endpoints, which execute a
caller-supplied Cypher or SPARQL query: the result rows are not attributable to pages and are not trimmed, so the
right carries whole-store read semantics — granting it gives read access to everything the wiki projects into the
store.

To decide: whether that default grant changes, and how deployments holding restricted content are expected to
configure it.

This decision was deferred when [ADR 27: Access Control](https://neowiki.ai/docs/adr/027-access-control) was
accepted.

> <sub>AI-authored — Claude Code, `Opus 5 (max)`, revised by `Fable 5 (max)`; detailed spec from @JeroenDeDauw, revised once on his review; not yet human-reviewed in full; text derived from the published ADR 27 wording, nothing executed.</sub>

alistair3149:
Fixes https://github.com/ProfessionalWiki/NeoWiki/issues/1342

The `neowiki-query` right stays granted to `*`. Gating a read or query surface behind a dedicated right is already the exception in MediaWiki, and a narrower default would break every wiki that runs raw queries at parse time rather than protect the exposed ones: the canonical parse runs as the anonymous user, so withholding the right from `*` denies raw queries in every save-time parse, job-queue parse and Parsoid render, restricted content or not.

What was missing was the other half of the question — how a deployment whose readers do not all see the same pages is expected to configure NeoWiki. That is a new operations page, with its sections in the order the operator decides.

Two facts had no home and now have one: that MediaWiki refuses every endpoint on a login-required wiki before NeoWiki's own checks run, and that a `neowiki-query` holder reads everything the wiki projects into the store.

`docs/authoring/parser-functions.md` promised without qualification that "what one reader may see does not reach another through the parser cache". That is false wherever page access is not group-derived, so it is corrected here rather than filed: same surface, same contract.

ADR 27 is left as it stands. Its Open decisions list records what was open at acceptance, which stays true of that date; the resolution lives on the issue it defers to.

## Scope decision recorded here

NeoWiki decides page `read` from the reader's effective groups plus the two wiki-level rights, and that is what the parser-cache key follows. A wiki deciding `read` per individual account, per IP or per session is outside what NeoWiki supports. The page states that boundary and points such a wiki at `{{#view}}`, which fetches per viewer over REST and performs no parse-time read. https://github.com/ProfessionalWiki/NeoWiki/issues/1405 stays open as the discussion record, with the expected disposition being to decline.

Relates to [ECHOLOT FR2.2](https://github.com/ECHOLOT-ECCCH/echolot-WP2/issues/19), which asks for flexible permission levels for different users and user groups. This PR does not build anything toward it; it documents which levers exist today and where each one stops, which is the starting point that requirement needs.

## Considered, omitted

- **`$wgParserCacheType = CACHE_NONE`.** The old wording in `installation.md` prescribed it, and it is deliberately not carried over. It disables every parser cache on the wiki, Parsoid's included, to close a leak confined to NeoWiki-derived output on a configuration NeoWiki does not support. The page names the boundary and the `{{#view}}` alternative instead.
- **How `$wgGroupPermissions` and `$wgRevokePermissions` behave.** MediaWiki's own manual owns rights semantics, and NeoWiki grants `neowiki-query` to `*` alone, so clearing it there is the whole recipe. Linked rather than restated.
- **Cross-wiki guidance.** A shared graph store makes `neowiki-query` a whole-farm read, but nothing in NeoWiki implements per-wiki scoping today (https://github.com/ProfessionalWiki/NeoWiki/issues/1341, https://github.com/ProfessionalWiki/NeoWiki/issues/1377), so the page would have described a problem with no lever.
- A recipe for un-configuring the Neo4j URLs to remove the Cypher surfaces. It contradicts the neighbouring statement that the right cannot be scoped per query language, and no operator drops their graph backend to fix permissions.
- The access-class and parser-cache mechanism, the per-endpoint read-gate contract, and the default-groups table. ADR 27, `rest-api.md` and `query-api.md` own those.
- The rationale for keeping the default. That belongs on the issue, not in operator docs.
- Naming the `temp` group in the recipe. Its only application is the hand-back case, where a temporary account silently not receiving a whole-store query right is the wanted outcome.

> <sub>AI-authored — Claude Code, `Opus 5 (1M context, xhigh)`; open question from @alistair3149 across a long analysis session, with the decision to keep the default, the scope call on per-account read, and the shape of the docs all his; text not yet human-reviewed; every claim re-derived from master and MediaWiki core source, a fresh-context blind judge and a fresh-context text review each found defects that are fixed, `DocsPathReferencesTest` green.</sub>

### DIFF
--- a/docs/operations/restricted-content.md
+++ b/docs/operations/restricted-content.md
@@ -67,6 +67,6 @@ Whether or not you narrow it:
 
 Restricting a page removes nothing from a store, and a rebuild reprojects it, so a store that has once held
 restricted content keeps holding it. Do not expose a SPARQL store directly on a wiki with restricted content, and
-treat a [bulk dump](../rdf/rdf-export.md#bulk-dump) and any store backup as readable by whoever can reach it.
+treat a [bulk dump](../api/rdf-export.md#bulk-dump) and any store backup as readable by whoever can reach it.
 
 The model behind all of this is [ADR 27: Access Control](../adr/027-access-control.md).
--- a/docs/operations/restricted-content.md
+++ b/docs/operations/restricted-content.md
@@ -1,0 +1,72 @@
+---
+title: Restricted content
+order: 5
+---
+
+# Restricted content
+
+What NeoWiki restricts, and what it does not, on a wiki where some content is readable only by certain users or
+groups. Page protection and `$wgNamespaceProtection` do not restrict reading; see
+[Permissions](../api/rest-api.md#permissions).
+
+## What limits each surface
+
+Enforced per page:
+
+| Surface | What limits it |
+|---|---|
+| REST reads | The page's `read`. |
+| REST writes | The page's `edit`. |
+| `{{#view}}` | The source page's `read`. |
+| `{{#neowiki_value}}` and the `nw.` accessors | The parsing user's `read`. |
+
+Enforced wiki-wide, or not at all:
+
+| Surface | What limits it |
+|---|---|
+| `POST /query/cypher`, `POST /query/sparql` | The caller's [`neowiki-query`](../api/query-api.md#permissions). |
+| `{{#cypher_raw}}`, `{{#sparql_raw}}`, `nw.query`, `nw.sparqlQuery` | The parsing user's `neowiki-query`. |
+| Graph-store status and rebuilds | The caller's `neowiki-admin`. |
+| Subjects from a [Subject Source](../extending/subject-sources.md) | Nothing. They have no page here to authorize against, so a Source must serve only what every reader may see. |
+| Graph stores, RDF dumps and their backups | Nothing. They hold restricted content in full. |
+
+## Choosing who may query
+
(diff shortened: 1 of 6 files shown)


docs/operations/restricted-content.md
```diff
@@ -0,0 +1,70 @@
...
+Grant it to whichever groups should keep it, and check the result on `Special:ListGroupRights`. OAuth consumers
+and bot passwords need no separate change.
+
+## What readers see
+
+After narrowing `neowiki-query`:
+
+- A page using `{{#cypher_raw}}` or `{{#sparql_raw}}` shows an error box where the results were, to anyone who no
+  longer holds the right.
+- A page using `nw.query` or `nw.sparqlQuery` shows a script error and joins the wiki's pages-with-script-errors
```
alistair3149, line 51: Unrelated to this PR: This is smelly since permission errors should not result in script errors, filed as #1411

PR #1406, opened 2026-09-10
Create a Subject together with the page it lives on, and ask which page it goes on

JeroenDeDauw:
Follows-up to https://github.com/ProfessionalWiki/NeoWiki/pull/1374. Serves the entity-first creation asked for in https://github.com/ProfessionalWiki/NeoWiki/issues/1355.

## What it does

**The creator asks what the Subject is called, and keeps where it goes out of the way.** The optional Subject label comes first, then the statements. Below them, a collapsible **Store the subject on** section sits above the Edit summary one, collapsed by default with its current choice in the header ("a new page", "this page", the page picked). Inside: with a host page **This page** (default), **Another page** and **A new page**; from Special:CreateSubject and Schema pages **A new page** (default) and **An existing page**. The two middle options are the same existing-page picker, worded relative to the host page or without one. Under "A new page" an optional **Page title** field, with the entered label as its placeholder and a help text saying what an empty title falls back to: the label, or the Subject's id. An error at either field opens the section.

**A new page and its Main Subject are created in one revision** by the new `POST /neowiki/v0/subjects` (body: `schema`, `statements`, optional `label`, `pageTitle`, `comment`). The title is `pageTitle` when given, else the label when it is a valid main-namespace title, else the Subject's id. An explicit `pageTitle` that cannot be a main-namespace title answers `400` naming it; a taken title answers `409` naming it; no title is invented. `403` is answered for a title the caller may not create, deliberately before any page is looked for, so the response does not reveal whether an unreadable page exists. `201` returns the Subject and page ids and the title. Adding to an existing page keeps using the per-page endpoints.

**A label-less Main Subject on a page titled with the id of any Subject stored on that page is shown as "(unnamed Person)"**, not the id, in the UI, the graph and the RDF projections. Any Subject's id, so a Subject moved onto or promoted on such a page is covered. Recognition compares the title with its first letter lower-cased, since MediaWiki capitalises titles (`SEpfw…` for `sEpfw…`). No existing page qualifies, so nothing needs re-projecting. ADR 31, the glossary and the API docs are amended here with the exception; ADR 27 records that a title-keyed create reveals to a caller allowed to create a title whether it is taken, as MediaWiki's own page creation does.

Also: **Create Person** on `Schema:Person` opens the creator in place; `Special:NewSubject` is an alias of Special:CreateSubject; the existing-page picker opens its menu on the first character typed.

## Worth challenging

**Collapsed by default instead of a setting to hide it.** A wiki-level switch can follow if a user group still stumbles over the section.

**The label is the name; the page title is derived from it once.** A page move does not rename a Subject created with a label this way; editing the label does. A label-less Main Subject keeps following its page name (ADR 31).

**Page creation through the Subject slot does not fire `EditFilterMergedContent`**, so AbuseFilter and similar extensions do not see it. Every NeoWiki slot write shares this.

## Considered, omitted

- A third radio for "a new page with a title of my choosing": one decision, where to store, split in two.
- An exact-match check while typing; the `409` on save covers it.
- Recording "titled by the system" as data instead of recognising the id from the page's Subjects.
- Wiring the move dialog onto the new endpoint.

## Screenshots (UI, Playwright)

| | |
|---|---|
| **`Schema:City`.** The section collapsed, header "Store the subject on a new page". | [SCREENSHOT: nw-1406-schema-page-new-page.png] |
| **The section open.** "A new page" and the Page title field with its help text. | [SCREENSHOT: nw-1406-store-section-open.png] |
| **From a page's tools.** "This page" is the choice. | [SCREENSHOT: nw-1406-on-page-this-page.png] |
| **An existing page picked.** The note names the outcome. | [SCREENSHOT: nw-1406-another-page-picked.png] |
| **Label "Amsterdam", a new page.** The `409` at the Page title field. | [SCREENSHOT: nw-1406-title-taken.png] |
| **The page created for a label-less City.** Shown as "(unnamed City)". | [SCREENSHOT: nw-1406-created-unnamed.png] |

## Testing

PHPUnit covers the endpoint's statuses and title precedence, the one-revision write and the display rule; vitest covers the section (collapsed by default, header per choice and mode, opened by errors), the Page title field, each save path, the previews and notes, and the picker's existing-only mode with the move dialog unchanged.

> <sub>AI-authored — Claude Code, `Opus 5 (max)` implementation, `Fable 5.1 (max)` review and description; spec from @JeroenDeDauw's review of the base PR and design sessions with him, three rounds of review fixes applied; diff not yet human-reviewed; phpcs, phpstan, the full PHPUnit suite, vitest, the TS build and lint pass locally, CI green, Playwright walk-through of the flows shown.</sub>


JeroenDeDauw:
Add some enhancements on top of this in the next 2 hours

Issue #904, opened 2026-06-21
Add ability to display incoming/inverse relations


JeroenDeDauw:
When a Subject points to another Subject via a Relation (e.g. London `is capital of` UK), only the source side shows that connection. Visiting UK's page gives no indication that London points to it.

Some users expect the reverse to be visible without having to author a Cypher query or drop a view onto each page. Writing a query is already possible and is explicitly **not** the capability being requested here.

## What we want

By default, incoming relations should not be displayed. Wiki admins should be able to change this default globally / per wiki. Finer-grained controls at the Schema and/or View level might also be useful.

When incoming relations are shown, they should be displayed in such a way that a non-technical reader understands the relationship.

## Context

- Raised concretely by Matej Đurčo (ECHOLOT / OEAW) in the 2026-04-17 WP4 meeting. Example he gave: "London is capital of UK" should surface on UK's page as the reciprocal.
- Design position taken in #630 section 3 is that storage stays one-directional and inverse visibility is a render/query concern. This issue is about delivering that capability as a default, built-in behaviour rather than a DIY pattern.
- Readable rendering likely benefits from inverse labels being configurable on Relation Property definitions. Whether that's in scope here or spun off is an implementation decision.

JeroenDeDauw:
The 2026-08-26 design discussion around the relation-target editor (https://github.com/ProfessionalWiki/NeoWiki/issues/971) identified this issue's editing-side counterpart, which had not been recorded anywhere yet: with CIDOC-CRM-style modelling the meaningful Subjects point *at* the one being edited — a birth event references its person, compositions reference their composer — so the subject editor offers nothing to reach or add them from. Displaying incoming relations (this issue) does not by itself let you edit or create them.

What came out of that discussion:

- **Sketched starting point:** an "add incoming relation" affordance in the subject editor — a lookup over the Schemas whose relation properties can target the current Subject's Schema, creating the referencing Subject with its back-reference pre-filled. This is always one hop, since the referencing Subject points directly back at the one being edited.
- **Schema gap:** nothing lets a Schema declare which incoming relation types it surfaces by default (a Person surfacing its birth event, but not every role Subject that happens to reference a person). Where that relevance line sits is unresolved — sometimes you do want to fix such a role from the person's editor.
[...]

PR #1408, opened 2026-09-10
Add Special:Subject, a Subject-centric view

JeroenDeDauw:
For https://github.com/ProfessionalWiki/NeoWiki/issues/1355

`Special:Subject/<id>` shows one Subject on its own, with the row UI a page's Data tab already has.

## What it does

- **`Special:Subject/<id>`** renders the Subject as a Data tab row, expanded on arrival: statements, the edit and
  delete controls, copy link, the overflow menu, and a footer with the id, the concept URI, the page it is stored on
  and the export menu. Under **Referenced subjects**, every Subject its relations point at gets the same row,
  collapsed. The H1 and browser title are the Subject's display name.
- **The concept URI resolver** (`GET /neowiki/v0/entity/{subjectId}`) sends browsers there instead of to the Data
  tab row, and redirects an id qualified with this wiki's own Source key to its bare form. RDF content negotiation
  is unchanged.
- **Data tab rows gain "Open subject"**, a row action leading to the Subject's own page: the one way a reader
  reaches this page from the wiki's UI.
- Reads go through the existing permission-enforced Subject read, for the title as for the body: a Subject on a
  page the reader may not read renders exactly like one the wiki never minted
  (https://github.com/ProfessionalWiki/NeoWiki/issues/1046).
- Left out as page-level operations: pin, reorder, add, and move, which needs the Main Subject flag the Subject
  read does not carry.

## Under the hood

- The Data tab's two inline row templates become `SubjectRow`, and its delete dialog `SubjectDeleteDialog`; the
  Data tab is unchanged to look at and to use.
- `$wgNeoWikiDereferenceSubjectsToDataTab` becomes `$wgNeoWikiDereferenceSubjectsToHostingPage`, default `false`:
  `true` is the plain-page opt-out that existed before; the Data-tab target is gone. A clean rename of the setting
  and of its `MediaWiki:NeoWiki` key, no shim: a `LocalSettings.php` still setting the old name takes the new
  default without notice.
- The repository throws `SubjectNotFoundError` for a Subject the wiki serves none of, so the page can tell that
  from a failed request.
- The referenced Subjects' own relation targets are fetched before the first paint, because `RelationDisplay`
  resolves a target from the store once, synchronously, and an unseeded one would show as an error inside those
  rows. Cost: one read per distinct Subject two relation hops away that the Subject read did not already return;
  three reads in total for the demo Subject with eight referenced Subjects, unbounded for a Subject whose
  referenced Subjects carry many relations.
- PHPUnit covers the special page (mount point, malformed and missing id, the title in all four states, the config
  variables) and the resolver under both settings and through the on-wiki configuration page; vitest covers the
  row, both pages, the not-found error and the Data tab's unchanged behaviour.

- The requested row keeps the blue emphasis the Data tab gives its Main Subject row; here it marks the Subject the
  page is about, and the row stays collapsible.
- `Special:Subject` is listed on Special:SpecialPages, where opening it without an id shows an error asking for
  one.
- One server-side Subject read per view for the title; the frontend reads again for the body.

## Known gap

The editor dialog reads page-scoped `mw.config` values (`wgArticleId` for edit notices, `wgIsProbablyEditable` for
the unregistered-property-type notice), which are 0 and false on a special page, so neither notice appears here.
The fix belongs in the dialog; not filed as an issue yet.

## Manual browser check

- `Special:Subject/s1demo1aaaaaaa1`, logged in: the tab reads "ACME Inc"; edit, delete, copy link, the overflow
  menu, and the footer's id, IRI, page link and export menu work on the open row and on a referenced row opened by
  a header click; **Open subject** on a referenced row, and on a Data tab row, leads to that Subject's page.
- `curl -sI -H 'Accept: text/html' …/rest.php/neowiki/v0/entity/s1demo1aaaaaaa1` answers 303 to
  `Special:Subject/s1demo1aaaaaaa1`; with `Accept: text/turtle`, the Turtle export as before.
- `Special:Subject/nope` shows the error; `Special:Subject/s1zzzzzzzzzzzz9` the not-found message.

## Considered, omitted

- In-place editing on this page; incoming relations (https://github.com/ProfessionalWiki/NeoWiki/issues/904);
  creating a Subject here; relation links inside the row leading to Special:Subject rather than to the page; a
  user-guide entry.
- Extracting the editor and delete wiring this page shares with the Data tab and the infobox into composables;
  moving the shared components out of `components/SubjectsManager/`.
- A cheaper fetch for the nested relation targets; the real fix is a reactive `RelationDisplay`.

> <sub>AI-authored — Claude Code, `Fable 5.1 (max)` orchestrating `Opus 5 (max)` implementers; detailed spec from @JeroenDeDauw, redirected three times mid-task; diff not yet human-reviewed; full PHPUnit and vitest suites, phpcs, phpstan, eslint, stylelint and the TypeScript build green locally and on CI, some seventy production-code mutations each confirmed to fail their tests, both pages driven in a scripted browser against the dev stack.</sub>

Design decisions and this description by `Fable 5.1 (max)`; the code by `Opus 5 (max)` subagents, reviewed by further subagents (code, security, tests, design) and by the orchestrator. The three redirections: the Data tab's row UI instead of a bespoke layout, the page link into the row footer, and "Open subject" as a row action instead of a link in the name. The in-browser check ran as a scripted Playwright session in a container, the interactive browser being held by another session.

</details>

### DIFF
--- a/docs/api/rdf-export.md
+++ b/docs/api/rdf-export.md
@@ -18,7 +18,7 @@ For an end-to-end example comparing the native and ontology-mapped output, see t
 | Setting | Default | Purpose |
 |---|---|---|
 | `$wgNeoWikiRdfBaseUri` | the wiki's canonical URL (`$wgCanonicalServer`) | Base URI under which all NeoWiki IRIs are minted. |
-| `$wgNeoWikiDereferenceSubjectsToDataTab` | `true` | Whether a browser dereferencing a Subject IRI lands on the hosting page's Data tab row (`true`) or the plain page (`false`). |
+| `$wgNeoWikiDereferenceSubjectsToHostingPage` | `false` | When `true`, sends a browser dereferencing a Subject IRI to the Subject's hosting page instead of to `Special:Subject`. |
 
 ## IRI scheme
 
@@ -149,12 +149,11 @@ content-negotiates it and answers `303 See Other` with an absolute `Location`:
 |---|---|
 | `application/trig` | the Subject's TriG RDF (`.../subject/{id}/rdf?format=trig`) |
 | `text/turtle` | the Subject's Turtle RDF (`.../subject/{id}/rdf?format=turtle`) |
-| `text/html`, `*/*`, absent, anything else | the Subject's hosting page |
+| `text/html`, `*/*`, absent, anything else | `Special:Subject/{id}` |
 
 TriG wins when both RDF types are acceptable; the RDF redirects use the native projection.
 
-The default HTML target is that page's Data tab (`/Example_Page/subjects`) opened on the Subject's row.
-When `$wgNeoWikiDereferenceSubjectsToDataTab` is set to `false`, the target is the wiki page itself (`/Example_Page`).
+When `$wgNeoWikiDereferenceSubjectsToHostingPage` is `true`, the HTML target is the Subject's hosting page instead.
 
 The negotiator is always reachable at the REST path, which needs no server configuration:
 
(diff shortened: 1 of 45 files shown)


JeroenDeDauw:
Product-level reviewed/iterated and LGTM. Did not review technical approach yet.

image

alistair3149:
I tested the UI and it works as it is.
- #1418 includes some minor fixes to the current PR
- #1420 includes my opinionated review on the UI, where I moved the navigate button to the subject label

PR #1415, opened 2026-09-11
Skip a View whose Subject does not load instead of blanking the page

### LINKED ISSUE #1414: A View of a Subject the viewer may not read blanks every View on the page
On a page that holds several Views, one `{{#view}}` of a Subject the viewer may not read takes every View on the
page down with it: nothing mounts, the readable ones included.

`NeoWikiApp.vue` loads the Subjects of all placeholders in one `Promise.all` and assigns `viewsData` only after it
resolves. `StoreStateLoader.loadForSubject` awaits `getSubjectWithReferencedSubjects()` with no catch, and the
not-found answer the REST read gives for an unreadable Subject (it answers as if the Subject were absent, per
`rest-api.md#permissions`) rejects it, so the rejection propagates to the app and `viewsData` is never set.
`Infobox.vue` already renders nothing for a null Subject (`v-if="subject !== null"`) but its `subject` computed
reads `subjectStore.getSubject()`, which throws on a miss, and carries `// TODO: handle not found`.

The documented contract ([Restricted content](https://neowiki.ai/docs/operations/restricted-content)) is that the
denied View renders nothing. Expected: a failed Subject load skips that View and the others mount; a View whose
Subject is absent from the store renders nothing rather than throwing. The same pattern is in RedHerb's
`RedHerbCard.vue`, which reads the store the same way.

Found while deriving the restricted-content page from the code for
https://github.com/ProfessionalWiki/NeoWiki/pull/1413; traced in the source, not reproduced in a browser.

> <sub>AI-authored — Claude Code, `Fable 5.1 (max)`; from a blind documentation draft's code reading, filed on @JeroenDeDauw's go-ahead for follow-up changes; not yet human-reviewed; traced in `NeoWikiApp.vue`, `StoreStateLoader.ts`, `SubjectStore.ts` and `Infobox.vue` on master, not executed.</sub>

JeroenDeDauw:
Fixes https://github.com/ProfessionalWiki/NeoWiki/issues/1414

One `{{#view}}` of a Subject the viewer may not read took every View on the page
down with it: that read answers as it does for an absent Subject, its rejection
escaped `StoreStateLoader`'s `Promise.all`, and `NeoWikiApp` never assigned
`viewsData`.

The loader now loads each Subject on its own, skipping one that does not load and
logging its id and the reason. It reads the Schema before storing anything, so a
Subject whose Schema does not load is skipped whole rather than stored without
one. `NeoWikiApp` then mounts only the Views whose Subject the store holds, so a
View Type still finds its Subject where `docs/extending/view-types.md` says it is;
the placeholder of a skipped View stays empty. `Infobox` keeps its own guard,
since the public API lets a host mount it for any id and RedHerb's subject finder
does. That finder, whose error path the no-longer-rejecting loader had left dead,
now asks the store and notifies as before.

A failed request is skipped like a denial, since neither may blank a page. On a
current-revision view the logged error tells them apart: the read answers a denial
with the same `SubjectNotFoundError` as an absent Subject, and a transport failure
with its own. On an old-revision view a denied read is a `404`, which the client
reports like a transport failure.

Verified against a dev wiki with a hook denying anonymous read of one page. Before
the fix, a page with two Views rendered neither and the failed load's error stood
in the console. With the change, a page holding four Views — an infobox and a
RedHerb card of a readable Subject, and the same two of the restricted one —
renders the readable pair, leaves both restricted placeholders empty, and logs
the skip.

Considered, omitted:

- Handling the missing Subject inside the View components instead. It puts the
  null case in every View Type — ours, RedHerb's, and the ones extension authors
  write against the documented contract that the Subject is read from the stores.
  Deciding it once, where the Views are mounted, keeps that contract; `Infobox`
  is the exception above, and only because it is public API.
- A distinct error type for the loader to branch on. Both outcomes are skipped
  either way, and the error it catches already distinguishes them in the log.
- Reporting the skipped ids back from `loadSubjectsAndSchemas`. Only RedHerb's
  subject finder asked the question, and the store answers it.
- Any visible marker where a skipped View was. The documented contract is that the
  denied View renders nothing.

> <sub>AI-authored — Claude Code, `Opus 5 (max)`; implemented from a written spec by a Fable 5.1 session on @JeroenDeDauw's go-ahead, and reshaped on the review's design verdict; diff not yet human-reviewed; each test was written first and seen failing, the full vitest suite (2163 tests), eslint/stylelint and the vue-tsc/vite build are green, and the behaviour was re-checked in a headless browser against a dev wiki with anonymous read denied on one page.</sub>


alistair3149:
Review PR opened at #1441

PR #1420, opened 2026-09-11
Make a Subject's name the way to its own page, and match the Data tab's section spacing

alistair3149:
Follows-up to https://github.com/ProfessionalWiki/NeoWiki/pull/1408. Stacked on #1418, which holds review fixes to
#1408; the name commit below builds on the one that makes the overflow menu's "Open subject" work from the keyboard,
so on #1408 alone that item would stay dead. The two commits here are independent of each other.

- **The Subject's name is the link to its own page.** On rows that name a Special:Subject page, the name links there,
  in Codex link colours without the visited one, and an arrow slides in beside it on hover or keyboard focus. Only the
  Subject a page is already about keeps a plain name. The inline "Open subject" button goes; the overflow menu keeps
  the item, with the same arrow.
  This reverses your choice of a row action over a link in the name. The button only appeared on hover, 190–460px from
  the name, and its right chevron read like the row's disclosure chevron and like the export menu's "opens a nested
  list" chevron. The name does not take the header's click: like the Schema badge beside it, it stops its own, and
  everywhere else in the header still toggles the row. The summary stays the toggle, and the link is one more
  focusable control inside it, as that badge already was.
- **Special:Subject spaces its sections like the Data tab.** Neither page's heading margin ever applied, because the
  skin's `.mw-body h2` outranks a single class. Special:Subject now uses the Data tab's list margins instead, and the
  heading stays the skin's.

## Manual Browser Check

1. On `Special:Subject/s1demo1aaaaaaa1`, hover a referenced Subject's name: it underlines and an arrow slides in.
   Clicking it opens that Subject's page.
2. Click the same row's header away from the name and the Schema badge: the row expands.
3. Tab to that name: the focus ring and the arrow show, and Enter follows the link.
4. On `Rijksmuseum/subjects`, the Main Subject's and the other rows' names link the same way, and no inline
   "Open subject" button remains.
5. The "Referenced subjects" heading sits as far from the rows around it as "Other Subjects on this Page" does on the
   Data tab.

## Considered, omitted

- A shared Subject list component: the Data tab's drag and drop needs its two lists as they are, and their styles had
  not drifted.
- Restyling the skin's h2 so the spacing holds on every skin; on Timeless its own heading margin wins on both pages.

> <sub>AI-authored — Claude Code, `Opus 5 (ultracode)`; directed by @alistair3149 while reviewing #1408, over roughly six rounds of correction; diff not yet human-reviewed, though the rendered result was checked in a browser by @alistair3149; each change test-first or confirmed by a surviving-mutant check, full vitest suite, eslint, stylelint and the build green locally, and the rest, hover, focus and spacing states measured in a browser against a dev wiki, with right-to-left, touch and night mode covered by the review passes below; no PHP touched.</sub>

Reviewed before opening by six independent passes — code review, security, tests, design, accessibility and i18n, and
cross-skin CSS — plus adversarial verifiers for each blocking claim. What they confirmed is folded into these two
commits: `:visited:hover` and `:visited:active` had to be qualified, because Vector paints `a:visited:hover` at the
same specificity and wins on source order; a test now pins that a click on the name keeps its default action; and the
emphasized row's blue label rule went, since after this change blue means "link".

</details>

### DIFF
--- a/i18n/qqq.json
+++ b/i18n/qqq.json
@@ -391,5 +391,5 @@
 	"neowiki-special-subject-referenced-heading": "Heading on [[Special:Subject]] above the Subjects the displayed Subject's relations point at.",
 	"neowiki-special-subject-not-found": "Shown on [[Special:Subject]] in place of the Subject when the wiki has no Subject with that ID. It is also shown for a Subject on a page the reader may not read, so the translation must not mention permissions.\n\nParameters:\n* $1 - the Subject ID that was asked for",
 	"neowiki-special-subject-load-error": "Shown on [[Special:Subject]] when loading a Subject failed with a network or server error: in place of the Subject on first load, and as a notification when a re-read after a change failed. Distinct from {{msg-neowiki|neowiki-special-subject-not-found}}, which says the wiki has no such Subject.",
-	"neowiki-managesubjects-row-open": "Aria/button label for the row action (an inline button on wide viewports, an overflow-menu item on narrow ones) that goes to [[Special:Subject]], the page showing that one Subject. It navigates; it does not expand the row."
+	"neowiki-managesubjects-row-open": "Label for the item in a Subject row's overflow menu, which replaces the row's inline actions on narrow viewports. It goes to [[Special:Subject]], the page showing that one Subject, which the Subject's name links to as well. It navigates; it does not expand the row."
 }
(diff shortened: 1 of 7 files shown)


alistair3149:
https://github.com/user-attachments/assets/b6b9c4fb-a562-4a56-8d06-d69d108e4aac

JeroenDeDauw:
LGTM, tested UI locally. Am rebasing and merging this

Issue #1327, opened 2026-08-31
Collapse the subjects editor tree by default, prepopulated one level deep


JeroenDeDauw:
For https://github.com/ProfessionalWiki/NeoWiki/issues/971. Follow-up to https://github.com/ProfessionalWiki/NeoWiki/pull/1323, which ships the tree fully expanded with no collapse control.

Direction from the 2026-08-31 design review:

* The tree starts collapsed but prepopulated one level (maybe two) deep, so opening the editor shows the subject being edited without an unexplained tree, while expansion is instant.
* It expands when the user hits the expand control or clicks a relation's edit icon, so both newcomers and power users are served.

Bounding belongs to the same design area: on a realistic graph an unbounded tree eventually pulls in the whole wiki, so expansion needs a depth limit — default one level, expanded on demand, per the design state on https://github.com/ProfessionalWiki/NeoWiki/issues/971. Where the limit sits needs feedback from real use and may end up customizable; the API side already plans depth and node caps (https://github.com/ProfessionalWiki/NeoWiki/issues/1324).

> <sub>AI-authored — Claude Code, `Fable 5 (max)`; filed on @JeroenDeDauw's go-ahead from the 2026-08-31 sync follow-up list, scope set by him; text not human-reviewed; content checked against PR 1323's body and the meeting record.</sub>

JeroenDeDauw:
@alistair3149 thoughts? I notice you skipped this one while it seems one of the more urgent things to me

alistair3149:
Yes this one needs some design work compared to the other subject editor works so I deferred it until now.
Will start working on it.

alistair3149:
The frontend is done in #1425. I am still deciding on whether the tree should be collapsed initially and how to implement it.
For the population mechanics, it is blocked by #1324 because the API will determine its shape.

JeroenDeDauw:
That's actually quite different from what I had in mind and does not solve my concern.

Most of the time people want to edit a single subject. Editing multiple is a secondary capability. Yet we always show the tree view. It's also shown "before" the subject editing UI itself since it's to the left. Which is the sensible position when it is shown, but compounds the issue. 

Imagine being a new user. You want to edit a subject. You now have to understand what this tree on the left is and somehow realize you are supposed to ignore it.

[...]

Issue #1324, opened 2026-08-27
Add a relations endpoint for the structure tree and graph views


alistair3149:
A Subject's neighbourhood is only readable one Subject at a time. Every surface that shows related
Subjects — the subject editor's structure tree (#971), a relation field's labels, anything listing what a
Subject points at — reads and parses a whole hosting page per node to recover the four fields it displays,
and the tree does it over HTTP, once per node per level.

Once the schema, label and Main Subject flag sit on the subject-page index (#1333) and relations are indexed
natively (https://github.com/ProfessionalWiki/NeoWiki/issues/1039), a whole neighbourhood is answerable in
roughly `2D+1` indexed queries at depth `D` with no slot read at all. Whether the incoming direction is
answered that way or from the graph store is that issue's open decision; the outgoing direction does not
depend on it.

## Proposal

`GET /neowiki/v0/subject/{id}/relations?depth=N&direction=outgoing|incoming|both` — defaults `depth=1`,
`direction=outgoing`; the root is the first node.

```json
{
	"nodes": [ { "id": "s…", "displayName": "…", "schema": "Person", "page": { "id": 42, "name": "…" }, "truncated": false } ],
	"edges": [ { "source": "s…", "property": "Birth place", "target": "s…", "relationId": "r…", "ordinal": 3 } ],
	"truncated": false
}
```

- Every relation-valued statement is an edge, and no Schema is read by either side. Edges are grouped by
  source in traversal order, then by the Subject's stored statement order. That is the Schema's property
  order for anything the subject editor saved, and drifts only after a Schema reorder, an import, or a
  per-statement write.
- `relationId` and `ordinal` address the statement value the edge came from, since one property holds
  several targets.
- Both directions read the same index rows, so a relation cannot be present one way and absent the other.
- Per-page read gating as on the Subject reads. An unreadable or nonexistent target is absent from `nodes`
  while its edge stays, so a client can tell an unreadable target from no relation — the red-link case
  (ADR 28). An unreadable *source* is the opposite case and drops its edges, or that Subject's statements
  leak. Unreadable nodes take no node budget and no total is reported (#1062).
- One node per Subject however many paths reach it. `truncated` on a node means it reaches Subjects this
  response does not hold; the top-level flag means a cap fired. Unfolding per path stays a client decision.
- Nodes are summaries, not Subjects: they stay out of the `SubjectStore` registry (ADR 30).

[...]

PR #1425, opened 2026-09-12
Open the subject editor tree to what is being edited, and say where it stops

alistair3149:
For https://github.com/ProfessionalWiki/NeoWiki/issues/1327

The structure tree opened every branch it could reach, down a fixed depth, and said nothing about
where it stopped. This makes what is open follow from what is being edited, and gives a row that is
not showing everything a way to say so.

**What is open.** The Subject being edited is open, and so is everything above it. Everything else
follows: on first open the root is the Subject being edited and has nothing above it, so the tree
draws the root and one level — the default is a consequence of the rule rather than a case of its
own. Selecting a nested row makes that row where the next expansion starts from. A row closed by
hand stays closed, until the reader moves to a Subject beneath it. The depth cap is gone: nothing
but a reader's gesture adds a level.

**Expanding is not selecting.** A row now carries a disclosure control that opens it without opening
its editor pane — a pane is a full Subject read and is never unmounted, so the cheap gesture and the
expensive one are now separate. The control spans the row's height, is laid over the guide line down
the level so a row with nothing to open reserves nothing, and answers Left and Right per the
WAI-ARIA tree pattern.

**Saying what a row is not showing.** Four situations drew one identical row — an empty one. Each
now says which it is: a closed row carries the number of Subjects behind it, a row whose Subject
already appears higher up its own branch carries a link mark, one whose Subject or Schema could not
be read carries a warning mark, and a Subject with nothing under it carries nothing. A pane that is
open but no longer reachable keeps its row under "Not linked here", with its count.

**Not built, deliberately.** Prepopulating further than the API already returns buys nothing before
the relations endpoint (#1324), so the walk simply takes relation data whenever it lands. Breadth
caps and a total-row limit were designed and left out: every candidate bound was unreachable on any
data that exists, so both numbers would have been picked by feel.

`NeoTree` gains three optional flags (`expandable`, `expanded`, `collapsible`) and an `end` slot,
and stays message-free — the subject tree resolves the wording and passes it in.

## Manual Browser Check

1. Open `/wiki/ACME_Inc` and press **Edit** on the infobox.
2. The tree shows the root, its three relation groups, and one level of rows. The three products
   and three departments have no control and no number — they hold nothing. Both offices have a
   chevron and a `1`.
3. Press the chevron on **ACME Amsterdam HQ**. It opens without the form pane changing, and its
   number disappears. Press it again to close.
4. Click the **name** of **ACME Vienna Office**. The pane opens *and* the row expands.
5. With focus on a row, use ↑ ↓ → ← : Right opens a closed row then steps into it, Left closes an
   open row then steps out, and neither ever lands on a row you cannot see.
6. The second **ACME Amsterdam HQ**, reached through Headquarters, is a row of its own: expanding
   one occurrence leaves the other closed.
7. Open the Vienna office's pane, then clear the **Offices** field that led to it. Its row moves
   under "Not linked here", keeping its name and count, with no chevron.

## Test plan

- [x] `make ts-build` clean
- [x] `make ts-lint` clean
- [x] vitest: 2111 passed (142 files), from 2086 on master
- [x] Checked in the browser on the dev wiki (steps above)

<sub>AI-authored — Claude Code, `Opus 5 (1M context)`; Alistair set the direction, rejected the issue's prepopulation premise, deferred the breadth cap, and specified the gutter and control geometry from a live study; verified with `make ts-build`, `make ts-lint` and the full vitest suite (2111 passing), and checked in the browser on the dev wiki.</sub>


alistair3149:
https://github.com/user-attachments/assets/6ccd1547-c4d2-4712-bc2e-9e89f8eedb11

JeroenDeDauw:
Closing unmerged: https://github.com/ProfessionalWiki/NeoWiki/issues/1447 replaces the tree with a list of the Subjects open in the dialog, so the disclosure controls, counts and marks built here have no surface to land on. The branch stays the starting point if the September user-testing sessions show the need to reach Subjects several relations away; the issue records that. The bounding survey on https://github.com/ProfessionalWiki/NeoWiki/issues/1324 stands on its own.

> <sub>AI-authored — Claude Code, `Fable 5.1 (max)`; posted on @JeroenDeDauw's instruction, decision from the 2026-09-14 DevTeam sync.</sub>

Issue #1333, opened 2026-09-01
Carry Subject schema, label and Main Subject flag on the subject-page index


alistair3149:
Showing a Subject needs three things that live only in the subject slot: its Schema name, its label, and
whether it is its page's Main Subject. A surface listing Subjects therefore opens every hosting page.
`SubjectIdsByHostingPage` already holds that to one read per page rather than one per Subject, and a page
still hands back all of its Subjects however few are wanted — 250 of them at ADR 29's Acceptable tier.

`PageSubjectsLookup::getMainSubjectId()` makes the third one worse: it re-reads the whole page, the page
the caller has usually just read and discarded, and it is needed for every Subject with no stored label.
Labels became optional in ADR 31, so that is now the common case.

What this is worth today is small, and worth stating plainly. Nine of the ten `SubjectDisplayName` call sites
already hold both the Subject and its page's `PageSubjects`; only `GetSubjectQuery` pays an extra lookup, and
only for the Main-Subject flag. The 78% of a depth-3 relation walk that slot reads account for does not
attribute here either: that walk loads whole Subjects for their statements, which these columns cannot spare
it. They pay off for a reader that needs names without statements — #1324's endpoint over #1334's relation
index — and not before, so this is a prerequisite rather than a win on its own.

## Proposal

Add to `neowiki_subject_page`:

| Column | Holds |
| --- | --- |
| `nwsp_schema` | The Subject's Schema name |
| `nwsp_label` | The stored label, null when the Subject has none |
| `nwsp_is_main` | Whether this Subject is the Main Subject of the page holding it |

All three are readable from the raw slot JSON the index writer already parses, at the call site that
already writes the table, so there is no new write path and no new staleness. `SubjectDisplayName` then has
everything it needs from one indexed query.

A request-scoped cache would remove the repeat reads within one request but not the first read of each
page, which is what a neighbourhood walk is almost entirely made of: it touches each page roughly once.
That is why this is a column and not a cache.

ADR 32 says the subject-page index is "not a fallback and not a cache". These columns are a cache of slot
data, so the ADR needs a rule. Proposed: the index may carry a field the slot defines when that field is
needed to *identify or name* a Subject in a list, is written in the same transaction as the mapping, and is
never read in preference to the slot by anything that edits. Page content stays the source of truth.

Needs a schema migration and a rebuild pass; `RebuildSubjectPageIndex.php` already exists to carry it.

[...]

Issue #793, opened 2026-04-28
Redesign DemoData content


JeroenDeDauw:
This content was created for an older, different, and less capable version of NeoWiki. Time for a redesign from scratch, including the main page.

Audiences / use cases of the demo content:

* ECHOLOT and CH/GLAMs
* knowledge managers and IT/data leads at organizations
* people from the MediaWiki ecosystem exploring NeoWiki, often comparing with SMW or Wikibase
* Jeroen doing demo's, often to the above audiences

Naturally, we want to show off all NeoWiki capabilities. Including the new special pages and such, so some things need linking from the main page. Current main page is a bit of a cluster f***, better structure, easier to understand by new people, is good.

Issue #525, opened 2026-02-04
Build nice Schema creation flow


JeroenDeDauw:
We want nice UX for MediaWiki/NeoWiki novices that want to create a new Schema

Navigating to `Schema:Name of your new schema`  is bad UX. Even if we handled it better than needing to click "create source", saving, and then using our actual editor UI, which is the current state on master.

## Proposed solution

Add a new special page. Special:NewSchema, or similar.

## Considerations

Name bikeshed time: if we go with a special page, we likley want to consider conflicts with existing extensions, maybe Wikibase has such a page. Not a _huge_ deal, as supporting Wikibase+NeoWiki is not a requirement, and not clearly sensible to run that combo in the first place, but we might want to support it ala SemanticWikibase at some point, and no need to make things hard on ourselves if we can easily avoid it.

A special page could show a Schema Name input and then create the schema page and send the user there, automatically opening up the editor there. That is the naive flow that comes to my mind first. But it comes with the big downside that the page is created early on, meaning the user can no longer change the schema name. It seems realistic that after filling out the property definitions, the user changes their mind about the best schema name.

We want to avoid users creating duplicate schemas. Perhaps the schema page should show a list of all existing schemas, and/or a way to easily filter/search them. Also showing a warning if it looks like they are creating a duplicate could be nice. Such a warning could be based on name similarity, or it could go further and check for similar property definitions (this has some interaction with the open technical question of if we go from our current "local properties" to "global properties"). All of this is not needed for an initial version and the last stuff particularly so.

malberts:
> Special:NeoWiki

This makes sense since we're already moving away from the typical MediaWiki approach of having dedicated pages (e.g. https://github.com/ProfessionalWiki/NeoWiki/issues/524).

However, having a dashboard might also be YAGNI if the only link there is "Create schema". Maybe it can also show some stats about schemas/subjects? And a list/table of Schemas?

I would just mention the point about direct links again. It might not be important right now for some things (e.g. https://github.com/ProfessionalWiki/NeoWiki/issues/521#issuecomment-3848858874), but having a direct link that opens the create form on load would be nice. And if we build things like "List/Table of Schemas" to put in the dashboard. It would be nice to be able to go to those parts directly, and not require the user to go to a page, and then navigate somewhere. Of course, this is not impossible, just a UX thing.

PR #1429, opened 2026-09-14
Add Special:NeoWiki, an overview page for working with the wiki as a database

JeroenDeDauw:
A fresh wiki gives no clue what to do with NeoWiki. `Special:NeoWiki` is the start page for people who
think in records rather than in wiki navigation. It became worth building once
https://github.com/ProfessionalWiki/NeoWiki/pull/1406 landed, letting the Subject creator make a Subject
together with its page.

## What it does

`Special:NeoWiki` renders a Vue page, mounted like the other NeoWiki special pages, top to bottom:

- a primary **Create subject** button, opening the Subject creator with the Schema picker and no host
  page, so the host defaults to **A new page**, as on `Special:CreateSubject`;
- a map of the other NeoWiki pages as a wrapping row of cards, one line each on what they hold:
  **Layouts**, **Ontology mappings**, **Graph stores** and **Configuration** (`MediaWiki:NeoWiki`);
- the list of Schemas, headed by a **Manage schemas** link to `Special:Schemas`, each row carrying the
  name linked to its Schema page, the description, and a **Create &lt;Schema&gt;** button opening the
  creator pinned to that Schema, under the same label and rights gating as the Schema page's button.

Everyone may read it; the create buttons follow the existing create-rights hints and are hidden
otherwise. No free text, nothing editable here: not a second main page. A sidebar entry in the NeoWiki
section links to it, and `docs/guide/getting-started.md` gains a sentence pointing at it as the place to
start. Out of scope: statistics or counts (the subject-page index does not carry the Schema yet,
https://github.com/ProfessionalWiki/NeoWiki/issues/1333), recent Subjects, anything per user.

## Decisions to confirm

- The sidebar entry reads **Overview**, first in the section: "NeoWiki" stutters against the section
  heading, "Dashboard" promises statistics this page does not show, "Home" collides with the main page.
  "All schemas" already links to a page titled "Schemas", so a label differing from its page title has
  precedent. The alternative is **Start here**.
- The map sits above the Schema table because that table is unbounded: below it, a map falls under the
  fold on any wiki with a dozen Schemas. `Special:Schemas` sits in the table's header rather than in the
  map, where it would head the same thing twice.
- `Special:GraphStores` is mapped only for holders of `neowiki-admin`, that page's restriction, the way
  `Special:SpecialPages` hides restricted pages.
- **Configuration** is mapped only while `$wgNeoWikiEnableInWikiConfig` is on, as it is by default, and
  the reader may edit `MediaWiki:NeoWiki` (`probablyCan`: `editinterface`, `editsitejson` and protection
  all count). The page may not exist yet; its create view carries NeoWiki's reference table and a
  preloaded example. Both gates are decided in PHP and reach the page as data attributes.
- The primary button is hidden while the wiki holds no Schemas: nothing to create, an empty picker. The
  table then shows the `Special:Schemas` empty message, with the link to that page right above it.
- One `SubjectCreatorDialog` serves both kinds of button: a click sets the Schema to pin and then opens
  the creator, which reads that Schema as it opens.
- The list is read whole through the Schema store, every page of the listing endpoint, unsorted and
  unpaginated, in the order the endpoint returns.
- A listing that fails shows that same empty table, plus an error notification.

## Considered, omitted

- A shared Schema list extracted from `Special:Schemas`: its table is interleaved with pagination, the
  Schema editor and the delete dialog, more cost than the ~20 lines saved.
- The NeoWiki sidebar section on special pages. The hook returns early for titles that cannot exist, so
  this page and the ones it links to carry no link back. Pre-existing.
- A dash for empty descriptions, as `Special:Schemas` renders them.
- The dialog race in https://github.com/ProfessionalWiki/NeoWiki/issues/1432, reachable here because
  each row pins a different Schema: reopening a pinned creator before its first Schema load lands leaves
  the user on the picker. The issue carries a verified fix, held back while the open
  https://github.com/ProfessionalWiki/NeoWiki/pull/1430 edits the same dialog.

**Prior discussion:** proposed and shelved as YAGNI in February 2026
(https://github.com/ProfessionalWiki/NeoWiki/issues/525#issuecomment-3849486479,
https://github.com/ProfessionalWiki/NeoWiki/pull/579), with the onboarding gap patched on the demo
wiki's main page instead (https://github.com/ProfessionalWiki/NeoWiki/issues/793,
https://github.com/ProfessionalWiki/NeoWiki/pull/910).

## Verification

PHPUnit covers the mount point, the graph-stores and configuration attributes (a sysop, a plain user,
the setting off), the page being registered, listed and unrestricted, and the sidebar entry's position
and visibility; a vitest spec covers the rows and links, the cards and their gating, the **Manage
schemas** link, the button-map-table order, both ways of opening the creator, an empty wiki and a failed
listing. Each was seen failing without the production code and against mutations. Full suites, phpcs,
phpstan and the linters pass locally; CI green on MediaWiki 1.43 through master. Driven headless with
the 21 demo Schemas as an administrator (pinning, picker, re-pin, sidebar entry) and anonymously (two
cards and the **Manage schemas** link, no create buttons); the cards wrap cleanly at 600 px.

> <sub>AI-authored — Claude Code, `Opus 5 (max)`; detailed spec from @JeroenDeDauw relayed by an orchestrating session, redirected once; diff not yet human-reviewed; full PHPUnit and vitest suites plus phpcs, phpstan and the linters green locally and on CI, new tests confirmed failing without the production code and against mutations, page driven in a headless browser as admin and anonymously.</sub>

Design, planning and final review by `Fable 5.1 (max)`; implementation, review fixes and this
description by `Opus 5 (max)` subagents. An independent pr-review pass, on `Opus 5 (max)`, found nothing
blocking; its suggestions are applied.

</details>


JeroenDeDauw:
image

JeroenDeDauw:
LGTM for a first version. Presumably lots we can improve on top but this already helps

Issue #1334, opened 2026-09-01
Answering "what points at this Subject"


alistair3149:
Nothing native answers "what points at this Subject": relations are reachable only inside slot JSON, keyed by
the Subject that holds them. The graph store answers it with a reverse edge query. #1324's
`direction=incoming`, #904's inverse relations in the subject editor and #1039's where-used all need an answer
from somewhere. Three ways to get one.

## Read it from the graph store

```cypher
MATCH (source:Subject)-[relation]->(:Subject {id: $subjectId})
WHERE source.wiki_id = $wikiId
RETURN source.id AS source, type( relation ) AS relationType, relation.id AS relationId
LIMIT $limit
```

Nothing to build but that query and its per-page read gating. The edges are already projected, and no save
writes more than it does today.

- ADR 32 records that relation-target suggestions are already read from Neo4j and offer nothing without it,
  and `getSubjectLabelLookup()` returns `NullSubjectLabelLookup` when no plugin is configured. This would be
  the second feature of that shape.
- A wiki with no graph store answers nothing. ADR 32 makes that a supported mode; whether any deployment runs
  it today is not known here.
- The edge names the Relation Type, not the property, and the two are not one-to-one:
  `DemoData/Schema/Everything.json` declares two relation properties both as `"relation": "Has product"`. For
  displaying what points here the type may be the better caption; editing the referring statement needs the
  source Subject, which that path reads anyway.
- The answer can be stale, and #1135 is the known case. Its code half is one expression:
  `Neo4jPageRelationUpdater` filters on `oldRelation.type`, a property nothing writes, where it meant
  `type( oldRelation )`; a characterization test already pins the wrong behaviour and would invert. Its data
  half is separate — `savePage` never clears a Subject's relations wholesale and rebuild has no wipe step, so
  graphs already carrying stale edges need a purge and rebuild.

## Index it in a MediaWiki table

One row per Relation value part, written beside the subject-page index and on the same paths:

```
neowiki_subject_relation
  nwsr_source_id   binary(32)    Subject holding the Relation
  nwsr_ordinal     int unsigned  Statement order, then value order
  nwsr_relation_id binary(32)    The Relation's own persistent id
  nwsr_page_id     int unsigned  Page holding the source Subject
  nwsr_property    blob          Property Name the Relation is held under
  nwsr_target_id   binary(32)    Referenced Subject; may resolve to none, the red-link case

[...]

Issue #1135, opened 2026-07-22
Renaming a relation leaves the old relationship behind in Neo4j


JeroenDeDauw:
Found by code inspection while inventorying relation naming for #630. Not yet reproduced against a live graph, so the fix should start with a failing regression test.

## Expected symptom

When a relation statement's relation name changes (e.g. a Schema's relation property gets a new `relation` value) and the Subject is re-projected, the old relationship stays in the graph next to the new one. The Subject then has two relationships with the same relation id but different labels, and queries matching the old name keep returning results.

## Cause

`Neo4jSubjectRelationUpdater::removeIfTypeOrTargetChanged()` is supposed to delete the outdated relationship:

```cypher
MATCH (subject {id: $subjectId})-[oldRelation {id: $relationId}]->()
WHERE oldRelation.type <> $relationType OR NOT (subject)-[oldRelation]->({id: $targetId})
DELETE oldRelation
```

`oldRelation.type` reads a **property**, but nothing writes a `type` property on relationships: the relation name lives only in the relationship **label** (`createOrUpdate()` interpolates it via `Cypher::escape()`), and `SET relation = $relationProperties` replaces the properties with the user relation-properties plus `id`. So `oldRelation.type` is always `null`, `null <> $relationType` evaluates to `null` in Cypher, and the type clause never matches. The intended expression was almost certainly the label function: `type(oldRelation) <> $relationType`.

The target-change half works (`NOT (subject)-[oldRelation]->({id: $targetId})` is a pattern predicate, no null involved), so retargeting cleans up correctly. And the stale relationship survives the other cleanup too: `removeNonexistentRelations()` keeps every relationship whose `id` is still live, which the stale one is.

## Rebuild does not heal it

`RebuildGraphDatabases.php` re-saves every Subject through this same updater without wiping the store first, so a plain rebuild preserves the stale relationship. Only a manual wipe followed by a rebuild clears it.

## Suggested fix

Change the comparison to `type(oldRelation) <> $relationType` (both sides are then the raw relation name, e.g. `Born in`), with a regression test that renames a relation and asserts the old-labelled relationship is gone. Coordinate with #630 / #1119: if relation edge properties give way to typed Subjects, this code changes shape anyway.

> <sub>AI-authored — Claude Code, `Fable 5 (max)`; filing requested by @JeroenDeDauw after an in-session inventory; issue text not line-reviewed; cause verified by code reading on master, symptom not reproduced live.</sub>

PR #1431, opened 2026-09-14
Show which Subjects reference the one on Special:Subject

JeroenDeDauw:
For https://github.com/ProfessionalWiki/NeoWiki/issues/904

`Special:Subject` now lists the Subjects pointing at the one shown, under **Referenced by**: the same collapsed rows,
ordered by name, captioned with the properties each points through ("via From father").

## How it works

Discovery from the graph, truth from the wiki. Neo4j names candidate ids, scoped to this wiki's Pages and `wiki_id`.
Each is gated on its page's read permission and read through the same published-revision read
`GET /subject/{id}` uses, and kept only while its current Statements still target the Subject shown; those Statements
name the properties.

- The graph can never add a wrong entry: the edge a renamed relation leaves behind
  (https://github.com/ProfessionalWiki/NeoWiki/issues/1135) stays invisible here.
- It can lag: a referrer projected late, or lost to a swallowed projection failure, waits for the next projection or
  rebuild.
- A Subject from another Source neither appears nor has referrers: its relation edges are not projected (ADR 23).

## API

`GET /neowiki/v0/subject/{subjectId}/referencingSubjects?limit=` (default 10, max 50) answers
`{ subject, propertyNames }` entries plus a `truncated` flag; each `subject` is the usual shape with page identifiers.
A Subject the wiki does not serve, unreadable or unpublished, answers the empty list, indistinguishable from an absent
one; `truncated` is decided after that gating, and no total is given (ADR 27,
https://github.com/ProfessionalWiki/NeoWiki/issues/1062).

## On the page

The section renders only when the list is non-empty; the client asks for the maximum and says below it when more were
left out. Nothing on the page waits for the read: not the Subject, not its Edit and Delete controls, not the editor's
save. A failed read costs the section alone; a failed re-read after a write keeps the rows. A referencing row's
relation targets are fetched when it is opened; a Subject in both lists keeps distinct DOM ids and independent
expansion. Rows are named as every other Subject read names them, label-less ones included, and sorted by that name.

## Under the hood

`RelationDisplay` now computes over the Subject registry, so targets seeded later resolve. That exposed a
`SubjectStore` bug, fixed here: the re-sync after a move deleted the fresh copy of a moved Subject still on the page.
`SubjectResponseItemFactory` is extracted from `GetSubjectQuery`; `SubjectRow` gains two optional props, the Data tab
unchanged. Docs: REST API, revision policy, `planning/Relations.md`.

- **Graph route, not an index table** (compared on https://github.com/ProfessionalWiki/NeoWiki/issues/1334, tracked
  on https://github.com/ProfessionalWiki/NeoWiki/issues/1039). No migration, no per-save write inside ADR 29's budget;
  an index can replace the lookup behind this interface.
- **Always on here, no setting.** The issue's default-off is right for pages and views; `Special:Subject` is a record
  view whose reader asked for this Subject's data. The setting comes with the page and view rendering.
- **Neo4j only.** A wiki with only SPARQL stores, or none, shows no section, like relation-target suggestions
  (https://github.com/ProfessionalWiki/NeoWiki/issues/1313). SPARQL has no typed result path in PHP yet, and which
  store holds the native projection is undecided.
- **Property names from the referrer's Statements, not the edge.** An edge carries the Relation Type, which a Schema
  may reuse across properties.
- **Its own endpoint, beside the one proposed in https://github.com/ProfessionalWiki/NeoWiki/issues/1324.** That
  proposal answers with summary nodes kept out of the registry; these rows need full Subjects seeded into it. Whether
  it folds into that one is open.
- **Cap 50, a flag, no total.** Neo4j orders and windows before verification, so the rows are the first by name.
  Verification stops one past the limit, so the common case costs about as many Subject reads as rows shown, a mostly
  unreadable or stale one up to four times the limit.

## Considered, omitted

A SPARQL candidate source; an index table; a Subject that references itself; grouping rows by property; pagination; a
"nobody references this" line; fixing https://github.com/ProfessionalWiki/NeoWiki/issues/1135 here; amending ADR 32's
Neo4j-dependency sentence; the editing and the page-and-view sides of
https://github.com/ProfessionalWiki/NeoWiki/issues/904; referencing-row edit and delete controls judged by their own
page; a user-guide page, since `Special:Subject` has none yet.

## Screenshots

| | |
|---|---|
| A person referenced by four birth events, the first row opened; its own relation targets were fetched when it opened. | [SCREENSHOT: referenced-by-bach.png] |
| A Subject nothing references: no section. | [SCREENSHOT: referenced-by-acme.png] |
| The same page at 420px. | [SCREENSHOT: referenced-by-mobile.png] |

## Testing

PHPUnit covers the lookup's scoping, ordering and limit, the endpoint's property names, a stale edge, an unreadable
referrer, an unreadable or unpublished target, truncation, the cap and no graph store; vitest covers the section, its non-blocking read and
failed re-read, DOM ids, lazy seeding and the store fix. Key behaviours were mutation-checked.

> <sub>AI-authored — Claude Code, `Fable 5.1 (max)` orchestrating `Opus 5 (max)` implementer and reviewers; detailed spec from @JeroenDeDauw's 2026-09-14 decision through an orchestrating session, then five steers; diff not yet human-reviewed; suites, linters and build green locally and on CI, key behaviours mutation-checked, the section walked in a scripted browser.</sub>

Steers: keep the over-fetch; rebase; simplify and de-noise; merge the fixes from @alistair3149's `/pr-review` pass
(https://github.com/ProfessionalWiki/NeoWiki/pull/1436, `Opus 5 (xhigh)`) with their comments trimmed; resolve the
conflict with master and squash. The last two were done by `Fable 5.1 (max)` directly, with vitest, build, linters and
the referenced-by PHPUnit classes rerun locally after the master merge and the full PHPUnit suite on CI.

</details>


JeroenDeDauw:
image

alistair3149:
LGTM, #1436 contains the blocker fixes and I filed #1437 as a follow-up

PR #1433, opened 2026-09-14
List the NeoWiki group first on Special:SpecialPages

JeroenDeDauw:
Special:SpecialPages now opens with the NeoWiki group instead of burying it among the core groups.

Group order follows where each group's first page sits in the special page list, so the `SpecialPage_initList` handler moves NeoWiki's entries to the front. They are matched on class namespace, so future NeoWiki special pages are covered and RedHerb's are not.

Caveats:

- Leans on ordering behaviour core does not document, because core offers no hook for group order. Checked against 1.43 and current master, where `SpecialSpecialPages` only sorts within a group and moves `other` last.
- Lifts NeoWiki above every core group on every wiki in a farm.

Considered, omitted: a setting for the group position.

Three tests cover it, from the rendered group order down to the special page list, so a core change that drops the behaviour breaks CI rather than passing quietly.

> <sub>AI-authored — Claude Code, `Opus 5 (max)`; one-line ask from @JeroenDeDauw relayed by an orchestrating session, no redirections; diff not yet human-reviewed; every test confirmed failing against a mutated hook, full PHPUnit suite and `make cs` green locally, group order checked on a dev wiki, CI green.</sub>


JeroenDeDauw:
image

JeroenDeDauw:
LGTM

alistair3149:
LGTM, though I think we should allow user to opt out from this. If that is desired, then it can be done as a follow-up.

PR #1435, opened 2026-09-14
Link relation values in Subject rows to the target's Special:Subject page

JeroenDeDauw:
For https://github.com/ProfessionalWiki/NeoWiki/issues/1355

On the two surfaces that show Subject records, the Data tab (`?action=subjects`) and
`Special:Subject`, a relation value inside a Subject row now links to the target Subject's own page,
`Special:Subject/<id>`, instead of to the wiki page the target is stored on. The link text is
unchanged. A reader of these rows is browsing Subjects, so a relation leads to the target's own
record. The page storing the target is about another Subject unless the target is its Main Subject,
and the target's `Special:Subject` page links that page in its footer. Targets the row cannot
resolve are shown as before, without a link.

Everywhere else, the infobox and other Views on articles included, relation values keep linking to
the wiki page, where those readers are.

How: `RelationTargetUrlKey`, declared in `ValueDisplayContract.ts` beside `ValueDisplayProps`, is
an injection key for a function from the resolved target to its link url. `RelationDisplay` injects
it with the page link as its default, so the display contract extensions implement is unchanged.
`SubjectRow` provides the `Special:Subject` variant once, so the referenced-Subject rows on
`Special:Subject` get it too.

Tests: `RelationDisplay.spec.ts` pins the default and an injected resolver. `SubjectRow.spec.ts`
renders a row through the real value-display components and checks the href.

Considered, omitted:

* Providing from `SubjectsManagerPage` and `SubjectPage` instead of from the row: the same line
  repeated per surface and per list of rows, with nothing gained.
* A boolean flag instead of a resolver function: the function keeps `Special:Subject` out of a
  generic value display.
* Assertions in the two page specs: the behaviour is the row's, and the row spec covers it.

https://github.com/ProfessionalWiki/NeoWiki/pull/1431 changes the same import line in
`RelationDisplay.vue`. Whichever merges second rebases.

> <sub>AI-authored — Claude Code, `Opus 5 (max)` implementation, `Fable 5.1 (max)` review and description; detailed spec from @JeroenDeDauw's 2026-09-14 decision through an orchestrating session, no redirections, one polish pass after review; diff not yet human-reviewed; vitest, build and lint green locally and on CI, the new tests seen failing without the change and against six mutations, relation hrefs checked in a headless browser on the Data tab, Special:Subject and an article infobox.</sub>


JeroenDeDauw:
LGTM, tested locally

Issue #1432, opened 2026-09-14
Reopening a pinned Subject creator before its Schema loads falls back to the picker


JeroenDeDauw:
`SubjectCreatorDialog` opens on the generic Schema picker instead of the Schema it was pinned to. No error, nothing in the console.

`loadSchema()` guards its own write with the `requestSequence` counter, but `pinInitialSchema()` then runs `if ( loadedSchema.value === null ) goBack()` unguarded, and `goBack()` bumps `requestSequence` itself:

```
click "Create Museum"  -> seq=1, awaiting
close                  -> resetForm(): seq=2
click "Create Museum"  -> seq=3, awaiting
first response lands   -> loadSchema returns early (1 != 3)
                       -> pinInitialSchema #1 resumes, loadedSchema is null -> goBack() -> seq=4, picker shown
second response lands  -> loadSchema returns early (3 != 4), never applied
```

Every open starts a fetch, so any caller pinning a Schema is exposed. It was only reproducible from Special:NeoWiki (https://github.com/ProfessionalWiki/NeoWiki/pull/1429), whose page of per-Schema Create buttons makes a quick close and reopen natural, at 150 to 400 ms of schema-fetch latency. Manual reproduction needs that branch; the diffs below apply to master.

Left out of https://github.com/ProfessionalWiki/NeoWiki/pull/1429 because https://github.com/ProfessionalWiki/NeoWiki/pull/1430 edits the same component and spec: this wants to land before or after that PR, not alongside.

<details><summary>Verified fix and regression test</summary>

The test fails without the fix. It was verified in a scratch copy taken before the `createOtherSubject` rename and re-derived against master, so these diffs are the same change rather than the exact files that were run.

```diff
--- a/resources/ext.neowiki/src/components/SubjectCreator/SubjectCreatorDialog.vue
+++ b/resources/ext.neowiki/src/components/SubjectCreator/SubjectCreatorDialog.vue
@@ -764,7 +764,8 @@
 	await loadSchema( schemaName );
 }
 
-async function loadSchema( schemaName: string ): Promise<void> {
+// False when a later load superseded this one, so its outcome was discarded.
+async function loadSchema( schemaName: string ): Promise<boolean> {
 	selectedSchemaName.value = schemaName;
 
 	const currentSequence = ++requestSequence;
@@ -773,18 +774,20 @@
 		const schema = await schemaRepo.getSchema( schemaName );
 
 		if ( currentSequence !== requestSequence ) {
[...]

PR #1430, opened 2026-09-14
Add {{#create_subject}}: a wikitext button opening the Subject creator

JeroenDeDauw:
Fixes https://github.com/ProfessionalWiki/NeoWiki/issues/1432
Follows-up to https://github.com/ProfessionalWiki/NeoWiki/pull/1406

Adds `{{#create_subject: schema=<Schema name> | page=<new | this | a page title> | text=<button label>}}`: a Codex
button, placed anywhere in wikitext, that opens the Subject creator in place. All parameters are optional and named;
without `schema` the dialog asks for one. An unknown Schema, an unusable page title, or an unknown argument renders a
parser error in place of the button. After saving, the dialog goes to the page the Subject was stored on, or reloads
this page. Named `#create_subject`: only `{{#neowiki_value}}` carries the prefix.

The button shows for everyone. A user without the wiki-wide `createpage` and `edit` rights, the hint the Schema page's
button uses, gets MediaWiki's own permission reason in a dialog on clicking it rather than no button: an author who
placed the button built the page around it, and the reason is what tells a reader whether logging in would help. The
endpoints enforce the per-page rights.

The parser function emits a placeholder with escaped `data-mw-neowiki-*` attributes and no other HTML, and loads the
frontend where the page does not already, so the button also works on Help: and Project: pages. Whether the hosting
page can hold Subjects, and whether it has a Main Subject, is resolved at parse time, which stays correct because its
Subjects live in its own revision. A named target page and the Schema page are resolved at parse time as link
dependencies, so creating or deleting either re-parses the page. Whether another page already has a Main Subject is
read when the dialog opens.

The creator dialog gains an `initialPage` prop and, like the other dialogs, an `open` prop with `update:open` in place
of reading the store flag; the flag stays in the store as the page-level signal, and the four call sites in the repo
bind it two-way. `SubjectCreatorDialog` is exported from the public API, so that is a change to its contract. Folded
in while in that code: the pre-existing race where reopening a pinned creator before its first Schema load landed
fell back to the picker.

A `page=<title>` that does not exist yet is refused outside the main namespace, where `POST /neowiki/v0/subjects`
cannot create it; existing pages in any namespace work. PHPUnit covers the parse-time resolution, escaping, the
errors and the link dependencies; vitest covers the button and the dialog's fixed and preselected modes.

### Follow-up

`page=this` falls back to `new` outside the content namespaces while `page=<that same title>` stores there. I would
[...]

PR #1438, opened 2026-09-14
Add a create-subject button to each Schema row on Special:Schemas

JeroenDeDauw:
Special:Schemas is where users browse Schemas, yet creating a Subject from one meant leaving for the Schema's own page or the Special:NeoWiki overview of https://github.com/ProfessionalWiki/NeoWiki/pull/1429. Each row now carries a page-with-plus icon button opening the Subject creator pinned to that row's Schema.

- It sits ahead of the row's edit and delete buttons, named "Create <Schema>" via the existing `neowiki-schema-create-subject` message, shown as a Codex tooltip too.
- A plain plus beside the Schema's edit and delete buttons read as adding to the Schema; a labelled button is too wide for the table.
- It shows when the user may create Subject pages (`canCreateSubjectPage`), independently of the `canEditSchema` gate on the edit and delete pair.
- One `SubjectCreatorDialog` serves all rows, mounted with no host page as on the overview page: a new page by default, or an existing one.

Coordination (merge order does not matter; the second to land adjusts in its rebase):

- https://github.com/ProfessionalWiki/NeoWiki/pull/1429 rewrites the qqq line of `neowiki-schema-create-subject`, left alone here; the second to land adds Special:Schemas to it.
- https://github.com/ProfessionalWiki/NeoWiki/pull/1430 makes `open` a required prop of the creator dialog; the second to land adds `v-model:open="subjectStore.subjectCreatorOpen"` to the dialog mounted here (its build stays red until then).

https://github.com/ProfessionalWiki/NeoWiki/issues/1432 (a pre-existing race when reopening a pinned creator) is reachable from here too; not fixed here.

Verification: vitest, `ts-build`, `ts-lint`, `cs`, PHPUnit and CI green. Browser-checked as admin ("Create Artist" opens the creator on Schema: Artist with a new page preselected; the saved Subject gets its own page) and anonymously (no icon, no dialog).

> <sub>AI-authored — Claude Code, `Opus 5 (max)` implementing, `Fable 5.1 (max)` reviewing; detailed spec from @JeroenDeDauw, tooltips redirected once; the page icon in place of the spec's plus was the session's own choice, reverted in a follow-up PR; diff not yet human-reviewed; tests and browser checks run locally, mutation-checked, CI green.</sub>


JeroenDeDauw:
image

JeroenDeDauw:
One more revision landing in the next hour

JeroenDeDauw:
LGTM

malberts:
Minor UI nits:
* the new button has a tooltip, but existing buttons do not, which makes the existing buttons look partially broken because they do not behave the same
* in general, Codex tooltips are awkward - they usually go under my cursor, and require moving the mouse again to see (but this might be a "me" issue)

[Screencast_20260915_213414.webm](https://github.com/user-attachments/assets/2f81f9ce-70ae-441f-8b5a-6fd5017fe2b1)

JeroenDeDauw:
> the new button has a tooltip, but existing buttons do not, which makes the existing buttons look partially broken because they do not behave the same

I originally told claude to add tooltip to the new + icon since otherwise it was not clear enough and then it added it to all buttons and I told it to just do it for the + icon because it looks ugly and isnt needed for the others and also isnt present on the other pages. So yes, its not consistent on the "every button has it" level but it is on the "every button that isnt obvious has it" level, and ofc consistency is not the only metric.

I think the main contender is adding full buttons with "Create $shemaName" like on Special:NeoWiki but idk, seems verbose here

image

alistair3149:
> the new button has a tooltip, but existing buttons do not, which makes the existing buttons look partially broken because they do not behave the same

I assigned the tooltip to other icon buttons in #1443.
CdxTooltip was dropped because it has significant performance issues when you use a lot of them, which is the case for Special:Schema.

PR #1440, opened 2026-09-14
Fixes for {{#create_subject}}: Schema names, landing after save, new pages, denial reason

alistair3149:
Follows-up to https://github.com/ProfessionalWiki/NeoWiki/pull/1430

- **Schema names are stored as their page is titled.** `schema=city` used to store Subjects under Schema "city",
  which relations targeting City reject. `schema=Schema:City` still works; a title in another namespace is now an
  unknown-Schema error.
- **Saving goes to the page the Subject went on, also with `page=`.** Content-page buttons with `page=new` or
  `page=<title>` used to go to Special:Subject.
- **A new page's button is rendered again once the page exists.** Saved through the edit stash, a new page kept
  output parsed before it had an id, so its `page=this` button created a separate page until the next edit. The
  output now records its page id dependency like `{{PAGEID}}`, costing one extra render on creation.
- **The denial reason is computed only on article views carrying the button.** It used to be built on every page that
  loads the frontend, through the deprecated `User::newFatalPermissionDeniedStatus`; `Authority::isAllowed` now gives
  the same text.
- **Buttons that store on an existing page need only the right to edit it.** `page=this` and `page=<existing page>`
  used to ask for wiki-wide `createpage` as well. Buttons that create a page, and buttons without `page=`, still need
  both.
- **A fixed `page=new` button's title-taken error asks for a different label.** It used to say to pick the page as an
  existing one or change the title, neither of which that dialog offers.
- **Tests for the creator's `open` bindings:** on Special:CreateSubject, the Schema page, the Subjects manager and the
  button, plus the button's denial dialog. A one-way binding at any of them passed every test.

Not changed, for you to decide: fixed `page=new` and named-page buttons now fetch edit notices they never show, as
unfixed buttons already do.

Deferred until someone needs it: a button placed through an interface message, such as MediaWiki:Sitenotice, shows the
generic denial text, as the reason is computed only for buttons in the page's own wikitext.

## Manual Browser Check

On the dev stack, where anonymous users cannot edit, with this branch checked out, the frontend built and the demo data
imported:

1. Create a main-namespace page with the test wikitext from
   https://github.com/ProfessionalWiki/NeoWiki/pull/1430#issuecomment-5669978645, plus
   `{{#create_subject: schema=city}}`.
2. Logged in, the added button reads "Create City". Save a Subject with it: its Special:Subject page shows
   "Schema: City".
3. Under "Fixed target: Rijksmuseum", save a Subject: you land on Rijksmuseum. Under "Fixed target: a new page",
   save one: you land on the page created for it.
4. Under "Fixed target: a new page", label a Subject "Rijksmuseum" and save: the error asks for a different label.
5. Logged out, click any button: the dialog says the action is limited to the Users group. In the browser console,
   `mw.config.exists( 'wgNeoWikiCreateSubjectPageDeniedReason' )` is `true` on your page and `false` on Main Page
   and Special:Schemas.
6. Set `$wgGroupPermissions['*']['createpage']` and `$wgGroupPermissions['user']['createpage']` to `false`. Logged in,
   "Fixed target: this page" and "Fixed target: Rijksmuseum" open the creator; "Fixed target: a new page" explains
   the missing right.

> <sub>AI-authored — Claude Code, `Opus 5 (1M context)`; review of #1430 and these fixes requested by @alistair3149 via `/pr-review`, the denial-reason and permission fixes added at their request; diff not yet human-reviewed; PHPUnit on the affected classes, phpcs, phpstan, vitest, lint and build pass locally, new tests mutation-checked, behaviour checked on a dev wiki except the title-taken message, which vitest covers; CI green on MW 1.43 through master.</sub>


docs/authoring/parser-functions.md
```diff
@@ -165,7 +165,8 @@ Renders a button that opens the Subject creator without leaving the page.
 
 - With `page`, the user cannot change where the Subject goes. Without it, the creator asks, starting from a new page.
 - The Subject becomes the page's Main Subject where that page has none, and joins the existing Subjects otherwise.
-- Saving takes the user to the page the Subject was stored on.
+- Saving takes the user to the page the Subject was stored on. When the button is outside the content namespaces,
+  it goes to the Subject's `Special:Subject` page instead.
```
JeroenDeDauw: Maybe better to just drop the bullet entirely

PR #1443, opened 2026-09-15
Show row icon button labels as a title on the Schemas, Layouts and Mappings pages

alistair3149:
Follows-up to https://github.com/ProfessionalWiki/NeoWiki/pull/1438

Each row's create-subject button on Special:Schemas showed its label in a `v-tooltip`. In the Codex that MediaWiki 1.43 ships, every mounted tooltip tracks its position while hidden, so the cost grows with each row: in Chrome, a 50-row table scrolled at 7–8 fps, against 60 fps without the tooltips. The tooltip also set its text only on mount, so after paging the table it named the Schema that held that row before.

The button now shows its label as a native `title`, as the row buttons on Special:ManageSubjects do. The edit and delete buttons in the rows of Special:Schemas, Special:Layouts and Special:Mappings, which showed no label on hover, get a `title` too. A `title` appears only on mouse hover, while the tooltip also appeared on keyboard focus. `aria-label` still names every button.

## Manual Browser Check

1. Open Special:Schemas as a user who may create pages and edit Schemas, on a wiki with more than 10 Schemas.
2. Hover each icon button in a row. The browser shows "Create <Schema>", "Edit schema" and "Delete schema".
3. Page the table forward and hover a create button. It names that row's Schema.
4. On Special:Layouts, hover a row's buttons: "Edit layout" and "Delete layout". On Special:Mappings: "Edit mapping" and "Delete mapping".
5. Optional, with 50 or more Schemas: show 50 rows per page and scroll. The page scrolls smoothly.

> <sub>AI-authored — Claude Code, `Opus 5 (1M context)`; found reviewing #1438 at the request of @alistair3149, redirected once (native `title` on all three list pages); diff not yet human-reviewed; new title tests fail without the change, full vitest suite, ts-build and ts-lint pass locally; browser-checked on MediaWiki 1.43 (titles on Schemas and Layouts rows, no per-row observers or scroll listeners at 50 rows).</sub>

<details><summary>Considered, omitted</summary>

- The per-scroll-event cost behind the fps figures: about 80 ms at 50 rows and 16 ms at 10, against under 1 ms without the tooltips.
- Whether later Codex releases stop tracking hidden tooltips.

</details>


JeroenDeDauw:
I think this is net-negative. Most of those buttons are obvious, adding tooltips to them makes them look less nice

I deliberatly did not do this during my own implementation as per https://github.com/ProfessionalWiki/NeoWiki/pull/1438#issuecomment-5689074200

alistair3149:
> I think this is net-negative. Most of those buttons are obvious, adding tooltips to them makes them look less nice

It is not the same tooltip implementation as in #1438.

This PR does two things:
1. Switch the CdxTooltip tooltip to a native tooltip with `title` attribute for the create subject button. There is a significant performance issue because it doesn't scale.
2. Apply the same title attribute to other buttons in the same row. This is more for consistency and missing affordnance.

I disagree on treating the other icon buttons differently since there aren't really downside to have the tooltip if it is using the `title` approach, and the upsides are valid. However, if we really want to go with the create subject tooltip only approach, then we need code comment or test to explain and pin that behavior specifically.

JeroenDeDauw:
Ok I'll see how annoyed I get by the UI next time I use it; didn't preview this PR locally yet

Issue #1449, opened 2026-09-16
Creating a Subject with a page of its own cannot take a pre-minted Subject ID


alistair3149:
`POST /neowiki/v0/subjects` — the endpoint that creates a Subject together with a page of its own — mints the Subject ID server-side with no way to supply one. Its two siblings, `POST /page/{pageId}/mainSubject` and `.../subjects`, share `CreateSubjectApi` and both take an optional `id`, [documented in the subject format](https://neowiki.ai/docs/api/subject-format#creating-subjects).

## Where that bites

**Import.** `POST /neowiki/v0/subject-ids` exists so an interlinked batch can wire its relations before any of its Subjects exists (#1100). Subjects that each get a page of their own fall outside it: an importer has to create those pages some other way, then call `mainSubject` with the ID it minted.

**Retries.** A lost response duplicates a page. With an ID, the retry answers 409 instead.

**Pointing back at the Subject being created.** Subjects created in flow from the Subject Editor's relation fields can be pointed at while still unsaved, because their IDs are minted up front (#1402). The Subject at the root cannot, since this endpoint is one of the three routes that writes it — so it is left out of `draftSubjects` in `SubjectEditorDialog`, and a Birth event created while filling in a Person cannot point back at that Person in the same save.

## What it takes

An optional `id` on `CreateSubjectPageApi` and `CreateSubjectPageRequest`, threaded into `buildSubject` with the locality and in-use checks `CreateSubjectAction` already makes, and a 409 on `CreateSubjectPagePresenter`. Then `id` on `createSubjectPage` *and* `createMainSubject` in `SubjectRepository`, `RestSubjectRepository` and `SubjectStore` — `createMainSubject`'s server side is already there, only the TypeScript omits it.

The third bullet is worth splitting off: what a Subject being created may point back along is a question the relation design (#630) may answer differently, while the endpoint and the client gap stand on their own.

> <sub>AI-authored — Claude Code, `Opus 5 (1M context)`; rewritten after @alistair3149 asked whether the issue was worth doing and the original's premise turned out to be half wrong; reviewed by him before posting; code facts checked against master at 2f3d1714 (the shared `CreateSubjectApi`, its `id` param, its absence on `CreateSubjectPageApi`, the `draftSubjects` exclusion).</sub>

JeroenDeDauw:
[...]

PR #1450, opened 2026-09-16
Add multi-Subject support to the Subject Creator

### LINKED ISSUE #1402: Add multi-Subject support to the SubjectCreator (like already in SubjectEditor)
Being able to create new Relation-target Subjects when creating a Subject might be even more important than when editing an existing Subject. Seems like we forgot to also apply this new capability to the SubjectCreator. It's quite jarring that it's not there.

Keep in mind the seperation between SubjectEditor and SubjectCreator is intentional, they should not be refactored to a single unit with context-dependent behavior.

---

In the Subject Editor, a relation property such as Products lets you create a new Product on the fly (https://github.com/ProfessionalWiki/NeoWiki/pull/1339). In the Subject Creator, the same property only lets you pick existing Products. Creating a Subject together with the Subjects it points at is an important use case, so this is surprising, both for users and technically.

## Cause

The relation input is the same component in both dialogs. Its picker only offers "Create new Product" when the hosting dialog provides the draft handling behind it: minting the draft, a pane to fill in its fields, and saving targets before the Subjects that point at them. The Subject Editor dialog has all of that. The Subject Creator dialog renders the bare form with its own save path, so the picker falls back to existing Subjects.

## Direction

Build the Subject Creator on the Subject Editor dialog: the creator is the editor opened on a draft root Subject, after the Schema has been chosen. That gives the creator on-the-fly targets, the tree view and a single save path, instead of adding draft handling to the creator a second time.

The page does not change: drafts go on the page the creator was opened from, as in-flow targets do in the editor today. https://github.com/ProfessionalWiki/NeoWiki/pull/1374 is not a prerequisite. With it, the chosen page is created first and the save then runs as before.

## Optional

- Whether the tree view belongs in the creator can be decided separately. Multiple Subjects in one go would be nice, can be done later.

Related: https://github.com/ProfessionalWiki/NeoWiki/issues/971, https://github.com/ProfessionalWiki/NeoWiki/issues/1355

> <sub>AI-authored — Claude Code, `Fable 5.1 (max)`; one-line ask from @JeroenDeDauw, direction and wording his, no revisions; text reviewed by @JeroenDeDauw; code facts (shared picker, injected draft handling, separate creator save path) checked against master f08093a6, fresh-context judge pass.</sub>

alistair3149:
Fixes https://github.com/ProfessionalWiki/NeoWiki/issues/1402

A relation field in the Subject Editor can create the Subject it points at; the same field in the
Subject Creator could only pick one that already existed. The creator now opens the editor on a
Subject the wiki does not hold yet, so the two behave alike and there is one save path rather than
two.

**What the creator keeps** is what only it knows: which Schema, and which page the Subject goes on.
Choosing the Schema is still its own step. Everything after it is the editor, with the page question
in its footer, answered by the same Save button.

**Which page each Subject is stored on** is resolved when the write runs: the page being viewed, the
page picked, or the page the first write settled. A Subject created while drilled into another one
goes on that Subject's page instead — the nearer answer wins. The whole answer is read once, before
the first write, and the footer freezes for the duration, so the page a Subject lands on and the
page the dialog then says it landed on cannot come apart.

**A save that stops part way** leaves what it wrote and says so, and the retry resumes rather than
repeating: a Subject already written drops out of the set, and the one being created is updated
under the id the server gave it instead of being offered a second time. An id the server already
holds is read as that create having landed with only its answer lost. Closing on a partial save asks
first, because the line recording it goes with the dialog. What a failed target leaves behind is a
relation naming a Subject that does not exist — wiki-native red-link behaviour, and a warning rather
than a refusal.

**Seams added to `SubjectEditorDialog`**, which is public API: `rootIsNew`, `saveDisabled`,
`hostHasUnsavedChanges`, `onSaved`, and a `before-actions` footer slot carrying the save's progress.
The dialog title and the Save button label follow `rootIsNew` rather than being passed in. No
existing host changes. Every write is authorized server-side per page, as before; the client's
permission hints are advisory and unchanged.

## Intended, not oversights

- **The back button is gone.** Returning to the Schema picker now costs closing and reopening — two
  clicks where it was one. Dropped to simplify and expedite this PR; it can be added back later.
- **The label moved into the pane**, edited with the pencil beside the name. Its placeholder no
  longer previews the title the page will have, only the Schema stand-in.
- **A refused page title no longer disables Save.** One answer to it is a different label, which is
  typed in the editor now, so a Save held shut on the title field could never be given it.
- **Edit notices for the page being viewed show even while the Subject is bound elsewhere**, since
  the editor loads its own.
- **A Schema drafted in step 1 and saved with the Subject** gets the Subject's edit summary on its
  first revision when none was typed. Pre-existing behaviour carried the typed summary across; only
  the empty case now reads differently.

Creating several unrelated Subjects in one pass is still not possible: every Subject beyond the
first arrives as the target of a relation.

Pointing a nested Subject back at the one being created needs
https://github.com/ProfessionalWiki/NeoWiki/issues/1449 — two of the three creation routes mint the
ID server-side, so the relation would name an ID the Subject never gets.

## Manual Browser Check

Needs the demo Schemas: `make import-demo-data`.

1. Open `Special:CreateSubject` and pick **Company**. The pane names the Schema beside the Subject's
   name, and the badge links to `Schema:Company`.
2. In **Products**, type `Widget` and pick **Create "Widget" as a new Product**. A pane and a tree
   node appear and the dialog widens.
3. Open **Store the subject on**, leave **A new page**, and click **Create subject**. The page radios
   grey out while the save runs. You land on the new Company; its **Products** relation resolves to
   the Widget on the same page, in two revisions.
4. Open `/w/index.php?title=ACME_Inc&action=subjects` and add a Subject there. Its in-flow targets
   land on that page instead of on one of their own.
5. Back on `Special:CreateSubject`, type `Main Page` as the page title and save. The refusal shows at
   the title field, Save stays enabled, and retyping the title lets the save through.
6. New-schema route: build a Schema, **Continue**, open it from the pane's badge, add a property and
   save it there. The Subject's own save must not drop that property.

## Considered, omitted

- Restoring the back button behind a `#header` slot: Codex renders that slot as a replacement, so
  the title, subtitle and close button come back by hand.
- A composable shared between the two dialogs instead of this composition: it would export ~23
  members against four props and a slot, and duplicate the two-surface grid, divider and Codex
  workarounds, which are template and LESS rather than script.
- Keeping the gate that skipped the dry-run validation while the Schema was still a draft. The
  creator held one and this drops it: on the new-schema route, between **Continue** and **Save**,
  each blur now validates against a Schema the wiki does not hold. The response is swallowed, so the
  cost is a wasted request and a console line. Restoring it means a draft flag threaded to
  `SubjectEditPane`, a fifth prop on a published component; treating "schema not found" as nothing
  to report would fix it with none, and can come later.
- Two pre-existing issues this change reaches but does not introduce, both byte-identical on master:
  `createSubjectPage` carries no id, so a lost response duplicates a page; and Save goes live before
  the picked page's Main-Subject read lands, which can refuse the write with an untranslated message.

<sub>AI-authored — Claude Code, `Opus 5 (1M context)`; issue and architecture chosen by @alistair3149 from a written A/B of two designs, with three further design questions answered by him before any code; diff not yet human-reviewed; vitest 2344 green, `ts-lint` and `ts-build` clean, each production change ablation-verified, both save routes plus the footer freeze driven in the dev wiki, and a full multi-agent review pass run over the branch with its findings fixed on it.</sub>


alistair3149:
https://github.com/user-attachments/assets/ed61ee01-4353-4fe2-84f2-fc0af70bfa2d

Issue #1454, opened 2026-09-16
Let each Property Type say what its definition still needs


alistair3149:
The Schema editor now holds back a save when a Relation property has no target Schema, because the wiki refuses such a definition and the refusal names the stored JSON rather than the field left empty. That check is hardcoded to Relation in `SchemaEditor.vue`'s `incompleteProperty()`, reading `missingRelationAttribute()` from the Relation domain module.

A Property Type defined by an extension has no way to join in. Its definition can carry required fields of its own, and the editor will happily hand the wiki something it rejects, with the same unhelpful message the Relation case used to produce.

## Proposal

Put completeness on the Property Type contract: each type answers what its own definition still needs, and the editor asks the registry rather than naming one type.

Sketch, against `BasePropertyType`:

```ts
/**
 * The message naming what this definition still needs before the wiki will store it, or null.
 */
public missingAttribute( property: T ): string | null {
	return null;
}
```

`RelationType` overrides it with the body `missingRelationAttribute()` already has; every other built-in type keeps the default. `incompleteProperty()` then loops the property definitions and asks each one's registered type.

## Decisions this needs

- Whether this belongs on the Property Type contract at all, or whether validating the draft against the content JSON Schema in the frontend covers it better. The content Schema already knows every required field, including an extension's, if extensions can contribute to it.
- Whether the returned value stays a message key or becomes a richer object (which field, which message), so the editor can mark the field as well as name the property.
- Whether the server should answer this instead, given the editor is not the only writer.

Documented in `docs/extending/extending.md` if it lands, and the Relation-specific comment in `SchemaEditor.vue` should point here or be removed.

## Background

The Relation case reached a user through the Subject creator: creating a new Schema, adding a Relation property and creating the Subject failed on save with `Schema content is invalid (1 error):` and nothing after the colon. Two other fixes came out of it — the target-Schema picker regression (#1453) and the REST error envelope dropping the per-error detail.

> <sub>AI-authored — Claude Code, `Opus 5 (1M context)`; written at @alistair3149's request after diagnosing the reported bug in-session; not yet human-reviewed; describes a proposal, not built — the sibling fixes it refers to were tested with vitest and PHPUnit and checked against a local dev wiki.</sub>

PR #1455, opened 2026-09-16
Say why a Schema, Layout or Mapping will not save

alistair3149:
Follows-up to https://github.com/ProfessionalWiki/NeoWiki/pull/1453

Creating a Schema with a Relation property through the Subject creator still failed after #1453 restored the target-Schema picker. Choosing no target left the property without one, which the Schema content format requires, and the refusal arrived at the very last step as:

```
Failed to update (unnamed RelationTest).
Error saving schema: Schema content is invalid (1 error):
```

Three things had to be true for that to be the message.

**The reason never left the server.** `SchemaContentHandler` reported the error count, then each error as a message of its own. MediaWiki keeps only the first message a `Status` holds when it builds a REST error (`ActionModuleBasedHandler::throwHttpExceptionForActionModuleError`), so the list the trailing colon promises was dropped. The details now ride in that first message as a list parameter. They stay separate messages rather than one pre-formatted string, so each is resolved in the reader's language — `uselang=qqx` shows the nesting intact. `LayoutContentHandler` and `MappingContentHandler` carried the same defect and get the same treatment.

**The text was lost before it was written.** `MediaWikiPageSaver.savePage` rejected with a plain object although the `PageSaver` interface it implements promises a resolved `PageSaverStatus` — as its own `SucceedingPageSaver` and `FailingPageSaver` do. Every caller's `!status.success` branch was therefore unreachable: the `await` threw the object first, and the dialogs, finding it is not an `Error`, rendered `String( error )` as `[object Object]`. It now resolves, on the edit-token leg as well as the save itself, and a refusal always carries a reason the type insists on.

**Nothing asked for the target.** The field's error waited for a blur that a user who never focuses the picker never triggers, and neither the Schema editor's save nor the Subject creator's Continue checked. `SchemaEditor` now reports the first property definition the wiki would refuse, and the three dialogs that save a Schema hold back on it, naming the property. It reads the Schema rather than the mounted property editor, because only the selected property has one and an unselected property is just as unsaveable.

Generalising that last check to the Property Type contract, so an extension's type can answer for its own definition, is #1454.

## Three changes worth objecting to

- **The target-schema error now shows without waiting for a blur, reversing the choice made in #953.** That commit deliberately deferred it. The save is now held back until a target is chosen, so a property switched to this type has to say what it is waiting for rather than look ready — but if the original reasoning still stands, this is the piece to drop.
- **`savePage` is a published surface** (`public-api.ts` exports `PageSaver` and `MediaWikiPageSaver`). An out-of-tree consumer that wrapped it in `try/catch` expecting a rejection now sees a failed save resolve. The declared interface always said otherwise, so this is the implementation catching up with its contract, not a redefinition.
- **On the wikitext edit form the errors become one semicolon-joined sentence** instead of one line each. The REST envelope is the reason for the change and only carries one message; this is the cost on the other path.

## Manual Browser Check

1. Open Special:CreateSubject, press **Create subject**, then **Create new**.
2. Name the schema, press **Add property definition**, and set its type to **Relation**. Target schema reads "Target schema is required." straight away, without being clicked into.
3. Press **Continue** without choosing a target. The dialog stays on the schema step; a notification titled with the property name says "Target schema is required."
4. Choose a target and press **Continue**, then **Create subject**. It creates the page and lands on Special:Subject.
5. Open an existing Schema, press **Edit schema**, clear a Relation property's **Relation type**, and save. Held back the same way, rather than failing against the server.

## Considered, omitted

- Merging `IncompleteProperty` into `UnparseableInput`. They are the same shape, produced by the same component, and every consuming block is identical, so one probe would collapse roughly 50 lines and make the guard impossible to half-apply. Left out because the merged name has to change `UnparseableInput` across the Subject editor too, which is outside what this fixes and is a naming call worth making deliberately. Filed as #1456.
- Selecting the offending property when a save is held back, so its inline error is on screen. The notification names it, matching the existing unparseable-input guard.
- Validating the draft against `schemaContentSchema.json` in the frontend rather than the hand-written rule. It would remove the PHP/TS duplication, but the JSON Schema is not served to the client and there is no JSON-Schema validator in the bundle; #1454 is where to decide that.
- Removing `SchemaPicker`'s now-unused `blur` emit. It is a public component and nothing depends on the removal.

> <sub>AI-authored — Claude Code, `Opus 5 (1M context)`; bug reported in chat by @alistair3149, who reproduced the remaining failure and chose the fix scope from four options; diff not yet human-reviewed; every regression test written first and seen failing, and the hardening and guard tests re-checked by reverting each change; 2379 vitest, `make stan`, `make ts-lint`, vue-tsc and `make phpunit filter=Content` (317) green locally, phpcs clean on the changed files; the whole flow walked in a local dev stack and the REST envelope checked directly, including under `uselang=qqx`.</sub>

Reviewed in-session before opening, by six parallel read-only agents (code-review, security-review, review-tests, a design-optimality pass, and correctness and edge-case second opinions). Security found nothing. The fourth commit is that review's output: the edit-token leg, the language premise, the Layout and Mapping handlers, the public-API leak, malformed-definition hardening, and five gaps a mutation pass found — including a guard that shipped with no test at all.

Same-session review, so the goal was reconstructed from the ADRs rather than taken from the PR: ADR 25 governs Subject-value validation through Property Type validators and has no Schema-validate route, so the client-side gate here is not the thing it removed.

</details>


alistair3149:
image

New schema can't be created when a relation property does not have a target schema anymore.
That is hardcoded to relation property, #1454 is filed for a more complete direction.

#1456 is filed as a follow-up instead of this PR to avoid growing the scope.

PR #1458, opened 2026-09-16
Make Subject labels and values findable through the wiki search

JeroenDeDauw:
A page's Subjects live in the `neo` slot and render in the browser, so neither MediaWiki's database search nor
CirrusSearch saw them: a Subject's label or value was findable only if the wikitext repeated it. In user testing,
people could not find Subjects they had just created.

Now indexed for every page: each Subject's label and its text, url, date, date-time, number and select values,
select ones by their option's label. Each Property Type answers what its values contribute through a new
`PropertyType::searchText()`, so boolean and relation return nothing, an extension type decides for itself
(RedHerb's color type indexes its string), and an unregistered type contributes nothing.
`SubjectSearchTextBuilder` in `src/Application/Search/` assembles the text for both backends, handing a type the
Property Definition only when the Schema still declares the property with the Statement's own type; it runs in
the deferred search update after the save, or in CirrusSearch's indexing job. Property and Schema names are left
out.

**CirrusSearch.** A `neowiki_text` field is declared through `SearchIndexFields` and filled through
`SearchDataForIndex2`. Extension registration adds it to `$wgCirrusSearchWeights`, which is what copies it into
the `all` field a plain query matches; CirrusSearch's own hook for that, `CirrusSearchMappingConfig`, is declared
but never fired. Verified against Elasticsearch 7.10 with CirrusSearch REL1_43: hits through the API and
Special:Search, a save from the Subject editor, a deletion.

**Database search.** Subclasses of MediaWiki's MySQL and SQLite engines append the Subject text in `update()`,
folded to lower case as MediaWiki folds the page text. They are registered through the `SearchMappings`
extension attribute, so every route to a core engine (default, explicitly configured,
`$wgSearchTypeAlternatives`) resolves to them, while CirrusSearch and an administrator's own engine stay
untouched. Core schedules its search update only for edits touching the main slot, so a save from the Subject
editor, and an undeletion of such a page, are scheduled from NeoWiki's hooks. Postgres is not covered: its engine
indexes from the page text through database triggers and ignores the text it is handed.

After deploying, existing pages need `rebuildtextindex`, or on CirrusSearch wikis
`CirrusSearch:UpdateSearchIndexConfig` followed by `CirrusSearch:ForceSearchIndex`. Search results were never
permission-filtered; now that Subject values are indexed, the restricted-content page says so.

Tested end to end on MySQL and on SQLite with FTS3; the CirrusSearch hooks through MediaWiki's
`getDataForSearchIndex()` plumbing.

Considered, omitted:

* Following the revision policy: the index would go stale on approvals made without an edit, and MediaWiki's own
  index does not follow it either.
* Indexing values as the UI displays them, such as a formatted date: `searchText()` allows it; nothing asks for
  it yet.
* Moving the RDF value mappers onto the interface the same way; a separate change.
* Search-box title suggestions, facets and filters, typed per-property fields, and Subject values in result
  snippets.
* A configuration flag, until a wiki needs one.

> <sub>AI-authored — Claude Code, `Fable 5.1 (max)`; one-line ask from @JeroenDeDauw after an in-session design pass, redirected twice (a Value-Type switch, then a per-type registry, before the interface method); diff not yet human-reviewed; tests written first, full PHPUnit green locally on MariaDB and on CI for MW 1.43–master, CirrusSearch path exercised against a live Elasticsearch 7.10.</sub>

Design, review and this description by `Fable 5.1 (max)`; the implementation passes and the live Elasticsearch run by `Opus 5 (max)` subagents, one of them finished by Fable after it hit a usage limit. AI code and text reviews were applied before the squash; the CirrusSearch run predates the last two reworks, which changed only how the indexed text is produced.

</details>


JeroenDeDauw:
Hmmm this is currently the second design and I still don't like it. Redesigning it again. PR will be updated soonish

JeroenDeDauw:
LGTM

Goal, architecture and design reviewed, including peeks at the code

Locally manually tested in browser

JeroenDeDauw:
For whoever picks up faceted search:

**Search-bar suggestions.** Not a bug: the dropdown is MediaWiki's title completion on both backends, so a Subject value produces no suggestion but does hit on the results page. Values should stay out. Labels of a page's other Subjects would be worth adding; that waits on https://github.com/ProfessionalWiki/NeoWiki/issues/1313.

**Faceted search.** Filters, ranges and counts need typed per-property fields. They join the same CirrusSearch document as this PR's text field, which stays; WikibaseFacetedSearch does the same and keeps no index of its own. Nothing here gets redone, and no second index is needed unless results should become Subjects rather than pages. Facets are CirrusSearch-only; the database backend keeps just the appended text. The choice then: page-level fields, or a nested object per Subject so a filter cannot combine values from two Subjects on one page. `searchText()` gets a typed sibling at that point.

> <sub>AI-authored — Claude Code, `Fable 5.1 (max)`; one-line ask from @JeroenDeDauw, answering two of his questions from testing this PR.</sub>

alistair3149:
Review PR: #1476

Tested searching for property values and changing them, they are reflected properly in search.
However, user cannot understand why a subject is in the search result since there are no match text in the search result. I know that it is deferred, but worth a mention.

JeroenDeDauw:
Merging to have this for the user testing session as per https://github.com/ProfessionalWiki/NeoWiki/pull/1476#issuecomment-5722379131

Issue #1456, opened 2026-09-16
Merge IncompleteProperty into UnparseableInput: one save-blocking probe, not two


alistair3149:
`UnparseableInput` and `IncompleteProperty` sit in the same directory and declare the same thing:

```ts
export interface UnparseableInput {   // a field holding text its widget cannot turn into a Value
	propertyName: string;
	message: string;
}

export interface IncompleteProperty { // a property definition the wiki will refuse to store
	propertyName: string;
	message: string;
}
```

Both are produced by `SchemaEditor`, both are forwarded unchanged by `SchemaCreator`, and every consumer runs the same five lines on them:

```ts
if ( x !== null ) {
	mw.notify( x.message, { title: x.propertyName, type: 'error' } );
	return;
}
```

No consumer distinguishes them. `SchemaEditorDialog`, `SchemasPage/SchemaCreatorDialog` and `SubjectCreatorDialog` each carry that block twice, once per type, differing only in the comment above it.

## Why this is worth doing

The parallel axis has already cost a defect. When the `incompleteProperty` probe was added in #1455, `SchemasPage/SchemaCreatorDialog.spec.ts` gained the stub, the module-level variable and the `beforeEach` reset — and no test. Deleting that dialog's entire guard left the full suite green. The two sibling dialogs got two tests each for the identical branch; the third was simply forgotten, because the second axis has to be remembered separately at every one of its copies.

Two probes cannot be half-applied if there is only one.

## Proposal

One type, one probe. `SchemaEditor` exposes a single method returning the first property-level reason the save must not proceed — the unparseable input, else the incomplete definition — and each dialog keeps exactly the block it already had.

That removes `IncompleteProperty.ts`, one member from each of two `Exposes` interfaces, the `SchemaCreator` pass-through, three duplicated dialog blocks, and four test-double members.

## The decision this needs

The merged concept is no longer "unparseable input", so the name has to change — and `UnparseableInput` is also used by `SubjectEditor.vue`, `SubjectEditPane.vue` and `SubjectEditorDialog.vue`, which have no incomplete case. Options:

- Rename to something neutral (`SaveBlocker`, `PropertyIssue`, …) across all of it. Honest, and a mechanical change across about ten files.
- Merge the probe only and keep the name. Removes the duplicated blocks, but leaves a type called `UnparseableInput` describing something that is often neither unparseable nor input.

[...]

PR #1459, opened 2026-09-16
Fix typing a property's name in the schema editor

alistair3149:
Typing a property's name in the schema editor lagged, and typing a name another property already has threw. One commit per problem.

**The editor was rebuilt on every keystroke.** `SchemaEditor` keyed `PropertyDefinitionEditor` on the selected property's name, so each keystroke in the name field renamed the property and remounted the whole editor. On Schema:Company that cost 17–25 ms of script per keystroke (3–6 ms now), moved the caret to the end of the name, and dropped initial-value text that could not become a value, so the save guard no longer saw it. The editor is now keyed on selection.

**Duplicate and empty names.** Typing another property's name, including on the way to a longer one ("Status" towards "Statuses"), threw `Duplicate property name`. The property kept its last name while the field showed the new one, so a save stored a name the user could not see. A cleared name was ignored the same way, behind `console.log( 'TODO: show error' )`. The name field now says why, the property keeps its last name meanwhile, and the save is held through the existing `UnparseableInput` guard. A taken name becomes the property's once the property holding it is deleted.

## Worth objecting to

- The name error rides the `UnparseableInput` guard, whose doc now covers text the definition cannot take rather than only unparseable values. Its notification is titled with the last accepted name, e.g. "Statu". How the guards are named is #1456.

## Manual Browser Check

1. `make ts-build`, open Schema:Company, press **Edit schema**, then **Add property definition**.
2. Type `Headcount`, click into the middle of the name and type: the characters land at the caret.
3. Select the name and type `Status`: "A property named "Status" already exists." shows under the field, and the list still reads `Statu`. Press **Save schema**: the save is held with that message.
4. Type `es`: the error clears and the list reads `Statuses`. The console stays clean.
5. Delete back to `Status`, then delete the existing Status property from the list: the error clears and the new property is named `Status`.

## Considered, omitted

- Case-insensitive uniqueness: the Schema format keys properties case-sensitively, unchanged here.
- Giving the editor a fresh instance when a new `initialSchema` arrives whose first property has the selected name: pre-existing, left alone.

> <sub>AI-authored — Claude Code, `Opus 5 (1M context)`; one-line report of the lag from @alistair3149, then asked to fix the duplicate-name throw found along the way as a separate commit; diff not yet human-reviewed; full vitest suite, eslint, stylelint and vue-tsc green locally after rebasing onto #1455, both fixes checked before and after in the browser on a local dev wiki on the pre-rebase revision, where CI was green.</sub>

The review before opening ran in the authoring session: the independent code-review, security, test-quality and design subagents could not start because of a usage limit, so no fresh-context review has happened yet. New logic was mutation-checked locally.

</details>


JeroenDeDauw:
Conceptual conflict with the already merged below PR?
* https://github.com/ProfessionalWiki/NeoWiki/pull/1457

alistair3149:
> Conceptual conflict with the already merged below PR?
> 
> * [Select the generated name of a new property in the schema editor #1457](https://github.com/ProfessionalWiki/NeoWiki/pull/1457)

No it is different and unrelated. This PR is to address two issues that I found with the SchemaEditor:
- Typing on the property name is laggy
- Duplicate and empty property name are silently ignored instead of showing an error

PR #1460, opened 2026-09-16
Open the Subject creator's close confirmations in front of it

alistair3149:
Regression from https://github.com/ProfessionalWiki/NeoWiki/pull/1450. Closing the Subject creator with unsaved changes could open the confirmation hidden behind the creator. As it opened, the confirmation made the creator inert, so the close button seemed to do nothing and the creator stopped responding. Two paths hit it: closing the Schema step of a creator reopened after a Schema had been picked, and closing the Subject step after drafting a new Schema.

Codex dialogs teleport into one container in mount order, and a later one paints on top. Since #1450 each step of the creator is a dialog of its own, mounted when the step is reached, while both confirmations were mounted once at the start, so a step dialog mounted after them covered them. Each confirmation now mounts with its step's dialog, directly after it, where the other editor dialogs place theirs.

The second commit fixes a related case, also on master: a close requested while a picked Schema loaded left "Discard unsaved changes?" pending, and it came back on its own after the creator closed. Opening the Subject step now withdraws it.

Considered, omitted: mounting each confirmation only while it is open, which stacks it above any dialog but drops Codex's closing fade.

## Manual Browser Check

1. On Special:CreateSubject, pick any Schema, then close the creator. It closes without asking.
2. Reopen it with **Create subject**, choose **Create new**, type a Schema name and press the close button: "Discard unsaved changes?" shows in front, and **Discard** closes the creator.
3. Reload the page. Choose **Create new**, type a Schema name, press **Continue**, then close: the question offering **Save schema only** shows in front, and **Keep editing** returns to the Subject step.

> <sub>AI-authored — Claude Code, `Opus 5 (1M context)`; bug report in chat from @alistair3149, no revisions; diff not yet human-reviewed; regression tests seen failing without each fix, full vitest suite, lint and vue-tsc green locally, both flows re-checked in a headless browser on a dev wiki; same-session review, with its code-review and edge-case passes cut short by a usage limit; CI pending.</sub>


alistair3149:
image

PR #1462, opened 2026-09-16
Mount the close confirmations only while they ask

alistair3149:
Follows-up to https://github.com/ProfessionalWiki/NeoWiki/pull/1460, now merged.

Codex stacks open dialogs in the order they mounted, so a confirmation mounted alongside its host opens behind any dialog the host opens later, under a dialog Codex has marked inert: the close button appears to do nothing. #1460 answered that inside the Subject creator, by keeping each confirmation next to the step dialog it confirms — an arrangement nothing enforces.

`CloseConfirmationDialog` and `SchemaAbandonmentDialog` now mount only while they ask, so they open in front wherever their host puts them. That covers all six dialogs that ask before closing, the creator's template goes back to plain siblings, and callers keep the same `open` prop.

It costs the closing fade, which Vue skips for a component that is unmounting; the one it opens with still plays.

Upstream's own answer, a confirmation nested in its parent's slot, needs Codex 2.3 and so MediaWiki 1.45, while NeoWiki supports 1.43.

## Manual Browser Check

1. On Special:CreateSubject choose **Create new**, type a Schema name and press the close button: "Discard unsaved changes?" is in front, the creator behind it does not respond, and **Keep editing** hands it back.
2. Reload. **Create new**, a Schema name, **Continue**, then close: the question offering **Save schema only** is in front, and **Discard all** closes everything.
3. The other dialogs that confirm before discarding still do: the Schema editor and creator, the Layout editor and creator, and the Subject editor.

> <sub>AI-authored — Claude Code, `Opus 5 (1M context)`; design discussed with @alistair3149, no revisions to the change; diff not yet human-reviewed; tests written first and seen failing, and against two refactors that would bring the bug back; full vitest suite, lint and vue-tsc green locally; both paths and the fade cost checked in a headless browser; reviewed in-session across five passes, whose test gaps are fixed here.</sub>


JeroenDeDauw:
Did you check this in the browser yourself?

alistair3149:
Yes I checked it already

PR #1465, opened 2026-09-16
Reduce the subjects editor's left pane to the subjects being edited

### LINKED ISSUE #1447: Reduce the subjects editor's left pane to the list of Subjects being edited
For https://github.com/ProfessionalWiki/NeoWiki/issues/971. Supersedes https://github.com/ProfessionalWiki/NeoWiki/issues/1327.

**Problem.** Most edits touch one Subject, yet the dialog opens with a tree of that Subject's neighbourhood in the pane a reader meets first. A newcomer has to work out what the tree is for before finding out they can ignore it, and it duplicates the relation fields on the right, which already list the same targets and open them.

**Change.** The left pane becomes a plain list of the Subjects open in the dialog, in opening order, each with its unsaved marker; selecting one switches the pane on the right. It is hidden while only one Subject is open and appears when a second one opens, from a relation's edit control or a mid-edit creation. Nothing is fetched for it: it lists what the dialog holds. The multi-Subject creator (https://github.com/ProfessionalWiki/NeoWiki/issues/1402) uses the same list.

**Out.** Navigating to Subjects the dialog has not opened, prepopulating targets, relation-property grouping, and a depth cap. The editor no longer waits on the relations endpoint (https://github.com/ProfessionalWiki/NeoWiki/issues/1324). Reaching a Subject several relations away is an exploration need, to be designed on its own if the September user-testing sessions show the need, judged after them. https://github.com/ProfessionalWiki/NeoWiki/pull/1425 is closed unmerged; its branch is the starting point if they do.

**Alongside.** The per-relation edit control becomes the only way a second Subject opens, so its reading as "change the link" is fixed in the same pass: https://github.com/ProfessionalWiki/NeoWiki/issues/1448.

> <sub>AI-authored — Claude Code, `Fable 5.1 (max)`; filed on @JeroenDeDauw's instruction, recording the 2026-09-14 DevTeam decision with scope confirmed by him in-session, one blind-judge pass; text not human-reviewed; checked against the meeting record, the dialog code on master and the linked threads.</sub>

### LINKED ISSUE #1448: The relation field's edit icon reads as changing the link, not as opening the target
For https://github.com/ProfessionalWiki/NeoWiki/issues/971.

**Problem.** A filled relation field shows the target's name in a lookup with a pencil beside it. A user who wants to point the relation elsewhere reads the pencil as "edit this value", clicks it, and lands in the target Subject's own editor pane instead. Once the left pane lists only open Subjects (https://github.com/ProfessionalWiki/NeoWiki/issues/1447), this control is the only way a second Subject is opened, so it has to read as "open the target" and never as "change the link". Recorded as a fail-once cost in https://github.com/ProfessionalWiki/NeoWiki/issues/1328.

**Done when** the only way to open a target from a relation field is an action that names the target and says it opens it, repointing the relation needs no click on that action, and both work by keyboard and on touch. Observed in the September user-testing sessions.

**Directions, undecided; picked here before implementation.**

1. The target is a record chip, not text: clicking the chip opens it in the editor, its × unlinks, and typing in the field replaces a single-valued relation's target or adds to a multi-valued one. No edit glyph exists. This is how Airtable, Notion and Salesforce show linked records.
2. Keep a control beside the field but make it a labelled button naming the action and the target, such as "Open Amsterdam", with an open or arrow glyph rather than the pencil.
3. Move opening the target off the value row: into the target's preview card, or into the picker's dropdown footer next to the planned "create new" action.

> <sub>AI-authored — Claude Code, `Fable 5.1 (max)`; filed on @JeroenDeDauw's instruction, problem in his words from the 2026-09-14 DevTeam sync and https://github.com/ProfessionalWiki/NeoWiki/issues/1327#issuecomment-5653442962, directions drafted here and not yet chosen by him, one blind-judge pass; text not human-reviewed; control read off master's relation input components, prior art from recollection, not re-checked.</sub>

alistair3149:
Fixes https://github.com/ProfessionalWiki/NeoWiki/issues/1447
Fixes https://github.com/ProfessionalWiki/NeoWiki/issues/1448

## The left pane becomes the Subjects being edited

#1447 asked for a plain list of the Subjects open in the dialog, in opening order, hidden until a
second one opens. The pane instead opened on a tree of the root Subject's relation neighbourhood,
fetching every target and every Schema it needed to draw rows that duplicated the relation fields
beside them.

It now lists the Subjects the dialog is holding, each with its unsaved marker. Nothing is fetched for
it. It is a `listbox` of `option` rows, following `PropertyList` in the schema editor — which opens
inside this same dialog — rather than inventing a pattern.

## The relation field opens its target, instead of editing it

#1448 asked that the control never read as changing the link, which this pass has to answer because
it is now the only way a second Subject opens. The pencil becomes a quiet arrow over the end of the
field, naming what it opens. Repointing a relation is typing in the field, as before.

The arrow is gated on the target the relation holds, **not** on the lookup's own selection — Codex
drops that on the first keystroke matching neither the selected item's label nor its value. Gated the
other way it vanished while the user typed, leaving the Subject it points at with no route to it at
all, which the tree used to provide.

The overlay is grid placement, not positioning: a positioned ancestor there recaptures the containing
block Codex gives its menus inside a dialog, silently clipping every dropdown.

## What goes, and why that is safe

The walk, its depth cap, the "Not linked here" rows and the generic `NeoTree` — about 2,600 lines,
three spec files among them. `NeoTree` had no other caller: not in `public-api.ts`, not in `docs/`,
not in RedHerb. `relationTargetsOf` folds into `SubjectDraftGraph`, its only remaining consumer, and
gains a test for the rule that only relations the Schema *declares* count — load-bearing for the save,
and previously covered only through the deleted tree.

Two behaviour changes to know about: the pane no longer appears for a relation target nobody has
opened, and a Subject reachable by two routes is one row rather than two.

## Verification

2328 tests, lint and build green, and green on the first commit alone. Twelve ablations — the gate
forced on and off, the list reversed, the dot stripped of its label, the arrow keys robbed of focus
movement, the Schema no longer deciding what a relation is — were each caught by a test.

### Manual browser check

Run against the head commit on a dev wiki with the demo data; all steps passed.

1. Open the editor on `ACME Amsterdam HQ`: no left pane and no divider, although `City` holds
   `Amsterdam`.
2. Click the arrow in `City`: `Amsterdam` opens and the list appears with both Subjects.
3. Edit a field in `Amsterdam`: its row takes the amber dot, the other keeps none.
4. Arrow keys move selection, focus and the form together, stopping at the ends; a mouse click
   leaves no focus ring; the divider still narrows the pane to its floor.
5. On `ACME Inc`, whose relations hold several targets each: every filled row carries its own arrow,
   the trailing empty row none. Tab reaches it; clicking mid-field still types.
6. Type over a filled field: the arrow stays, still naming the stored target. Empty the field and it
   goes with the relation.
7. The dropdown is not clipped by the dialog body.

Considered, omitted: the tree-versus-list discussion, which is on #1447, and the deleted components'
design, which stays in the history and on #1425.

> <sub>AI-authored — Claude Code, `Opus 5 (1M context)`; built from the two issues at @alistair3149's direction, redirected three times (the list's shape, the open control's design, its icon size); reviewed in-session by five read-only passes whose fixes are folded in, so the authoring session reviewed itself; diff not yet human-reviewed; 2328 TS tests, lint and build green locally, twelve ablations each caught by a test, and the browser check above run on a dev wiki.</sub>


alistair3149:
Tested both the subject edit and creation flow


https://github.com/user-attachments/assets/3c126353-1e83-46d2-b730-4325ae0ca48f

PR #1467, opened 2026-09-16
Serve each Schema as a JSON Schema document

JeroenDeDauw:
`GET /neowiki/v0/schema/{schemaName}/json-schema` serves a Schema as a JSON Schema (draft 2020-12)
document (`application/schema+json`) describing a Subject of that Schema in the JSON the Subject
endpoints accept and return. Import pipelines can generate client models from it and check data
before uploading, which ECHOLOT partners have asked for.
[ADR 009](https://github.com/ProfessionalWiki/NeoWiki/blob/master/docs/adr/009-move-away-from-json-schema.md)
names this endpoint as the way to serve Schemas in a standard format.

The contract, documented in `docs/api/json-schema.md`: a Subject that validates against the document
draws no Violations for its values, except for what needs the wiki's content or JSON Schema cannot say
(relation targets, date ranges, whitespace, Constraints of extension-registered Property Types). The
reverse does not hold: the document rejects some payloads the write endpoints accept, `select` labels
and undeclared property names above all. The Schema is read through the same lookup as
`GET /schema/{schemaName}`, so the read gate and the published revision are the same, and a `404`
covers both a missing Schema and one the caller may not read.

## Decisions

* Every Constraint is emitted whatever its severity. Validity means "no Violations"; whether one
  blocks a write is the wiki's enforcement setting, which `POST /subject/validate` answers.
* `statements` rejects undeclared property names, although the wiki stores them: for a client
  generating a payload, a mistyped name is the common mistake, and the document describes the Schema.
  The top level stays open for `comment`, `pageTitle`, `id` and the fields a read adds, and requires
  only `statements`, since the replace body carries no `schema`; when present, `schema` must equal the
  Schema's name.
* `url`, `date` and `dateTime` get a `pattern` that is the type's own: `UrlType` now validates with
  one case-explicit pattern constant instead of a scheme check plus an `/i` regex (same accepted set,
  checked on a fuzzed corpus), and the date classes expose theirs likewise. `format: uri` was rejected
  because it disagrees with the wiki in both directions; the date types get `format` as well.
* Each Property Definition states its own value subschema through a new
  `PropertyDefinition::toJsonSchema()`, beside `nonCoreToJson()`, the existing per-type serializer
  extension point; it restates what the type's `validate()` checks. The serializer only assembles the
  document, and an extension's type is described as fully as its definition class describes it
  (RedHerb's `ColorProperty` does). A type no extension registered here gets `true`. Extension
  Property Definition classes must implement the new method.
* `$id`, `title` and the `schema` const carry the Schema's page title, so a document fetched as
  `Validation_Demo` still describes Subjects of `Validation Demo`.
* The handler calls the Schema lookup directly rather than going through `GetSchemaQuery`, like the
  RDF export endpoints.

Two behaviour changes outside the endpoint: a Subject without Statements reads back with
`"statements": {}` rather than `[]`, as `docs/api/subject-format.md` specifies; and a reserved name
such as `page` answers `404` here instead of `500`.

## Tests

`JsonSchemaMatchesValidatorTest` keeps the projection honest: a table of (property definition, wire
value) rows across every built-in type and Constraint, each validated with opis/json-schema (already a
dependency) and run through the validator, asserting the two agree. Rows where they cannot agree carry
the reason and are asserted to disagree, so those rows are the list of what the document leaves out.

## Considered, omitted

* An all-Schemas bundle, `default` annotations, severity annotations, a link from Schema pages: none
  has a consumer yet.
* Validating the generated documents against the draft 2020-12 meta-schema, which would mean checking
  in its eight-file bundle.

> <sub>AI-authored — Claude Code, `Fable 5.1 (max)`; task from @JeroenDeDauw via the delegate-issue skill, redirected twice (a per-type extension point instead of a serializer that binds to the built-in types; a serializer readable without comments), plus one design change he approved from a review finding; diff partly read by @JeroenDeDauw mid-session, not yet reviewed as a whole; full PHPUnit suite, phpcs and phpstan green locally and on CI (16 checks), served documents byte-identical across both design moves, about 250 targeted mutations over three review-tests passes all caught after fixes, endpoint checked against a dev wiki.</sub>

Design brief, holistic passes and the final serializer rewrite by `Fable 5.1 (max)`. Implementation by `Opus 5 (max)` subagents running fix-issue, then an `Opus 5 (max)` pr-review pass whose five confirmed contract findings (canonical Schema naming, the empty `statements` map, relation property values, the relation id grammar, and a `required` exclusion later dropped again) were folded in, three `/simplify` rounds and two `/review-tests` runs on `Opus 5 (max)`, and a fresh-context design review on `Fable 5.1 (max)`. The value subschemas moved twice: from an `instanceof` dispatch in the serializer onto `PropertyType::toJsonSchema()` at the lead's request, then onto `PropertyDefinition::toJsonSchema()` after an altitude finding showed every implementation read only the definition. The last implementation pass ran on a `Fable 5.1 (max)` subagent after the Opus spend limit tripped mid-move.

</details>


JeroenDeDauw:
Doing another design/architecture pass over it

JeroenDeDauw:
LGTM

PR #1471, opened 2026-09-17
Tell namesakes apart in the subject picker

alistair3149:
A relation field's Subject picker showed nothing but the label, so it could offer several rows reading identically with no way to choose between them — several `Appellation` Subjects on one page, say. Two unsaved Subjects created from its own "Create a new X" option were always identical, being named after their Schema. Raised in user testing.

The picker appears in every relation field: the subject editor (infobox edit button, Manage subjects, nested tree editor), a relation property's default value in the schema editor, and RedHerb's example subject finder.

## What a row shows now

| the menu also holds | the row shows |
|---|---|
| no other row with this label | the label |
| another row with this label, on a different page | the label, and that page on a second line |
| another row with this label, on the same page | both, plus the full Subject id in brackets beside the label, never shortened |
| another row with this label, and this row's page only repeats the label | the label alone — the page would print it twice |
| another *unsaved* row with this label | the label plus the id; an unsaved Subject is on no page |

A row whose page title is unavailable loses the page line and falls through to the id, rather than showing an empty one.

The comparison runs in the frontend, over the rows the menu is about to show, rather than server-side: which rows need what is a property of the answered set. `disambiguateSubjectLabels` holds the rule and is tested in `tests/presentation/disambiguateSubjectLabels.spec.ts`; its wiring into the menu in `tests/components/common/SubjectPicker.spec.ts`.

## Supporting changes

- `GET /neowiki/v0/subject-labels` returns each result's prefixed `pageTitle`.
- `SubjectLabelResult` gains `pageTitle` as a required field. Breaking for anything implementing `SubjectLabelSearch` outside the extension; nothing in-repo does.
- Equal names are now ordered by id as well, since ordering by name alone left the membership of a run of namesakes undefined under the limit.
- A page name that is not a string now names no page instead of throwing.

## Manual Browser Check

On a wiki with the demo data, using the `Artist` Schema and any Artwork page's **Creator** field (infobox edit button):

1. Create two Artists both labelled `Jan Vermeer`, on pages `Jan Vermeer of Delft` and `Jan Vermeer the Younger`. Type `Jan` in **Creator**: each row carries its page, neither carries an id.
2. Add a third Artist labelled `Jan Vermeer` to one of those pages via Manage subjects. Search again: the two sharing that page now also carry their ids; the third does not.
3. Create an Artist labelled `Jan Vermeer` through `Special:CreateSubject` without naming a page, so the page is titled after the label. Search again: that row shows no page line.
4. Clear the **Creator** field and pick `Create a new Artist` twice without typing a name. Empty the field and focus it: both unsaved rows read `Artist` and carry their ids.

## Follow-ups, not in this PR

- The same collision exists wherever several Subjects are listed side by side — the editor's open-subject list, `SubjectRow`, `RelationDisplay`. The rule is reusable there, but the frontend `Subject` carries no page title yet.
- A namesake beyond the answered window (`limit` defaults to 10) goes unmarked. A `truncated` flag would need care: deciding ambiguity from over-fetched rows the caller may not read would leak existence across ACLs.
- Reading the page title through `SubjectHostingPageResolver` rather than the graph would also drop candidates resolving to nothing, fixing the *deleted Subject still offered* report from the same testing session.
- Open product question: two unsaved namesakes identify themselves by id, where a "not saved yet" marker might read better.

<sub>AI-authored — Claude Code, `Opus 5 (1M context)`; design question from @alistair3149, 3 rounds of direction; diff not yet human-reviewed; test-first throughout, each new test seen failing without its production code; 2343 vitest and 2265 non-database PHPUnit tests, phpcs, phpstan, eslint, stylelint and vue-tsc green locally, full CI green; browser-verified on a dev wiki, including the unsaved-Subject flow.</sub>


alistair3149:
This issue was raised in the ECHOLOT Developer meeting on 2026-09-16.

Screenshot of the UI:
image

malberts:
I wonder if a "simpler" solution would be to just copy Wikibase's item description. Perhaps not as a manually created property, but rather as a dedicated field on the subject, similar to the label. It is always there, and optional. Then there is exactly one way to provide the additional information to the lookup, and the UI does not change depending on the available data.

It does not seem like nice UX to show those internal IDs. Both for the subject, and for nameless pages (e.g. https://neowiki.dev/wiki/SEyWe2j23BUgeDM). While it does visually indicate a difference, it does not provide additional meaningful information. If at some point we want to make the lookup also search IDs, in case a user wants to find it that way, then that is a separate question.

With the current dynamic UI, I can imagine some minor confusion:
* do some of those subjects not have pages?
* are those IDs part of the subject label? Or why do some not have IDs?
* I am familiar with Wikibase, is "Vermeer workshop" the description of "Jan Vermeer"?

Always showing all those extra values would make the UI consistent at least. But with a dedicated description field, it always carries meaningful information (because that is the purpose of the field), and if it is not there, then there is no description. And in the case of missing descriptions, that seems more like a data fixing task, not a NeoWiki problem.

alistair3149:
> I wonder if a "simpler" solution would be to just copy Wikibase's item description

I considered subject description but I didn't go with it for a few reasons:
- We still need to rely on something else to disambiguate the subjects when the description doesn't exist
- Subject description requires change to the data modeling (I also recalled there was a request for it from ECHOLOT, can't locate the issue), that would be a bigger change since we also need the data modeling and the UI to support the change
- We can always promote subject description to replace the page name when it exists.

> It does not seem like nice UX to show those internal IDs

Yes I agree. That's why it is only shown when we have both the subject label and page collision.
Internal ID is the only unique and reliable identifier that we have, and the UI only show it when it provides a difference.

> do some of those subjects not have pages?

The impact is limited. Subjects can't exist without a page, so the confusion would be short-lived.
If NeoWiki reconsiders that ADR, then it would be time to reconsider.

> are those IDs part of the subject label?

It is a valid concern. That's why it is styled slightly differently. But I suppose that using different font weight or font size would make it more obvious. I will amend that change.

> Or why do some not have IDs?

The confusion would only happen on the first time when they see the ID.
The alternative is that we always show the ID, which is not nice UX as you mentioned. It is the lesser of the two evils.

> I am familiar with Wikibase, is "Vermeer workshop" the description of "Jan Vermeer"?

Yeah we can make the it more obvious perhaps. For now I would align with the `Stored on PAGENAME` convention that we have in SubjectEditor. We can reconsider it holistically in the future if needed.

> And in the case of missing descriptions, that seems more like a data fixing task, not a NeoWiki problem.

We can treat it that way but this would be bad default UX to not be able to disambiguate subjects. And like I mentioned above, we can always add subject description on top of what is being done here.

JeroenDeDauw:
Some quick thoughts. I did not look at everything here yet.

* Showing page titles seems likely problematic for subject-first wikis (where people don't want to care about pages)
* I'm currently doing work on the search index to move it from page-focused to subject-focused, and on subject-first wikis we want search results to go to special:subject/id.
* Great we are working on disambiguation, relevant for real usage
* Would also hold off on adding a description like in wikibase to the data model. We can do simpler stuff now without precluding adding that later.

JeroenDeDauw:
@alistair3149 did you consider my above comments? does the behavior in the pr work sensibly for both page-first wikis and subject-first ones?

alistair3149:
> Showing page titles seems likely problematic for subject-first wikis (where people don't want to care about pages)

We probably need #1518 first to determine an approach
Nothing to add for the other parts of the comment

JeroenDeDauw:
The screenshots on that PR look good. That PR is high up my list of things to review but maybe not happening today

PR #1472, opened 2026-09-17
Stop offering to open a target the field is no longer showing

alistair3149:
A relation field that already points at a Subject shows a right-arrow at its end, which opens that Subject. The arrow followed the *stored* target, so it stayed while you typed a search over the field — offering to open a Subject whose name was no longer on screen.

The arrow now follows what the field is showing:

| the field shows | arrow |
|---|---|
| its target's name | shown, and its tooltip names that target |
| a search you are typing | gone |
| only whitespace | gone |
| nothing yet, while the target's name is being fetched | gone |

Picking a row brings it back, naming what you picked.

## Manual Browser Check

On any Artwork page whose **Creator** already points at an Artist, press the infobox edit button. The arrow should be present in 1 and 4, absent in 2 and 3:

1. The field shows the target's name.
2. Type over the name.
3. Reduce the text to a single space.
4. Type enough to match, and pick the row.

## Known consequence

Typing over a target and leaving the field without picking keeps the relation pointing where it did, but no longer offers a way to navigate to it: clearing the field empties the relation rather than restoring what it held, so the target can only be reached again by retyping its name.

<sub>AI-authored — Claude Code, `Opus 5 (1M context)`; bug reported by @alistair3149, one round of direction, then reworked under multi-agent review; diff not yet human-reviewed; test-first, every new test mutation-checked individually; 2335 vitest tests green on a serial run, eslint, stylelint and vue-tsc clean, full CI green; browser-verified on a dev wiki across all four field states.</sub>


alistair3149:
https://github.com/user-attachments/assets/03e8d7d7-c57c-40d1-a042-23d21a6084ae

Issue #1022, opened 2026-07-11
Imported revisions do not update the graph projection


JeroenDeDauw:
When a page carrying a NeoWiki content slot is exported and imported into another wiki, its Subjects do not appear in the graph. Queries and views on the target wiki return nothing for those pages.

The import itself is fine: the page and its slot content arrive intact (the revision slot is the source of truth, [ADR 4](docs/adr/004-use-dedicated-slot.md)), and running `RebuildGraphDatabases.php` afterwards populates the graph correctly. What is missing is the projection update at import time.

## Cause

The projection is kept in sync by the hooks registered in `extension.json`:

- `RevisionFromEditComplete` -> `StoreContentUC::onRevisionCreated()`
- `PageDeleteComplete` -> `StoreContentUC::onPageDelete()`
- `RevisionUndeleted` -> `StoreContentUC::onPageUndelete()`

Imported revisions go through the import path rather than the edit path, so `RevisionFromEditComplete` never fires and `onRevisionCreated()` is never called for them.

## Fix

Imported revisions need to reach `onRevisionCreated()` too. Candidate hooks: `AfterImportPage`, or a revision-insert hook that covers every path. Worth checking whether one hook can subsume the edit, undelete and import cases instead of adding a third special case.

## Workaround

Run `RebuildGraphDatabases.php` after importing. Heavy if only a handful of pages were imported.

Export/import between wikis is a normal workflow (for example maintaining content centrally and distributing it to other wikis), so imported pages should be queryable without a full graph rebuild.

> @JeroenDeDauw asked in one line for this to be filed as an issue; written autonomously and not reviewed before posting. Reported by a user; root cause traced by reading the hook registration in `extension.json` and `NeoWikiHooks.php`, but not reproduced locally.
> Context: the NeoWiki extension source and its ADRs.
> <sub>Written by Claude Code, `Opus 4.8 (max)`</sub>

PR #1476, opened 2026-09-17
Correct what reaches the Subject search index

alistair3149:
Follows-up to https://github.com/ProfessionalWiki/NeoWiki/pull/1458

Four ways a page could end up with the wrong search-index entry, or none.

**Undeletion indexed the wrong revision.** `onPageUndeleteComplete` indexed `$restoredRev`, the last restored
revision, which a partial restore does not make current: restoring history older than a live page's current revision
rewrote that page's index entry from a stale revision's wikitext, so the page's own text stopped matching and the
restored revision's text started matching. The method's docblock already says why `$restoredRev` is the wrong
revision to read, which is why the graph projection beside it reads from the primary database instead.

**An import that changed only the Subjects never reached the index.** Imported revisions do not go through the edit
path, and the importer inherits the main slot when the imported wikitext matches what the page already holds — the
shape a save from the Subject editor exports — so core skipped indexing and nothing else scheduled it. This is what
#1022 was for the graph. Both now schedule the update for the page's current revision, read from the primary
database, and only when its main slot is inherited, so no page is indexed twice.

**An unreadable Subject slot took the page's own text down with it.** MediaWiki's JSON content model accepts any
valid JSON, so a `neo` slot can hold a bare string or a scalar that no NeoWiki write path produces but an import or
a hand-edit can leave behind. `SubjectContentDataDeserializer` passed whatever `json_decode` returned into a
parameter typed `array`, raising a TypeError that `SubjectSlotReader` does not catch. That used to cost one page its
rendering; now the trait runs for every page an engine indexes, so it also aborts `rebuildtextindex` — which empties
the index before refilling it, leaving the wiki with no search at all. Reads now show the Subjects they can, as
`SubjectSlotReader` already says they do.

**Subject text skipped `normalizeForSearch()`.** Core applies it to page text before indexing and to query terms
before matching; the Subject half applied only `lc()`. It folds full-width Roman characters for every language and
converts script for the variants that set one, so a full-width label was indexed in full-width bytes while the query
for it was folded to half-width, and the two never met. Chinese and Serbian wikis were affected the same way.

`rebuildtextindex` now carries what it costs: search finds nothing while it runs, and it refuses to run on
PostgreSQL, which the upgrade page asked for without that caveat.

Each fix has a regression test that fails without it. The first two were also reproduced by hand on a dev wiki.

## Found in review, not fixed here

Recorded so they are not lost; none blocks this branch.

* Schemas are resolved through the request authority, so `rebuildtextindex` runs anonymous and, on a read-restricted
  wiki, silently drops every select value from the index it just rebuilt. Verified chain, no clean fix to hand.
* Revision deletion never refreshes the index, and `rebuildtextindex` reads the main slot at public audience, so a
  suppressed Subject value stays searchable until a Subject-changing edit. `onArticleRevisionVisibilitySet` already
  exists for this class of problem.
* Deleting a page leaves its row in `searchindex` (core builds the update with page id 0). Not query-exposed, since
  the engine joins `page`, but it survives into dumps.
* Renaming a select option's label leaves the old label indexed on every page using it: nothing re-indexes a
  Schema's Subject pages when the Schema changes.
* `date-time` values are indexed in a form no query produces — `2026-09-17T14:30:00Z` tokenises with a `17t14`
  component — while `maintenance.md` lists them as findable.
* `SearchUpdate` calls `update()` once per configured engine, so with `$wgSearchTypeAlternatives` the whole Subject
  text is rebuilt N times per page; core memoizes `getLatestPage()` for exactly this reason.
* `scheduleSearchUpdateOfCurrentRevision()` re-reads a revision `rebuildFromPrimary()` fetched on the line above.
  Fixing it means widening `PageRebuilder`'s return contract.
* Adding `searchText()` to `PropertyType` fatals third-party Property Types at autoload, with no note in
  `upgrading.md`.
* Of the tests covering the database engines, only `DatabaseSearchEndToEndTest` fails if the trait is removed from
  both production classes; the others drive a spy.

> <sub>AI-authored — Claude Code, `Opus 5 (1M context)`; review of #1458 requested by @alistair3149 via /pr-review, defects found across review subagents and verified against MediaWiki core before fixing; diff not yet human-reviewed; every fix has a regression test seen failing first, and the non-database suite, the search, import, deserializer and subject-repository test classes, phpcs and phpstan are green locally on MariaDB.</sub>


JeroenDeDauw:
huh. @alistair3149 you marked this as draft after requesting review. Ready for review?

alistair3149:
It is not ready yet, I found some additional blocking issues that I have to go through

JeroenDeDauw:
Btw some thoughts here: correct behavior on undelete is nice and important. But it is NOT a blocker. Right now it never works on master. Going from 0% to 95% is fantastic. Blocking that on the remaining 5% is nonsensical to me. So for this type of issues, I would merge and then either do follow up from master or file an issue. Sometimes the last 1% isn't important/urgent. And indeed, for our current urgent cases things like undeletion simply do not matter, so blocking on supporting undeletion is harmful.

alistair3149:
The blocking issue was not the undelete part, it was the how the unreadable subject can take down the search index with it.
At the time I was identifying the severity of the issues, hence I mentioned that there are additional blocking issues.

PR #1477, opened 2026-09-17
Cache the MediaWiki checkout, and stop the cache pinning the version under test

alistair3149:
Two independent changes, one per commit.

**Cache the MediaWiki checkout in the image build.** Checking MediaWiki and its ~40 submodules out costs 112 of the job's 324 seconds and repeats work that rarely changes: over the last 58 days REL1_43 advanced 37 times against 198 master pushes, so 81% of runs rebuild a tree an earlier run already produced. Measured on this branch:

| run | total | |
|---|---|---|
| before | 324s | |
| cache miss | 311s | writes the 317 MB entry, +5s to save |
| cache hit | 249s | checkout and submodule steps skipped |

Keyed on the commit `git ls-remote` resolves REL1_43 to, so a MediaWiki update invalidates the entry, and the checkout pins that same commit so the key always describes the tree it names.

**Stop the cache pinning the version under test.** The key carried a hand-bumped `-v2` suffix and no commit, so it never moved on its own. `Install MediaWiki` is skipped on a hit and a cache is only written on a miss, so the entry created on 2026-07-29 was still being served seven weeks later — PHPUnit and PHPStan were both testing against that day's MediaWiki. Both jobs now key on the resolved commit. A miss costs about 25 seconds.

That immediately surfaced real drift the stale cache was hiding. Against today's MediaWiki master, `PHPUnit: MW master` reports `Tests: 3661, Errors: 6`, all of them:

```
TypeError: ModuleManager::__construct(): Argument #4 ($jsonLocalizer) must be of type
MediaWiki\Rest\JsonLocalizer, MediaWiki\Rest\ResponseFactory given,
called in tests/phpunit/EntryPoints/REST/ModuleSpecHandlerNeoWikiTest.php on line 63
```

Production code also reaches `Database::getCacheSetOptions`, deprecated in MediaWiki 1.47, from `CachingSchemaLookup.php:101`. Both are left alone here: that job is `experimental: true` and does not block, and fixing them is not this change. They want their own issue.

Verified by dispatching the image build on this branch twice, a miss then a hit, both green through docker build, MediaWiki install and smoke test. The image-push steps are gated on `refs/heads/master`, so neither published anything.

> <sub>AI-authored — Claude Code, `Opus 5 (1M context)`; follow-up ask from @alistair3149, no revisions; diff not yet human-reviewed; measured hit rate and tree size from run history before building, then confirmed both commits on real CI runs — including a failed first attempt where the resolve step inherited the job's `working-directory: mediawiki` before that directory existed.</sub>


JeroenDeDauw:
LGTM but CI fails

alistair3149:
> LGTM but CI fails

It needs #1523 because this PR fixed the master CI.

PR #1478, opened 2026-09-17
Let Dependabot update GitHub Actions versions

alistair3149:
The repository has no Dependabot configuration, so nothing raises action versions. The Dependabot pull requests it does get are security updates, which run without a config file, and they only cover npm. Action versions therefore move only when somebody notices — `actions/cache` sat on v5 from v6's release in June until now.

Enable version updates for the `github-actions` ecosystem, grouped into a single weekly pull request so it arrives as one review rather than six. All six actions in use live in `.github/workflows/`, and there are no composite actions, so one entry at `/` covers them.

This does not affect the existing npm security updates, which are independent of this file. It deliberately does not enable npm *version* updates, which would be a much larger change in review volume and is a separate decision.

> <sub>AI-authored — Claude Code, `Opus 5 (1M context)`; follow-up ask from @alistair3149, no revisions; diff not yet human-reviewed; verified the absence of any tracked Dependabot config and that the existing Dependabot pull requests are security updates, and inventoried the actions in use; the config itself cannot be exercised before it is on the default branch.</sub>


alistair3149:
@malberts or we can use Renovate if you think that fits better, just need a mechanism to update dependencies automatically.

malberts:
There is an outstanding PR for Renovate: https://github.com/ProfessionalWiki/NeoWiki/pull/1223

alistair3149:
Superseded by #1223

Issue #1201, opened 2026-07-27
Subjects store the schema name as typed, so a miscased spelling splits Neo4j labels and RDF classes


alistair3149:
Schema **lookup** is case-insensitive on the first letter, because a Schema resolves through a
MediaWiki title. Everything downstream compares the Subject's **stored** spelling byte-for-byte.
Nothing canonicalizes in between, so a Subject created with `"schema": "person"` against
`Schema:Person` validates cleanly and then diverges from its siblings in the graph and in RDF.

## Why the two spellings are one Schema

`extension.json` declares `NS_NEOWIKI_SCHEMA` (7474) without `capitalLinks: false`, and NeoWiki sets
no `CapitalLinkOverrides`, so `$wgCapitalLinks` applies. Confirmed against a dev wiki:

```
$ curl -s ".../api.php?action=query&titles=Schema:person|Schema:Person&format=json"
{"query":{"normalized":[{"from":"Schema:person","to":"Schema:Person"}],
          "pages":{"14":{"pageid":14,"ns":7474,"title":"Schema:Person"}}}}
```

Only the **first character** is folded, so this affects `person` vs `Person`, but not `Person name`
vs `Person Name`.

## Where the raw spelling gets in and stays

- `EntryPoints/REST/CreateSubjectApi.php:88` — `schema` is a required `string` with no enum and no
  normalization; line 46 passes `$body['schema']` straight into the request object.
- `Application/Actions/CreateSubject/CreateSubjectAction.php:114` —
  `$schemaName = new SchemaName( $request->schemaName )`, stored on the Subject as given.
- `Domain/Schema/SchemaName.php` — rejects only empty strings and the reserved names
  `page`/`subject` (via `strtolower`). No case normalization.
- `Application/Validation/ProposedSubjectValidator.php:36` — resolves with
  `$this->schemaLookup->getSchema( $subject->getSchemaName() )`. Because the title folds, this
  **succeeds**, so the `schema-not-found` violation never fires and the Subject is accepted.
- `Persistence/MediaWiki/Subject/SubjectContentDataDeserializer.php:60` — the revision slot
  round-trips the stored spelling, so imports and hand-edited slots preserve it too.

## What it breaks

- **Neo4j labels split.** `GraphDatabasePlugins/Neo4j/Persistence/Neo4jSubjectUpdater.php:125` —
  `$newLabels = [ 'Subject', $subject->getSchemaName()->getText() ]`. One Subject gets `:person`,
[...]

PR #1480, opened 2026-09-17
Resolve a Schema name as the whole title of its page

JeroenDeDauw:
Follows-up to https://github.com/ProfessionalWiki/NeoWiki/pull/1468

For https://github.com/ProfessionalWiki/NeoWiki/issues/1201

Targets your branch, because as it stands it closes the issue without fixing it: `{"schema": "Employee#Details"}` still stores a name no Schema has. Two commits. The first is a simplification; the second rebases cleanly without it. Against master your branch adds 525 lines net; with both commits, 483.

## Two derivations disagree

`TitleBasedSchemaReferenceNormalizer` and `CachingSchemaLookup` each derive the Schema page with `newFromText()`, which splits a namespace prefix, an interwiki prefix and a fragment off the name, and the normalizer refuses what it has split off while the lookup does not. Reproduced on your branch:

* `Employee#Details`: the `hasFragment()` guard leaves it as written, the lookup re-parses it and finds `Schema:Employee`, so it is created with no violations and stored as `Employee#Details`. New in your branch.
* `Category:X` with a page of that name: the lookup follows the prefix into the Category namespace, where `WikiPageSchemaJsonLookup` throws a `LogicException`. Also on master.
* `Schema:Schema:Employee` is stored as `Schema:Employee`, since one prefix is stripped per parse, so normalizing twice differs from normalizing once. New in your branch.

## One primitive

A local Schema name is the whole page text in the Schema namespace and is never split, which is what `SchemaReference`'s docblock states and why `ISO:9001` works. The MediaWiki call with that meaning is `makeTitleSafe( NS_SCHEMA, $name )`, which `MappingPageHtmlBuilder` already uses for these names. With it in the normalizer, the lookup and `MappingContentValidator`, the three guards have nothing left to do: every prefix stays in the name, a fragment names the page before the `#`, and a name that titles no page is left as written for the lookup to report as missing. Normalization is idempotent. A `#` goes silently where a `Schema:` prefix does not because no page title can hold a `#`, so the fragment has one reading, while `Schema:` can be part of a title.

On the lookup path, the lookup hands its `Title` to the JSON lookup, so no third derivation exists, and names the Schema inside the deserializer's `try`, so a page titled `Schema:Page`, on which your branch's `new SchemaName( $title->getText() )` threw, resolves to null. `GET /schema/page` still answers 500, from `GetSchemaQuery` constructing the name before the lookup: pre-existing, https://github.com/ProfessionalWiki/NeoWiki/issues/1481.

One behaviour change: `Schema:Employee` now names the page `Schema:Schema:Employee` and reports `schema-not-found` instead of being silently mapped to `Employee`. Nothing we ship spells a reference that way. Two parts of your description stop holding: `Help:Person` is normalized to itself rather than left alone, and the "Not addressed here" note is closed, since `Help:Person` is now its own canonical Mapping key.

## The commits

1. **Carry no more than the Schema parser needs**: no behaviour change. The normalizer's memo saved a measured 2.5 µs per repeated Subject read, about 20 µs on an eight-Subject page against [ADR 029](https://github.com/ProfessionalWiki/NeoWiki/blob/master/docs/adr/029-scalability-targets.md)'s 100 ms save budget; the private accessor in `NeoWikiExtension` existed only to hold it, so the parser is built per call like `getSubjectIdParser()`. The docblocks of the parser, the normalizer and its port lose what narrated the change rather than the code. `TestSources::newPropertyTypeRegistry()` replaces 26 identical call sites, which is most of the file count here; `FixedSchemaReferenceNormalizer` becomes an anonymous class inside the one helper that built it; the unused imports go; the renaming map moves out of `CreateSubjectActionTest`'s shared helper; `SubjectContentDataDeserializerTest::testSchemaNameIsNormalizedWhenRead` goes, dominated by `testSlotWrittenWithAnotherSpellingIsReadUnderTheSchemasName`, which runs the real normalizer.
2. **Resolve a Schema name as the whole title of its page**: everything above, the `MappingContentValidator` change included, with a test per reproduced input, plus one test each for `LayoutPersistenceDeserializer` and `ValidateSubjectQuery`, where bypassing the parser had passed the whole suite.

## Verified

Every new test seen failing first; full local suite green; phpcs and phpstan clean; CI green. The table is manual, both branches on one wiki through the REST API:

| | your branch | this branch |
|---|---|---|
| `{"schema": "Employee#Details"}` | 201, stored `Employee#Details`, no violations | 201, stored `Employee`, no violations |
| `{"schema": "Schema:Employee"}` | 201, stored `Employee` | 201, stored `Schema:Employee`, `schema-not-found` |
| `{"schema": "Category:<existing page>"}` | **500** | 201, stored as written, `schema-not-found` |
| RDF read of such a Subject, the page created after it | **500** | 200 |
| `{"schema": "employee"}` | 201, stored `Employee` | unchanged |

## Considered, omitted

* Keeping `newFromText` and teaching the lookup the normalizer's three guards: two derivations to keep in step, and the guards are what made normalization non-idempotent.
* A strict lookup that resolves only the canonical spelling: it would turn `GET /schema/person` into a 404, which nobody has asked for.
* A test for the Lua `getSchema` accessor: the lookup names every Schema after its page, so nothing Lua returns can tell parsed from unparsed.
* Reading the `schema` filter of `GET /subject-labels` through the parser: `schema=employee` finds nothing on your branch, but the relation picker sends the `targetSchema` the API hands it, which is already the page's name, so nothing we ship is affected. Same for `GetSubjectEditNoticesApi`, whose `schema` is context handed to notice providers.
* `CreateSubjectParserFunction` and `Special:CreateSubject` still parse with `newFromText`; they store nothing, so they are not tracked here.

> <sub>AI-authored — Claude Code, `Fable 5.1 (max)`; two-line ask from @JeroenDeDauw after a session review, no redirections; diff not yet human-reviewed; every new test seen failing first, mutations run, full local suite green, phpcs and phpstan clean, CI green, table rows measured on a dev wiki against both branches.</sub>

Code and commit messages by `Opus 5 (1M context)` from a spec written by `Fable 5.1 (max)`; a second `Fable 5.1 (max)` context reviewed the result, and its one added test and two doc fixes are folded into the commits. The description is Fable's.

</details>


JeroenDeDauw:
@alistair3149

PR #1484, opened 2026-09-18
Move select value normalization onto the property type

### LINKED ISSUE #857: Select value normalization (id/label resolution) is hardcoded in the Application layer
> <sub>Filed from analysis @JeroenDeDauw directed during the #822 review. Context: the NeoWiki codebase, #822, #747, and the `SelectValueResolver` / `SelectStatementResolver` code. Written by Claude Code, `Opus 4.7`.</sub>

## Problem

On write, an incoming select value can be an option `id`, a bare `label`, or an `{id, label}` object, and is canonicalized to the id by `SelectValueResolver` + `SelectStatementResolver`. This logic lives in the Application layer and is special-cased across the write Actions (`CreateSubjectAction`, `ReplaceSubjectAction`) and the validate path (`ValidateSubjectQuery`).

- **No plugin parity.** A `PropertyType` owns its schema-parsing (`buildPropertyDefinitionFromJson`) and validation (`validate`), but *not* input normalization — so a custom property type can't accept/normalize lenient input the way `select` does, and `select` is wired into shared services by name.
- **Duplication.** The resolve-then-build sequence recurs at each call site, split into a throwing `resolve()` and a non-throwing `resolveOrLeave()`.

## Desired outcome

Lenient-input normalization shouldn't be select-specific code in shared Application services. Whatever normalization the core path does, a property type should own it — consistent with how it already owns schema-parsing and validation.

## Initial thoughts — suggestions, not requirements (pick or deviate when this is picked up)

- **Relocate onto the plugin:** a `PropertyType` normalization capability (e.g. `normalizeInputValue(raw, definition)`, no-op for most types — or a per-type resolver registered alongside the type to avoid no-op interface methods), with a generic walker dispatching per statement. Smallest behavior-preserving change.
- **Fold into value construction:** make "raw input → `NeoValue`" a type-owned concern, absorbing both `StatementListBuilder`'s `ValueType` switch and the select resolver. More cohesive, larger change.
- **Remove from the core write path:** make the core API canonical-ids-only and move label / `{id, label}` friendliness to an import/reconciliation layer. Smallest core surface; mostly deletes rather than relocates.

## Couplings to weigh (shouldn't block filing, but they shape the choice)

- Orthogonal to the option-**id scheme** (#747, nanoids vs `id=label`) — resolution is the same either way.
- Coupled to the open **"keep lenient label-input on the core write API?"** question (also #747): if the answer is "core is strict," the third option is the natural one and there is little to relocate.
- Adjacent to **#631** (enforcement): running the validator on writes is expected to collapse the `resolve()` / `resolveOrLeave()` split — coordinate with @alistair3149 so the write Actions are not reworked twice.

alistair3149:
Fixes https://github.com/ProfessionalWiki/NeoWiki/issues/857

Select's id/label resolution moves from the Application layer, where it was hardcoded to `select`,
onto `SelectType`, behind an optional interface any Property Type can implement. A value a type
cannot take is reported as a `Violation`, the object `validate()` already returns, and the write
endpoints answer it as a 400 carrying that Violation.

## What changes for API callers

A `select` value may still be sent as an option id, its label, or an `{ "id", "label" }` object.

| Request | Before | After |
| --- | --- | --- |
| Write with a label or id that names no option | 400, fixed English `message` | 400; `message` localized, plus `violation` and `messageTranslations` |
| Write with an object whose `id` or `label` is not a string | 500 (TypeError) | 400 with `violation` |
| Dry-run validate with such an object | 500 (TypeError) | 400 (plain malformed input) |
| Dry-run validate with an unknown label | 200, `invalid-option` warning | unchanged |

The 400 body keeps the `status` and `message` keys it had; `violation` is serialized as the validate
endpoints serialize theirs, so a client parses one vocabulary. Documented under "Refused values" in
`validation-codes.md`.

## What changes in the code

| Before | After |
| --- | --- |
| `SelectValueResolver` and `SelectStatementResolver`, select-only, special-cased at six call sites | `SelectType::normalizeRawValue()`, via the new `NormalizesRawValue` interface |
| Twin loops `resolve()` / `resolveOrLeave()`, differing only in whether they threw | One walk in `StatementNormalizer`; `normalizeOrThrow()` for writes, `normalize()` for dry-run validation |
| Throwing a fixed-string exception | `RejectedValueException` carrying a `Violation`; extends `InvalidArgumentException`, so every existing 400 catch still works |
| Each endpoint building its own 400 | `RejectedValueResponse` trait |

`StatementNormalizer` asks each property's registered type to canonicalize its value and passes
through any type without the interface. It knows nothing about `select`.

## The reported Violation

- `propertyName`: null from the type; the walk attaches it.
- `valuePartIndex`: counted on the value as sent, 0 for a value that is not a list, so the write 400
  and the dry-run 200 index a bad scalar the same way.
- `severity`: fixed at `error`, as ADR 26 classes a value that cannot be interpreted as its type.
- `args`: substituted into the message as plaintext, after wikitext transformation, so templates
  cannot be expanded through the error body.
- `normalizeRawValue()` never throws. It returns a `NormalizationResult` with the best-effort value
  (every part that could be canonicalized, canonical) plus the first Violation, so dry-run validation
  can report each bad part by index instead of the whole list showing as invalid.

## Why a Violation

The definition side of the same defect is #1454: a relation definition without `targetSchema` is
refused only by a check hardcoded in the schema editor, where `buildPropertyDefinitionFromJson`
should own the rule. Each side reports in the currency its consumers already parse: values as a
`Violation`, which the input components and validate endpoints render today; definitions, in the
follow-up, in the JSON-pointer envelope Schema saves already use. Neither needs a new error type.

## Preserved

Writer's-schema drift: an entry whose `propertyType` is not the type the Schema currently gives the
property is left alone, and `SubjectValidator` reports it as `type-mismatch`.

<sub>AI-authored — Claude Code, `Fable 5.1`; direction, design choices (optional capability interface, Violation as the value-side error currency) and two local reviews by the human; verified: `make cs` (phpcs + phpstan) green, and SelectTypeNormalizeTest, StatementNormalizerTest, ViolationMessageTest, RejectedValueExceptionTest, ViolationTest, SelectTypeTest, SetStatementApiTest, ReplaceSubjectApiTest, CreateSubjectApiTest, CreateSubjectPageApiTest and ValidateSubjectApiTest green after rebasing onto master; browser-checked on the demo wiki: editing a Subject whose select option was removed from the Schema meanwhile gets the 400 with the `violation` body (the editor shows the dry-run warning inline and a generic failure toast, as before), picking a valid option saves, and the Company creator saves a select value; the frontend itself is unchanged.</sub>


alistair3149:
Also did a browser test on the schema creation and subject editing flow, and noticed no regression

JeroenDeDauw:
Review from Fable. Little steering from me. Little brainpower put into either your PR or this review.

---

The relocation half does what https://github.com/ProfessionalWiki/NeoWiki/issues/857 asked; keep it.

Rework the write refusal: drop the fixed-`error` 400 and let validation handle unknown options. [ADR 26](https://github.com/ProfessionalWiki/NeoWiki/blob/master/docs/adr/026-validation-severity-levels.md#decision) makes `invalid-option` an authorable severity (default warning) and says severity never blocks a write on its own, only enforcement does. The editor also cannot read the new 400 body (`violationParsing.ts` wants `violations[]`), so a user gets the dry-run's inline warning and then a generic failure on save. The deleted `SelectStatementResolver` docblock predicted this collapse for when the validator runs on the write paths, which it now does in all four Actions.

Change:

- One non-throwing `normalize()` at all six call sites. An unknown option is left as sent, and `SubjectValidator` reports it as `invalid-option` at the author's severity. Under enforcement that is the existing 422 from `presentValidationFailed()`, which `SubjectEditorDialog` already renders inline.
- A malformed shape (a number, `{"id": 42}`, a mismatched `{id, label}`) throws `InvalidArgumentException`: the plain 400 `StatementListBuilder` gives every other type.
- Delete `RejectedValueException`, the `RejectedValueResponse` trait, `ViolationMessage`, `NormalizationResult`, `Violation::withValuePartIndex`, the four new codes, the five i18n keys and the "Refused values" doc section. No frontend change.

The trade: a typo'd label in an import is stored with a warning instead of refused at the door; a schema author can tighten that with `"options": { "severity": "error" }`.

If you keep the refusal envelope instead, at minimum let `SelectType` stamp `invalid-option` with `$property->severityOf( 'options' )` and make `StatementNormalizer::normalizeOrThrow()` throw only when the violation `isBlocking()`.

<details><summary>Smaller findings</summary>

Fix in this PR either way:

- **`SelectOption::normalizeLabel()`** uses `strtolower`, ASCII-only since PHP 8.2, so `öl auf leinwand` does not reach an option labelled `Öl auf Leinwand` although the new docblock promises case-insensitivity. `mb_strtolower`.
- **Naming.** `normalize` now names two things: input canonicalization (`NormalizesRawValue`, `StatementNormalizer`) and label folding (`SelectOption::normalizeLabel`, `normalizedLabel`); they meet in `SelectType::normalizeObject`. The docblocks already say "canonicalize"; the identifiers could too.
- **Docs, `{}`.** A top-level `{}` decodes to `[]`, takes the list branch and clears the Statement with a 200, while `validation-codes.md` says an object without `id` or `label` is refused. State that `{}` and `[]` both clear the value.
- **`ValidateSubjectUpdateQuery`** was rewired but `ValidateSubjectUpdateApiTest` sends no select value, so removing the call there keeps the suite green. One test mirroring `ValidateSubjectApiTest::testValidSelectLabelIsResolvedAndValidates` covers it.

Optional:

- **Duplicate ids.** `["o1demo1aaaaaaa1", "Active"]` stores the same id twice, and `SelectType::validate()` never checks `uniqueItems` (pre-existing). Deduplicating in `normalizeList` is cheap.
- **`valuePartIndex`** counts the value as sent on writes but the stored parts in the validate response (`StringValue` drops blanks), so `["", "Bogus"]` reports index 1 and index 0 for the same input, and the frontend matches on the exact index. The single-`normalize()` path removes this.

Only if the refusal envelope stays:

- The PR table's row "object whose `id` or `label` is not a string: 500 becomes 400" has no endpoint test; a sibling of `SetStatementApiTest::testUnresolvableSelectValueReturns400LocatingTheValue` sending `['id' => 42]` pins it.
- `ViolationMessage`'s plaintext guard on the property name is unpinned: swapping it to `params()` survives every test. A violation named `{{uc:hello}}` in `ViolationMessageTest` catches it.
- `ViolationMessage::valueOf()` and `RejectedValueException::describe()` spread `Violation::$args`; a third-party type returning string-keyed or non-scalar args throws inside the handler's catch block, turning the documented 400 into a 500. `array_values` plus scalar coercion at the boundary, or a tighter contract.
- The four `neowiki-field-select-*` keys are missing from the ResourceLoader `messages` list, unlike every other key of that family; latent until the frontend reads the 400.

</details>

> <sub>AI-authored — Claude Code, `Fable 5.1 (max)`; read-only review run and posted at @JeroenDeDauw's request, his one-line ask, no revisions; comment text not human-reviewed (he read the verdict in-session); claims checked by the full PHPUnit suite and `make cs` on `9368bae4`, 36 mutation runs, and live probes of every row of the PR's table on a dev wiki at this head.</sub>

PR #1485, opened 2026-09-18
Show the subject picker on a bare Special:Subject

JeroenDeDauw:
`Special:Subject` without a Subject ID showed an error box. It now shows the subject picker, with its field focused on load, searching Subjects of every Schema; picking one navigates to `Special:Subject/<id>`. For a subpage that is not a Subject ID the error box stays and the picker sits under it, so a stale link still leads somewhere.

The `schema` parameter of `GET /neowiki/v0/subject-labels` became optional, and the Neo4j lookup's Cypher gained a null guard so an omitted `schema` filters nothing; the limit cap and the per-result read filter are untouched. `SubjectPicker`'s `targetSchema` prop became optional to match: omitting it searches every Schema, while `null` still means a Schema from another Source. Relation fields and the create option are therefore unchanged — creation still needs one Schema to create under, which the bare page names none of.

The single `#ext-neowiki-subject` mount point now carries `data-mw-neowiki-subject-id` only when there is an ID, and the Subject view's configuration variables are set only where a Subject is shown, since reading them costs a permission check per Mapping page. The lookup, which has no visible label, takes its accessible name from `neowiki-special-subject-picker-label`.

The picker's placeholder now reads "Search by label or subject ID" rather than "Search by name or subject ID", since Subjects have a label, not a name.

https://github.com/ProfessionalWiki/NeoWiki/pull/1471 touches the same picker, label search and endpoint; whichever lands second will conflict textually in the Neo4j lookup and the REST doc row.

Left for later: with every Schema in scope, two Subjects sharing a label but not a Schema look identical in the menu. Showing the Schema per row would settle that; the page titles https://github.com/ProfessionalWiki/NeoWiki/pull/1471 adds do not.

> <sub>AI-authored — Claude Code, `Opus 5 (max)`; detailed spec from @JeroenDeDauw relayed by an orchestrating session, three corrections from him mid-flight (focus on load, placeholder wording, the picker under the error box); diff not yet human-reviewed; PHPUnit, vitest, phpcs and phpstan run locally with every new test mutation-checked, picker, relation scoping and focus verified in a browser against a dev wiki, CI green.</sub>

Implementation by `Opus 5 (max)` in a subagent; design, review triage and final edit list by `Fable 5.1 (max)` in the orchestrating session.

</details>


JeroenDeDauw:
LGTM

image
image

alistair3149:
LGTM, review PR at #1510
Just want to note that we still need to disambiguate the subjects when they share the same label (#1471).

JeroenDeDauw:
Yeah, disambiguate is still TODO, and more important here, since it shows all Subjects rather than only those with a specific Schema.

PR #1489, opened 2026-09-18
Show which Subject a search hit matched and lead to its view

JeroenDeDauw:
Follows-up to https://github.com/ProfessionalWiki/NeoWiki/pull/1458

A page found through one of its Subjects was listed as just the page: no matching text, and a link to the page
even when the page does not show the matched Subject, or exists only to hold it. On a wiki used like Wikibase
every hit led to a page titled by an id.

Special:Search now builds those rows from the page's Subjects at render time, reading each row's page once. The
index, the search API and REST are unchanged.

**Where a row leads.** To the page, except when the page's main slot is empty (entity-first creation leaves it
empty): then the row leads to `Special:Subject/<mainId>`, and so does the Go button. That rule needs no
attribution, so a row the matcher cannot attribute is still retargeted; a page without a Main Subject keeps its
row. Where a hit lands beyond that will be governed by the wiki-level mode setting `$wgNeoWikiSubjectFirst`,
built separately.

**What a row shows.** A row leading to a Subject is titled with the Subject's display name, never an id, and drops
the page-size line. The extract names the matched Subject's Schema, its name when the title does not carry it
already, and up to three matching values as `Property: value`, with the matched words highlighted as in page
extracts. A reader who may not read the page keeps today's plain row. File rows are left alone.

**Attributing the match.** MediaWiki's database engines hand each hit the regexes they searched with, so a Subject
line counts as matched on the same terms the page was found by. CirrusSearch hands over nothing, so the query's
words of three or more characters are matched as word prefixes against the Subject's values; a miss only costs
the row its Subject block. When several Subjects match, the Main Subject wins, then page order, one per row.

Considered, omitted:

* A "+N more" affordance when several Subjects of a page match.
* Resolving a typed label to a Subject for the Go button; that needs a label lookup that exists only with a graph
  store.
* Retargeting core's "There is a page named X" notice when X is an id-titled page.
* Guarding the attribution regex against PostgreSQL's unescaped search terms. An invalid pattern leaves the row
  plain and logs a PHP warning, as core's own highlighter does.
* Styling for the new row markup.

> <sub>AI-authored — Claude Code, `Fable 5.1 (max)`; spec from an in-session design pass with @JeroenDeDauw, who redirected the design twice before a one-word go and once more after review, when a wiki-level mode setting took over the landing question; diff not yet human-reviewed; tests written first and confirmed failing without the code; a second mutation pass after review, 71 mutations, caught 56 and with the tests it added 66; search test classes plus phpcs and phpstan green locally, CI green on MW 1.43 through master.</sub>

Design, review and this description by `Fable 5.1 (max)`; implementation and its own review passes by an `Opus 5 (max)` subagent; the pr-review, simplify and review-tests passes ran on Fable after Opus hit its usage cap. Changed after the implementer's mutation run: the attribution-based landing and the "on page" line removed for the coming mode setting, the page-size line dropped on Subject rows, the short-word rule, the lookup moved to Persistence, `MatchedSubject` folded into the builder, the tests the two later passes added, and a final pass that rewrote or dropped tests coupled to internal shapes (split arrays, cache identity, log context, markup structure) so the suite asserts what a reader sees.

</details>


JeroenDeDauw:
Tested locally. LGTM

JeroenDeDauw:
Merging to get this into today's user testing session. Review/testing still welcome

PR #1493, opened 2026-09-18
Explain what a Mapping page is when editing one

JeroenDeDauw:
In an ECHOLOT user-testing session a tester opened a Mapping page expecting a mapping builder, found only the raw JSON editor and no documentation. Editing a page in the `Mapping` namespace now carries this notice above the editor:

> A Mapping page defines an ontology projection of your Subjects. There is no UI for editing Mappings yet. You define them via the [Mapping JSON format](https://neowiki.ai/docs/authoring/mapping-format).

It reuses the `AlternateEdit` hook that already frames the on-wiki configuration page's JSON editor, whose framing is unchanged. The notice also shows on a Mapping page that does not exist yet, and on preview.

Considered, omitted:

* Core's `TitleGetEditNotices` seam, which renders in the intro block rather than immediately above the editor, and would conflate this notice with the admin-authored `MediaWiki:Neowiki-editnotice-*` messages.
* The Mapping page's `view` action, which already renders an overview of the mapped Schemas and prefixes.
* The `Schema` and `Layout` namespaces, which are edited through their own UI.

> <sub>AI-authored — Claude Code, `Opus 5 (max)`; detailed spec from @JeroenDeDauw relayed by an orchestrating session, no redirections, one word changed in the supplied notice text ("defining" to "editing", since Special:Mappings does have a create UI); notice seen in a browser by @JeroenDeDauw, diff not yet human-reviewed; written test-first and mutation-checked, reviewed by two independent agents and revised, full PHPUnit and phpcs/phpstan green locally, notice checked live on an existing, a not-yet-existing and a non-Mapping page.</sub>


JeroenDeDauw:
image

JeroenDeDauw:
LGTM

JeroenDeDauw:
Looking at getting rid of this junk for users who cannot edit. Will push to this PR

image

JeroenDeDauw:
Ah nvm we already take care of removing such junk via another extension. Doesn't make sense to put that into NeoWiki.

Issue #630, opened 2026-02-23
[EPIC] Review and flesh out Relations design


JeroenDeDauw:
_Written by Claude Code, Opus 4.6. Result of reading all NeoWiki docs, ADRs, context files, and PHP/TS domain models, prompted by @JeroenDeDauw._

_Updated 2026-07-05: added the Status table and per-section updates pointing to since-created issues and ADRs; corrected the cross-wiki use case. Claude Code, `Fable 5 (max)`, reviewed by @JeroenDeDauw._

_Updated 2026-07-21: refreshed the Status table after the relation integrity pass merged (#1080, #1082, #1083, #1084) and the Relations model was proposed as Draft [ADR 28](https://neowiki.ai/docs/adr/028-relations-model) with [planning/Relations.md](https://github.com/ProfessionalWiki/NeoWiki/pull/1125). Claude Code, `Fable 5 (max)`, not yet reviewed by a human._

_Updated 2026-08-26: extended row 3 and section 3 after the relation-target-editor design discussion identified inverse relations on the editing side as its own open area — recorded in https://github.com/ProfessionalWiki/NeoWiki/issues/904#issuecomment-5429070444 and proposed for planning/Relations.md in https://github.com/ProfessionalWiki/NeoWiki/pull/1322. Claude Code, `Fable 5 (max)`, not yet reviewed by a human._

See also: [issues labeled "relation"](https://github.com/ProfessionalWiki/NeoWiki/issues?q=state%3Aopen%20label%3A%22relation%22)

---

## Status (2026-07-21)

Each item here should end up as a decision (ADR where architectural), a dedicated issue, or documentation; once every row points elsewhere, this issue can close. **As of 2026-07-21 every row points elsewhere** — closing is a maintainer call; rows 1, 2, and 6 finalize when [ADR 28](https://neowiki.ai/docs/adr/028-relations-model) is ratified.

| # | Item | Type | Status |
|---|---|---|---|
| 1 | Relation properties need schema-level definitions | Decided (ADR 28, Draft) | Won't build: qualification uses typed Subjects; the edge-`properties` bag is removed. Removal tracked in #1119, gated on ADR 28 ratification. |
| 2 | Target Schema is mandatory and singular | Decided (ADR 28, Draft) | `targetSchema` widens to a list of Schemas, tracked in #991; the supertype variant was rejected. The unconstrained "any Subject" case is parked in planning/Relations.md until a concrete need arrives. |
[...]

PR #1494, opened 2026-09-18
Hide the relation type from the schema editor

JeroenDeDauw:
In two ECHOLOT user-testing sessions the tester stopped at the "Relation type" field, one of them asking why the property name appeared there a second time. The field names the edge label of NeoWiki's own projections — the Neo4j relationship type and the native RDF predicate — which is not a choice a wiki editor has reason to make.

The relations redesign (https://github.com/ProfessionalWiki/NeoWiki/issues/630, draft ADR in `docs/adr/028-relations-model.md`) merges the relation type into the property name. This takes only the user-facing step, ahead of it.

## What changed

* The schema property editor drops the relation-type input, its validation and its two messages.
* `RelationType.createPropertyDefinitionFromJson()` supplies the property name when the JSON has no relation type — the point where a property becomes a relation property, so a new one gets what the field used to default to. A Schema read from the wiki always carries one, and it passes through untouched on both read and save.
* ADR 028 records that the UI half is shipped; the decision is still open.

## Deliberately left

* Everything behind the editor: the stored value, the JSON format, the REST API, the PHP backend, the Neo4j projection, the RDF export and the demo schemas. A relation type differing from the property name still works; setting one now means editing the Schema page's JSON.
* Renaming a property leaves its relation type alone. Switching a property's type away from Relation and back, or to Relation before naming it (leaving `New Property N`), still ends at the property name — unchanged behaviour, now off screen.
* No other frontend surface reads a property's `relation` field.

> <sub>AI-authored — Claude Code, `Opus 5 (1M context, max)`; detailed spec from @JeroenDeDauw, no redirections on the change itself; diff not yet human-reviewed; TDD with mutation checks on the one added line, full Vitest suite (2322 tests), build and lint green locally, CI green; verified in the browser on a dev wiki — a property switched to Relation shows no relation-type field and saves with the relation type set to its name, and editing a stored relation property leaves its relation type alone.</sub>

Implemented by an `Opus 5` subagent from a written design, then reviewed by a second `Opus 5` subagent with no
implementation context. Two of its findings are applied here: a shared Vitest helper change reverted in favour of
`setActivePinia` in the single spec that needs it, and a unit test pinning the new fallback's empty-string branch,
which no other test caught. The motivation text was corrected during implementation: the relation type is not
Neo4j-only, it is also the native RDF predicate.

</details>


JeroenDeDauw:
LGTM

verified in browser

image

Issue #1363, opened 2026-09-07
Create Specialization of Main Menu or Tool bar for ECHOLOT / Structured Data Type Instances?


Habennin:
NeoWiki provides a rich extension to MediaWiki enabling complex but versatile data modelling for analytic research data associated to wiki pages.

That said, the main menu and tool bars do not obviously provide the quick set of tools you might want as a hardcore user of these new features. Main features of the product are hidden in text on the 'main page'.

Suggestion:

Isolate and promote main tools and schemas as readily accessible and usable from sidebar tools, possibily based on permissions (e.g.: admin has access to a schemas editing tool set that hoi polloi user does not).

Question:

I am not sure of all the options for setting such a thing up but my small research suggests that the main menu can be controlled by any user with permissions, while the tool menu is controlled via php config settings etc.

Where would it be most sensible to setup such system level settings?

Representative Context: 

Takin wants to enter a NeoWiki instance and have a suite of tools for data modelling / mapping etc. listed as first citizens, obvious things in tool boxes, rather than as links in a long descriptive text.

End Users: want to see a list of the schemas that are in their instance and how many thing that are in them and a quick link to search in them or make a new one of that thing.

Habennin:
Also with NeoWiki the things entered in the schemas, I believe are the most important things in the system. The Wiki pages are kind of a leftover. But what would be interesting perhaps is if there were special pages automatically generated that were something akin to aggregator pages for instances of the schemas.

So let's say I have a schema person, there could be automatically generated a special page for the person schema which would be a main place of interacting with the NeoWiki instance. I would go here and I would know the population of the person model (1 record or 1 million) latest entries, answers to popular questions, graphs showing health of data etc. etc. And there would be a 'create new' button at the top of it so I could make another instance.

I'm not sure if this is totally related, but there is this question of the relation of the page to one or more subject instances. Well I would think normally it would be 1. Like I have a person record "George Bruseker" and I have a page for "George Bruseker" and they should necessarily be 1:1. The semantic relation between them is that the Wiki Page is about the instance. But for these schema special pages, I guess it could list all the instances of itself so 1:n relation and the semantic relation is actually instance of.

alistair3149:
> I am not sure of all the options for setting such a thing up but my small research suggests that the main menu can be controlled by any user with permissions, while the tool menu is controlled via php config settings etc. Where would it be most sensible to setup such system level settings?

That is right.

[...]

Issue #748, opened 2026-04-20
SidebarBeforeOutput handler does not check for wikipage context


osnard:
Looks like the SidebarBeforeOutput hookhandler in Extension:NeoWiki breaks, when accessing read access protected page:

```
MediaWiki internal error.

Original exception: [31aae8a9b2ef7f09e613632f] /wiki MediaWiki\Page\PageAssertionException: The given PageIdentity Special:Badtitle does not represent a proper page
Backtrace:
from /app/bluespice/w/includes/page/WikiPageFactory.php(49)
#0 /app/bluespice/w/extensions/NeoWiki/src/Persistence/MediaWiki/Subject/SubjectContentRepository.php(35): MediaWiki\Page\WikiPageFactory->newFromTitle()
#1 /app/bluespice/w/extensions/NeoWiki/src/Presentation/ViewHtmlBuilder.php(45): ProfessionalWiki\NeoWiki\Persistence\MediaWiki\Subject\SubjectContentRepository->getSubjectContentByPageTitle()
#2 /app/bluespice/w/extensions/NeoWiki/src/EntryPoints/NeoWikiHooks.php(309): ProfessionalWiki\NeoWiki\Presentation\ViewHtmlBuilder->pageHasSubjects()
#3 /app/bluespice/w/includes/HookContainer/HookContainer.php(159): ProfessionalWiki\NeoWiki\EntryPoints\NeoWikiHooks::onSidebarBeforeOutput()
#4 /app/bluespice/w/includes/HookContainer/HookRunner.php(3528): MediaWiki\HookContainer\HookContainer->run()
#5 /app/bluespice/w/includes/skins/Skin.php(1615): MediaWiki\HookContainer\HookRunner->onSidebarBeforeOutput()
#6 /app/bluespice/w/includes/skins/SkinTemplate.php(608): Skin->buildSidebar()
#7 /app/bluespice/w/includes/skins/SkinTemplate.php(593): SkinTemplate->getPortletsTemplateData()
#8 /app/bluespice/w/includes/skins/Skin.php(834): SkinTemplate->getCategoryLinks()
#9 /app/bluespice/w/includes/skins/SkinTemplate.php(216): Skin->getCategories()
#10 /app/bluespice/w/includes/skins/SkinTemplate.php(156): SkinTemplate->prepareQuickTemplate()
#11 /app/bluespice/w/includes/skins/SkinTemplate.php(180): SkinTemplate->generateHTML()
[...]

PR #1495, opened 2026-09-18
Show the NeoWiki sidebar section on special pages

JeroenDeDauw:
The NeoWiki sidebar section was absent on every special page, Special:Subject, Special:Schemas and
Special:NeoWiki included: onSidebarBeforeOutput returned early for a title that cannot exist, and
Title::canExist() is false for all of NS_SPECIAL. The user guide tells readers to start from Overview in
that section, then sends them to Special:Schemas, where it was not there.

The section is now built whenever the skin has a title. On a title that cannot exist it holds the links
that describe no page: Overview first, Create subject last, still gated on createpage and edit. Manage
subjects, View RDF, Edit JSON and the inline Create-subject page tool all describe a page, so they stay
out. Ordinary pages are unchanged in every namespace.

The early return came from https://github.com/ProfessionalWiki/NeoWiki/issues/748, where the handler
fataled on Special:Badtitle by asking WikiPageFactory for a page that cannot exist. That call is long
gone, so the guard now wraps the page-dependent tools only: they describe a page, and building them
without one would also put a Subject permission lookup on every special-page view.

Related: https://github.com/ProfessionalWiki/NeoWiki/issues/1363

> <sub>AI-authored — Claude Code, `Opus 5 (max)`; one-line ask from @JeroenDeDauw, implemented by a subagent to a spec from this session and revised after a review pass; diff not yet human-reviewed; tests written first and confirmed to fail without the change, full PHPUnit suite and `make cs` green locally, sidebar checked live on special pages, Schema and Layout pages and an ordinary page.</sub>


JeroenDeDauw:
LGTM

tested locally

alistair3149:
LGTM, follow up at #1505

Issue #1491, opened 2026-09-18
Special:Subject UI optimization


JeroenDeDauw:
## Issue 1: page title is not reactive

If I edit the label of the subject and save it, the NeoWiki UI naturally updates, but the title of the special page remains what it was. Observed during user testing session, where the user noticed how the title did not change and wondered why their change did not work (they did not see the updated label in the neowiki UI itself)

<img width="1245" height="724" alt="Image" src="https://github.com/user-attachments/assets/96bd7733-8746-4463-ba95-fcbe5b85de66" />

## Issue 2: special page UI

The very top of the page shows "Special Page". That is highly random and opque to non-MW users.

If we need to have a tab, then "Subject" would be clearer.

Perhaps we could get rid of the tab? Maybe also of the title? We're duplicating the label of the subject via the header in the subject's box and the special page title. This could also fix the first issue.

The task here is basically to zoom out from the "neowiki UI dropped in the place MediaWiki provides" to looking at whatever the user sees in the browser wholistically and seeing how we can optimize it. What would a design-from-scratch (using the NeoWiki design system, but not starting from any MediaWiki UI premises) look like for the entire page (everything the user sees in the browser when viewing special:subject/id)? Maybe we cannot realize all of that, but it seems a usefull thought experiment and plausably great north star.

JeroenDeDauw:
@lozanaross you seemed to be saying something about issue 2 today. In any case, you inspired me to file this github task.

@alistair3149 seems like the kind of thing you like to think about

PR #1499, opened 2026-09-18
Add the subject-first wiki mode

JeroenDeDauw:
NeoWiki serves page-first wikis, where the page comes first and Subjects hold data about it, and subject-first
wikis, where the Subject comes first and the page only stores it. `$wgNeoWikiSubjectFirst`
selects which this wiki is; `false`, the default, is page-first. It can also be set on `MediaWiki:NeoWiki`
as `subjectFirst`, and it replaces `$wgNeoWikiDereferenceSubjectsToHostingPage`, removed without alias.
Recorded as [ADR 33](docs/adr/033-page-first-and-subject-first-wikis.md), a draft open for review.

| Behaviour | Page-first | Subject-first |
|---|---|---|
| Creator: "Store the subject on" | shown; defaults to the current page; a new page needs a title | hidden; always a new page |
| Target created inside the editor | non-main Subject on the host page | its own page |
| Landing after save | the page | Special:Subject |
| Links to Subjects in infoboxes and views | the page | Special:Subject |
| Links to Subjects on the Data tab | relation values lead to the target page's Data tab with the row highlighted; the row title is not a link; "Open" leads to Special:Subject | Special:Subject |
| Concept URI in a browser | the page | Special:Subject |
| "Move" on the Data tab | offered | not offered |
| Deleting a Subject that is its page's only one | the Subject goes, the page stays | the page goes with it |

The mode changes defaults, landing surfaces and link targets only; nothing in the REST API, the parser
functions or the handling of pages holding several Subjects depends on it. RDF output is unchanged: only
the browser redirect of a concept URI follows the mode.

On a page-first wiki, which every existing wiki is, three things change: a browser dereferencing a concept
URI lands on the page rather than on Special:Subject; the creator lands on the page also when opened
without one, as on Special:CreateSubject; and on the Data tab the row title is plain text, relation values
inside a row lead to the target's row on its page's Data tab, and the inline actions gain "Open subject",
since the overflow menu that carried it is hidden on desktop.

Two additions the subject-first mode needs, neither conditional on it:

* `POST /neowiki/v0/subjects` accepts an optional `id`, as `POST /neowiki/v0/page/{pageId}/subjects`
  does (`400` malformed, `409` in use, minted when omitted): a Subject drafted inside the editor already
  carries the id the relations pointing at it name. The endpoint's existing `409` for a taken title
  stays; only that one carries `pageTitle`, which is how the client tells the two apart.
* Where nobody is asked which page a Subject goes on, a label that names an existing page does not title
  the new one: the Subject's id does, so namesakes can be created. Where the page is asked about, the
  page-first creator included, a taken title still answers `409` at the question.

Every row is tested in both modes: PHPUnit for the setting, the concept URI and the `id` parameter; vitest
for the creator, the editor, the store, the rows and relation links. To try it: `$wgNeoWikiSubjectFirst = true;`
in `LocalSettings.php`, or `{ "subjectFirst": true }` on `MediaWiki:NeoWiki`.

## Open question

What titles a Subject's page in a subject-first wiki. Here the label does, and the Subject's id only where the
label cannot or the title is taken, so page renames stay allowed and a label edit leaves the title behind. The
alternative is the id as the title, as Wikibase does: uniform titles, no namesake handling, renames and moves
refused as an invariant, at the cost of every MediaWiki surface that prints a title showing the id unless link
rendering and display titles are hooked. The ADR carries the same question.

## Follow-ups

Independent of this PR; whichever lands second adapts.

* https://github.com/ProfessionalWiki/NeoWiki/pull/1489: where a search hit and Go land must follow the mode.
* https://github.com/ProfessionalWiki/NeoWiki/pull/1471: the picker's disambiguation line must follow the mode.

## Considered, omitted

* In page-first mode, an infobox link to a target that is not its page's Main Subject leading to that
  page's Data tab row rather than to the page: the frontend has no per-Subject Main Subject flag.

> <sub>AI-authored — Claude Code, `Fable 5.1 (max)`, implementation by `Opus 5 (max)`; detailed spec from @JeroenDeDauw, redirected twice (ADR to draft and Move hidden; plain-words docs and a simplify pass); diff not yet human-reviewed; tests written first and verified to fail without the production code, `make cs` and `make phpunit` green locally, TS checks run individually since `make tsci` OOM-kills in the container, both modes exercised in a browser, CI green.</sub>

Design, review and the docs by `Fable 5.1 (max)`; implementation, a pr-review pass and a text-review pass by `Opus 5 (max)` subagents. Two blocking findings of that review were fixed before publishing: namesakes in subject-first mode, and the missing desktop route from a page-first Data tab to Special:Subject.

</details>


JeroenDeDauw:
Conceptually LGTM (ofc, since it mostly came from me)

LibrErli:
Just for clarification:
* Your above table shows in the first line, the for a creator of a new subject the "create on" possibility for new subjects on subject-first wiki is hidden because it's "always a new page" It is always a new page? 
* Wouldn't it be a possiblity for subject-first wiki, that subjects are always created within a "Subject:"-Namespace? imho as a dedicated namespace, which would allow both: clearly separation/identification of subjects and usage of the full MediaWiki-API stack to deal with subjects. 

My thoughts on your open question:
The "page"-title on subject-first wiki should be the subject-id, but the UI should show more prominent the label of the subject; both at is realized in Wikibase.

JeroenDeDauw:
On suject-first wikis the subjects always end up on a new wiki page via most creation paths. You'll still be able to use the multi-subject per page capability if you want to via some paths, though we could add an option to disable that and enforce one subject per page. Currently we can only guess what is optimal there, so it's wait and see. Enforcing one subject per page or exposing the multi-subject capability more is something we can easily do later.

Good point on the Subject namespace or similar namespacing. Yes, we could do that, though it's orthogonal to the modes in this PR.

> My thoughts on your open question:
> The "page"-title on subject-first wiki should be the subject-id, but the UI should show more prominent the label of the subject; both at is realized in Wikibase.

Yes, makes sense. Presumably with fallback to just the ID if there is no label set.

Lower priority because on subject-first wikis, standard viewing of subjects is via Special:Subject, which already shows the label, rather than via normal wiki pages. Would do this after https://github.com/ProfessionalWiki/NeoWiki/issues/1491. Tracked by https://github.com/ProfessionalWiki/NeoWiki/issues/1503

JeroenDeDauw:
Merging to get this into today's user testing session. Further review welcome!

JeroenDeDauw:
Actually, re-prioritizing this as urgent, since its an easy-to-fix issue that the user testing sessions still run into:
* https://github.com/ProfessionalWiki/NeoWiki/issues/1503

PR #1500, opened 2026-09-19
Add a Monolingual text Property Type

### LINKED ISSUE #1497: Add Monolingual text Property type
A text value with a BCP 47 language tag, so one property can hold the same text in several languages and each part keeps its language: a film's original title as `"Zinema"@eu` and `"Cine"@es`, with the [RDF export](https://github.com/ProfessionalWiki/NeoWiki/blob/master/docs/api/rdf-export.md) emitting exactly that. Wikibase calls this [monolingual text](https://www.wikidata.org/wiki/Help:Data_type), Semantic MediaWiki [Monolingual text](https://www.semantic-mediawiki.org/wiki/Help:Type_Monolingual_text) (written `Text@en`); users coming from either expect it.

## Who asked

| When | Who | What |
| --- | --- | --- |
| 2026-09-16 | David Lindemann (EHU), ECHOLOT user-testing session | Wanted to attach a language to a film's original title, found no way ("I cannot associate text with language … so I need to use text") and fell back to plain text. |
| 2026-09-10 | Matej Ďurčo (OEAW), requirement FR8.7 | "question to NeoWiki Core about language support for property values": https://github.com/ECHOLOT-ECCCH/echolot-WP2/issues/61 |
| 2026-05-21 | Lozana Rossenova (TIB), WP3 call | Of the levels multilinguality could live at (UI, ontology, data), start with the data. |
| 2026-04-07 | George Bruseker (takin), consortium discussion | It must be a requirement that the data can be in multiple languages. |
| 2026 | Outside ECHOLOT | Requested as well, for multilingual vocabulary data. |

Requirement FR2.10 (https://github.com/ECHOLOT-ECCCH/echolot-WP2/issues/27) covers multilingual data, but the tasks linked from it today are property labels and select options only. Consortium case-study data already models this: one collection keys its titles by BCP 47 tag across most of its entity types; another carries trilingual descriptions plus an explicit name-language field.

## Where NeoWiki is today

* A `text` value is a list of plain strings, and [every Property Type uses one of four value shapes](https://github.com/ProfessionalWiki/NeoWiki/blob/master/docs/api/subject-format.md#value-formats): strings, a number, a boolean, relation objects. No value carries a language.
* The RDF side is ready: [`Literal`](https://github.com/ProfessionalWiki/NeoWiki/blob/master/src/Domain/Rdf/Literal.php) carries an optional BCP 47 tag, reserved for "future projections", and the Turtle writer already emits `"x"@eu`.
* A Mapping can set [`lang`](https://github.com/ProfessionalWiki/NeoWiki/blob/master/docs/authoring/mapping-format.md#format-version-1) per property: one blanket tag for every value of that property, never a tag per value.
* Property Types are an [extension point](https://github.com/ProfessionalWiki/NeoWiki/blob/master/docs/extending/property-types.md), but a new value *shape* is a core change.

## Open questions

Nothing below is decided; the list is meant to be argued with.

**Model**

* Value shape: a new Value Type whose parts are `{ text, language }` objects (the `relation` shape is already a list of objects), or plain strings with an in-band convention like SMW's `Text@en` (no core change, but the tag hides inside the string and every consumer has to parse it)?
* A tag per part, or one tag per Statement with one Statement per language?
* Is one value per language a constraint (as for Wikibase labels), or may a language repeat (as Wikibase monolingual text statements can)? What does `uniqueItems` compare, the text or the text plus tag?
* Which languages: any BCP 47-shaped tag, MediaWiki's language list, a per-property allowlist? Is an undetermined language (`und`) needed?
* Considered and declined (2026-09-19): modelling the value as a Subject (text plus language, linked by a relation). It has no identity, would need inline nested creation per value, and would export as a node instead of a tagged literal.

**Stores and export**

* Neo4j has no tagged strings and no maps on properties: parallel lists (`title` and `title_lang`), encoded strings, a JSON string, one node per value? Whatever is picked becomes a Cypher query contract. SPARQL stores carry the tag natively.
* Precedence on export: when a value carries its own tag and the Mapping also sets `lang` for the property, does the value win, the Mapping win, or is the combination rejected at save time as `lang` plus `datatype` is today?

**Editing, display, scripting**

* Input: a language picker per part, and what the default language is (wiki content language, user language, last used).
* Display in Views: all languages, the reader's language with fallback, or a Display Attribute per Layout? The frontend has no language concept today.
* What `nw.getValue` and the parser functions return for such a value.
* Constraints: `required` as "at least one language" or "this language is required"? `minLength` and `maxLength` per part, as for `text`?

## Related

* https://github.com/ProfessionalWiki/NeoWiki/issues/1490 asks whether a monolingual text type would make multilingual Subject labels unnecessary; that thread decides Subject labels, not this issue.
* https://github.com/ProfessionalWiki/NeoWiki/issues/710 multilingual property labels
* https://github.com/ProfessionalWiki/NeoWiki/issues/726 multilingual select options
* The cohesive pass across surfaces: [OntologyMapping.md Q8](https://github.com/ProfessionalWiki/NeoWiki/blob/master/docs/planning/OntologyMapping.md#q8-multilinguality) and https://github.com/ProfessionalWiki/NeoWiki/discussions/996 ("We have not done a cohesive pass over multilingualism", 2026-07-23).

<details><summary>Where it would land in the code</summary>

Near-leaf if built as a new Value Type: a `ValueType` case and a value class shaped like `RelationValue`, a `PropertyType` with its `PropertyDefinition`, an RDF value mapper, a Neo4j value builder, and a Vue input and display, registered as documented in [Property Types](https://github.com/ProfessionalWiki/NeoWiki/blob/master/docs/extending/property-types.md). Only two exhaustive PHP `match` sites break (`StatementListBuilder`, `StatementDeserializer`), plus the TS equivalents in `Value.ts` and `ValueDeserializer.ts`. `StringValue` is best left alone: its `string ...$strings` constructor cannot hold a tag and it has several hundred call sites. The writer's schema is stored per Statement ([ADR 011](https://github.com/ProfessionalWiki/NeoWiki/blob/master/docs/adr/011-include-writers-schema.md)), so a new type needs no migration. Documented contracts that gain a fifth value shape: the Subject JSON format, the REST API and the Lua API.

</details>

> <sub>AI-authored — Claude Code, `Fable 5.1 (max)`; detailed spec from @JeroenDeDauw building on a prior research session, no revisions; text not yet human-reviewed; code claims checked against master (8814161), quotes against the session transcripts and issue threads, links checked.</sub>

JeroenDeDauw:
Fixes https://github.com/ProfessionalWiki/NeoWiki/issues/1497

Adds a `monolingualText` Property Type, backed by a Value Type of its own whose parts are `{ "text", "language" }` objects, so one property holds `"Zinema"@eu` and `"Cine"@es`. `docs/adr/034-monolingual-text-value-type.md` records the shape and the rejected alternatives.

Canonical form is set at construction, as for `StringValue`: the text is trimmed, a part without text is dropped, and the language is lowercased after a BCP 47 well-formedness check, the rule `Literal` already applied to a Mapping's `lang`, now shared.

Validation reuses the existing codes: `required`, `min-length` and `max-length` per part, `unique` comparing text and language together, so the same text in two languages is not a duplicate, and `single-value-only` when a single-valued property holds several parts, a check `text` and `url` do not make. That message now reads "Only one value is allowed." The type is registered in `PropertyTypeRegistry` and listed in the Schema editor's type picker. Statements written while a property was `text` keep their writer's type and are reported as `type-mismatch`; nothing converts them.

RDF emits one `rdf:langString` literal per part. Neo4j stores `Zinema@eu`, split on the last `@`. A Mapping's `lang` and `datatype` do not apply to a value that carries its own tag.

In the editor each part has a text field whose language tag sits as a button at its end. The button opens a search panel listing the reader's languages first, then the rest by name, with what is typed ranked by how well it matches. A well-formed tag that names no known language is offered as an entry of its own, so `und` and ISO 639-3 codes such as `ara` can be entered. A new row defaults to the wiki's content language. A single-valued property whose value holds several parts shows them all rather than dropping the rest on save. A row emptied of its text is dropped when focus leaves it, except the trailing row a new part is typed into, and except while the window itself has lost focus.

A View shows the texts in the first of the reader's languages that any text is in: their interface language, then MediaWiki's fallback chain for it, then the content language. A small "N more languages" link follows, revealing the other texts with their tags and becoming "Fewer languages". When the reader reads none of the languages, the texts in the first text's language are shown. Every rendered text carries `lang` and `dir="auto"`.

The schema editor's bound fields (character length, number range, date bounds) now sit side by side whenever the property pane has room, instead of only above the desktop breakpoint: a change to the shared `NeoNestedField`, so it reaches the text, number, date and datetime forms too.

Related: https://github.com/ProfessionalWiki/NeoWiki/issues/1501, the sibling inputs do not show field-level violations on multi-valued properties; this one does.

Considered, omitted:

- Enforcing `multiple` on `text` and `url`, where it only switches the editor between one row and a list: their own change.
- A shared base for `TextProperty` and `MonolingualTextProperty`, and a shared length-validation helper: both change `text`, so they belong in their own change.
- Loading `mediawiki.language.names` on demand: it becomes a dependency of `ext.neowiki`, so the language table (7 KB gzipped) ships on every page NeoWiki loads on.
- An explicit remove button per row instead of clearing the text: a decision for every multi-part input, so not here.
- Converting `text` Statements when a property changes type, and a language allowlist.
- Requiring a well-formed BCP 47 tag rather than one of the right shape: today `en-a` and `x` can be stored, and Oxigraph rejects the whole update carrying one, leaving that page stale. It changes ADR 34 and the Mapping `lang` rule.
- Keeping `Ger`, on the way to "German", from being offered as the tag `ger`.
- Naming the property and the row in the language button's accessible name.
- Finding a language by its English name on a wiki without CLDR, which `Intl.DisplayNames` could do in the interface language.
- Fixing a side effect of the `NeoNestedField` overrides: their `min-width: 0` shrinks RedHerb's palette inputs to 44px and clips the datetime bounds in a 400px pane.

<sub>AI-authored — Claude Code, `Fable 5.1 (max)` and `Opus 5`; detailed spec from @JeroenDeDauw after an in-session design pass, four rounds of his feedback applied; the language button, the content-language default and the View link designed by @alistair3149 and built in https://github.com/ProfessionalWiki/NeoWiki/pull/1517; diff not yet human-reviewed; PHPUnit, vitest, phpcs, phpstan, eslint and the vite build green locally and on CI, new tests mutation-checked, behaviour verified in a live worktree wiki.</sub>

Design settled in-session by `Fable 5.1 (max)`; implementation and four review-fix rounds by `Opus 5 (max)` subagents; an independent five-lens review pass and a simplify pass on Opus; final review and the original text by Fable. The editor's language button, the content-language default and the View link came from @alistair3149's review of the first version and were built by `Opus 5` in https://github.com/ProfessionalWiki/NeoWiki/pull/1517. Rebased onto master and squashed by an `Opus 5 (max)` subagent, with this description updated to match, after a six-pass review on Fable.

</details>


JeroenDeDauw:
Tested locally and did some iteration. Can't think of how to improve the UI further but imagine there might well be notably better designs

image
image
image
image

JeroenDeDauw:
There is some issue with language selection. When I clear the language and start typing something, "English" appears, and my typing is appended to it.

Also finding specific languages somehow seems buggy.

And trying to select your language by finding it in the scroll doesnt work at all because there are a trillion obscure languages in the dropdown, with it only showing the first so many, so most major languages are not selectable without filtering first. This is highly unusual, basically all selectors on the internet allow for scroll and select, with some even putting specific languages like English at the top for easy selection.

JeroenDeDauw:
The expansion/collpase icon used in the value display component is probably not optimal. And it probably should look different in the collapsed and expanded states, and in the expanded state perhaps not be in the middle of the list.

alistair3149:
I redid the UI, changed the default language to use content language instead of interface language, and added behavior tests in #1517.

PR #1517, opened 2026-09-21
Pick a monolingual text part's language from a button in its field, and show other languages behind a link

alistair3149:
Follows-up to https://github.com/ProfessionalWiki/NeoWiki/pull/1500

Five commits on top of #1500. They apply in order, and each passes the tests on its own. "The reader's languages" below means the interface language, its MediaWiki fallback chain, then the content language.

1. **Tests for behaviour no test pinned**: the Property Type's JSON reader, a new part's default language, the input re-seeding and its trailing row, clearing one row's violations, and the View's fallback order. They pass against #1500's code.
2. **The language as a button in the text field.** The language field goes; each part's tag (`EU ⌄`) sits over the end of its text field, copying the relation input's grid overlay, and opens a search panel: the reader's languages first, the rest by name, matches ranked. `LanguagePicker.vue` is rewritten on Codex's button and menu; there is no new component, and no shared one changes. Pressing the button keeps the focus where it was, since Safari does not focus a pressed button. This fixes the old field's problems: typed text replaced by the held language's name, "German" stored as the tag `german` on wikis without CLDR, languages that could not be scrolled to, and rows overflowing a phone-width dialog. `domain/languageTag.ts` goes; the picker was its only user.
3. **An emptied row survives switching to another browser tab**, so text copied from elsewhere does not land in the next row's language.
4. **New parts default to the wiki's content language** instead of the editor's interface language, so every editor gets the same default.
5. **The View shows the value like a text value**: the reader's text, then a small "3 more languages" link, in place of the "+3" toggle button that sat between the texts. The link reveals the other texts with their tags and becomes "Fewer languages". When the reader reads none of the languages, the View shows all texts in the first text's language, not only the first text, so the link counts only languages still hidden.

Considered, omitted:

- Tags are accepted on their shape, so `en-a` or `x` can be stored, and Oxigraph rejects the whole update carrying one, leaving that page stale. Requiring well-formed BCP 47 changes ADR 34 and the Mapping `lang` rule. No issue exists for it yet.
- #1500's change to NeoNestedField, its `min-width: 0` overrides, shrinks RedHerb's palette inputs to 44px and clips the datetime bounds in a 400px pane.
- Typing `Ger` on the way to "German" still offers `ger` as a tag.
- The language button's accessible name does not say which property or row it belongs to.
- Without CLDR, typing an English language name finds nothing; `Intl.DisplayNames` could search in the interface language.

## Manual Browser Check

On a wiki without CLDR, such as the dev stack:

1. Check out `pr-1500-review-fixes` and run `make ts-build`.
2. Create a Schema with a multi-valued `monolingualText` property and enough other properties that the form scrolls, and a Subject whose value holds texts in Basque, Spanish, English and German.
3. On the page, the infobox shows the English text and "3 more languages". Click it: the other texts appear with their tags, and "Fewer languages" sits below them.
4. Open the editor. Each row ends with its tag, such as `EU ⌄`.
5. Click a tag: the panel lists your languages first. Type `deu`, press ↓ and Enter: the tag becomes `DE` and the focus is on the button.
6. Open a panel and press Escape, then open it again and scroll the dialog: each time the panel closes and the editor stays open.
7. Scroll to a row that was below the fold and open its panel: what you type goes into its search field.
8. Type `German` in a panel: no tag is offered. Type `und`: it is.
9. Empty a row's text, switch to another browser tab and back: the row is still there.
10. Type into the trailing row: its tag is the wiki's content language, also with `?uselang=de`.
11. In a 375px wide window, the editor does not scroll sideways.
12. In Safari, empty a middle row's text, then click its tag: the row stays and the panel opens.

<sub>AI-authored — Claude Code, `Opus 5 (1M context)`; review of #1500 and design direction (tag button, content-language default, View link) from @alistair3149; diff not yet human-reviewed; vitest, eslint, stylelint and vue-tsc green at every commit, no PHP changed; behaviour checked in Chromium on a dev wiki, not in Safari.</sub>


alistair3149:
https://github.com/user-attachments/assets/2e4ae787-e1e9-481d-a213-37479dd5aee5

JeroenDeDauw:
That UI is so much nicer :laughing:

Issue #1503, opened 2026-09-21
Show subject label instead of the ID page title on subject-first wikis


JeroenDeDauw:
Idea by @LibrErli at https://github.com/ProfessionalWiki/NeoWiki/pull/1499#issuecomment-5757914060

On subject-first wikis, pages like `/wiki/sIFFHEKJOIDJOJD` get created. So the page title ends up being the subject id. Like in Wikibase, it makes sense to show the subject label, when present, as page title. Clearly on the normal page view, and quite possibly also in many other locations where the page title is shown, like on special:recentchanges.

PR #1518, opened 2026-09-22
Show a Subject-id-titled page by its Main Subject's label and id

alistair3149:
For https://github.com/ProfessionalWiki/NeoWiki/issues/1503

A page titled by the id of one of its Subjects is now headed by its Main Subject's label with that Subject's id
beside it, muted: "Ada Lovelace (s1zz…)". The browser tab shows the label alone, an old revision is headed by the
label it held, and without a label the id stays. Pages whose title cannot be a Subject id do no extra lookup. This
applies in both wiki modes, since ADR 31 treats an id title as naming nothing in either.

Recent changes and watchlists link to such a page the same way. Each list reads the Subject slots of its id-titled
content pages in one batch, at the revision the wiki publishes, and shows a label only to a reader who may read the
page; otherwise the link keeps the title. A list reads at most 250 pages and skips Subject data over 64 KiB. Page
content keeps the title too, a Recent changes list transcluded into a page included, since the parser cache would
keep a label after it changed.

The heading is set when the page is viewed, not stored as a parse-time display title: the Subject slot's role is
registered with `display: none`, and core takes a page's heading only from displayed slots. As with a display title,
diffs and error pages keep their own heading.

A page whose Subjects do not deserialize now keeps its title and shows no Main Subject, where its view used to fail.

The heading and the list links are separate commits.

Related: https://github.com/ProfessionalWiki/NeoWiki/issues/1333

## Manual Browser Check

1. Make the wiki subject-first: `{ "subjectFirst": true }` on `MediaWiki:NeoWiki`.
2. On Special:CreateSubject, create a Subject labelled `Ada <Lovelace> [draft]`, a label that cannot title a page, so
   the page is titled by the id. The dialog lands on `Special:Subject/<id>`.
3. Open `/wiki/<id>`. The heading reads `Ada <Lovelace> [draft] (<id>)` with the id in grey; the tab title reads
   `Ada <Lovelace> [draft]`.
4. Open Special:RecentChanges. The page's row links by the same label and id; its "diff" and "hist" links are
   unchanged.
5. Open any other page. Its title is unchanged.

## Follow-up

Left in #1503: showing the label wherever else MediaWiki prints the title, such as the history page's heading, All
pages, What links here, search results and edit summaries.

## Considered, omitted

* Updating the heading after an edit on the page, in the Subject editor or VisualEditor; it follows on the next load.
* Letting a `{{DISPLAYTITLE:}}` in the page's own wikitext win over the label.
* Reading list labels from columns on `neowiki_subject_page`, as #1333 proposes: labels were kept off the index
  pending the multilingual design, and only the lookup would change if they land there.

> <sub>AI-authored — Claude Code, `Opus 5.5 (xhigh)`; one-line ask from @alistair3149 (a minimal fix via a display title), extended at their request with the id beside the label and labelled Recent changes links, with scope and label source decided with them; diff not yet human-reviewed; tests written first or confirmed to fail without the change, local PHPUnit and `make cs` green, checked on the dev wiki.</sub>

Reviewed in the authoring session with three pr-review rounds (code-review, security-review, review-tests, a design
pass, an edge-case pass and simplify); their fixes are folded in. The second round moved list labels from new index
columns to Subject-slot reads after the design pass found the index had excluded labels by design. The third added
the guard for error pages shown after core chose a revision, the main-namespace rule and bounds on list reads. Local
PHPUnit
covered the non-Database suite and the Database classes around the page view, the index, search and the special
pages, including Recent changes rendered end to end. Checked on the dev wiki: the heading for a view, a diff and an
error page; Recent changes in both list modes, the watchlist, and a page transcluding Recent changes.

</details>


alistair3149:
image
image

Issue #1490, opened 2026-09-18
Improve display of labelless subjects


JeroenDeDauw:
If you have a Subject that uses the Person Schema and it has no label it will show `(unnamed Person)`.

This confused 2/2 users in the user testing sessions. They perceive it as needing to provide a label even if they don't want to. They also find it confusing to see "unnamed person" when their Person Schema has a Name property, which they provided for their Subject.

JeroenDeDauw:
Both these users are familiar with Wikibase, which also has optional labels, and seemingly is less confusing there. Then again, Wikidata has multilingual labels, so it's structurally different.

Multilingual Subject labels where mentioned, though do not seem tracked yet: https://github.com/ProfessionalWiki/NeoWiki/issues?q=is%3Aissue%20state%3Aopen%20multilingual. We do have:
* https://github.com/ProfessionalWiki/NeoWiki/issues/710
* https://github.com/ProfessionalWiki/NeoWiki/issues/726

I cannot recall any calls for multilingual subject labels before this week. Adding it might be net-negative in the form of bloat 99% of users do not need. Thoughts on this multilingual subject label tangent, @lozanaross @despens @dlindem? It came up in our recent chats and I wonder how important it is. Could you do without? What if there is a "monolingual text" equivalent Property Type, would you still want multilingual Subject labels?

despens:
In places like Switzerland, Belgium, many places in the Balkans, ... multilingual labels will be a major boon I guess.

PR #1519, opened 2026-09-22
Label Subjects through a label template in their Schema

alistair3149:
For https://github.com/ProfessionalWiki/NeoWiki/issues/1490

Schema authors can now define how Subjects without a typed label are labelled, from their own properties, with a label template such as `{Name}`. A Subject whose name is in a property is then shown under that name everywhere, instead of `(unnamed Person)`. The Schema editor control for choosing the naming property, and a `{Title}` default for new Schemas, follow in the next PR; until then a template is set in the Schema's source.

| Area | Change |
| :--- | :--- |
| Schema format | Optional `labelTemplate`, such as `{Title}` or `{Museum} attendance {Year}`. Each placeholder reads the Subject's first value of that property: a select as its option's label, a monolingual text as its first text. A save is refused, with a message naming the problem, when the template names a missing property, a relation, a boolean, or no property at all. |
| Label order | Typed label > template label > page name (Main Subject) > `(unnamed …)`. The template label is computed on read, never stored, and not marked as generated. |
| Schema editor | Saving a Schema keeps its template, and renaming a property renames its placeholders. |
| Docs | Schema format, the glossary and the pages stating how a Subject is named; ADR 35 records the decision. |

The label shows in:

- REST `displayName`
- the graph `name`, which label search and the relation picker use
- `rdfs:label` and a Mapping's `labelPredicate`
- Lua and `{{#neowiki_value}}`
- search result titles
- the page title of `POST /subjects`, which falls back to the id rather than a 409 when that page exists or may not be created

In the Subject editor:

| Case | Behaviour |
| :--- | :--- |
| Schema with a template, no typed label | The heading shows the template label as the fields change; its button focuses the field the template names. No label of the Subject's own is offered. |
| Schema with a template, label typed before | The label stays in the rename field; clearing it hands naming back to the template. |
| Schema without a template | The rename field, as before. |
| Creating on a new page | The page is titled after the template label. |
| Creating a relation target from typed text | The text goes into the field the template names when it is a text field, not into the label. |

Follow-ups, in the frontend PR that comes next:

- New Schemas start with a Title text property, removable and renamable, and the template `{Title}`. It is optional: an empty Title falls back to the page name for a Main Subject and to `(unnamed …)` otherwise, and a Schema author who wants every Subject titled marks it required.
- The Schema editor shows naming as "Named by: [property]", with a free-form template as an advanced option for Schemas named by several properties.
- Removing a property the template names is refused on save, with a message naming it; the "Named by" control will let the editor block that save and point at it instead.
- Draft Subjects in the relation pickers keep the name they were created with until saved; their names do not follow their fields yet.

Known limitations:

- Editing a template does not reproject: the graph `name` and `rdfs:label` follow at each page's next save or a Graph Store rebuild (ADR 35).
- The template label follows the Schema's readability: it comes through the same permission-gated Schema lookup as the rest of a Schema's content, which on a rendered page is that of the user the page is parsed for. Projections resolve Schemas as the user who triggered them, as they already did; a system lookup for projections is a separate change.
- When the fields leave the template nothing to read, the editor shows the Schema stand-in, while the server may give a Main Subject its page name.
- The editor formats numbers the JavaScript way and the server the PHP way; they differ only for very small, very large or long decimals.
- Placeholders reading another Subject's label through a relation are left to a later decision.

Considered, omitted:

- One naming mechanism per Schema in the data model, with the template replacing typed labels rather than only hiding them in the editor: a separate decision.
- `guide/getting-started.md` still says a new page is titled after the label or the title given; a template label is the Subject's label in the glossary's sense.

## Manual Browser Check

1. On a dev wiki with demo data, edit `Schema:Publication` as source and add `"labelTemplate": "{Title} ({Year})"`.
2. Open `A Survey of Provenance Models in Open Knowledge Graphs`, edit its Subject, clear the label and save. The infobox heading reads "A Survey of Provenance Models in Open Knowledge Graphs (2020)".
3. Edit it again and change Title: the heading follows as you type. Its button, "Edit Title, which names this subject", puts the cursor in Title.
4. Edit `ACME Inc`, whose Schema has no template: the heading still offers "Rename subject".
5. Open `/w/rest.php/neowiki/v0/subject-labels?schema=Publication&search=A%20Survey`, the search behind the relation picker: the Subject is listed under that label.
6. Change the template to `{Title} by {Authors}` and save the Schema: the save is refused, naming "Authors" as a relation property.

> <sub>AI-authored — Claude Code, `Opus 5.5 (1M context)`; design chosen by @alistair3149 from a comparison of approaches, with the naming (`label`, `labelTemplate`) and editor behaviour agreed over several rounds; diff not yet human-reviewed; code-review, security-review, test-review, design, edge-case, text-review and simplify passes applied, blocking findings verified by refute-first agents; phpcs, phpstan, the non-Database PHPUnit suite and the Database classes on the changed paths run locally, full vitest, ts-build and ts-lint green, new tests mutation-checked, behaviour checked on a worktree wiki through REST and a headless browser; CI green.</sub>


alistair3149:
This is a MVP implementation, with the follow-up noted in the PR message.
The label template is currently only exposed in the schema JSON.
I did not include a frontend UI to edit that yet to avoid growing the PR.
Existing schema without the `labelTemplate` will follow the previous behavior for now.

https://github.com/user-attachments/assets/f76429b1-6914-41bc-977a-a23dba720dc0

JeroenDeDauw:
Let's defer this till after we have an MVP release and some real usage

Some thoughts:
* Current demand signals are weak and in part a response to the "(unnamed subject)" in the UIs, which has simpler fixes
* There are many design choices to be made here about what exactly the feature is and how it behaves for which I'd want to talk to people that need it before implementing.
* We might end up with multilingual labels, which would change things
* There won't always be a "title/name" property
* `{name} ({born})` works poorly when "born" is not set
* https://professional.wiki/en/extension/automated-values is good to compare against, even though we do not have to follow the same approach. For instance in AV the label always ends up being what the rule is, you cannot override it with a "set label" as you could in this PR. You can argue for either behavior, depends on your use case.
* This is additional product complexity and the fraction of users that need or want it is possibly small. Might be best put in an extension.

Getting this into production-ready shape will take a bunch of implementation, review, and testing time. Let's direct that capacity towards getting the core product used first.

alistair3149:
I agree on your points, let's defer this for now and I will keep this as a draft.

